PR #61, opened 2025-09-12
feat: added functionality for refresh token for openedx

shavaiz-hussain:
**Changes**
- Added refresh token tool and endpoint for refresh token

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ chrono = "0.4.41"
 schemars = "1.0.4"
 url = "2.5.4"
 form_urlencoded = "1.2.1"
+bytes = "1.10.1"
 
 [profile.dist]
 inherits = "release"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,8 +26,6 @@ chrono = "0.4.41"
 schemars = "1.0.4"
 url = "2.5.4"
 form_urlencoded = "1.2.1"
-bytes = "1.10.1"
-
 [profile.dist]
 inherits = "release"
 lto = "thin"
--- a/src/tools/openedx_tools.rs
+++ b/src/tools/openedx_tools.rs
@@ -25,6 +25,7 @@ use crate::{
     },
     utils::cached_schema_for_type,
 };
+use crate::plugins::openedx::types::OpenEdxRefreshTokenPayload;
 
 impl SparkthMCPServer {
     async fn openedx_create_basic_component(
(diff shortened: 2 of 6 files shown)


src/tools/openedx_tools.rs
```diff
@@ -138,10 +139,17 @@ impl SparkthMCPServer {
...
             .get_token(&username, &password)
             .await
-            .map(|token| {
+            .map(|auth_json| {
                 let who = client.username().unwrap_or(&username);
+                let access_token = auth_json
+                    .get("access_token")
+                    .and_then(|v| v.as_str())
+                    .unwrap_or_default();
+                let refresh_token = auth_json.get("refresh_token").and_then(|v| v.as_str());
```
zaira-bibi, line 148: This could be made more idiomatic by defining a struct for `auth_json`. It would help reduce duplication and improve readability.

src/tools/openedx_tools.rs
```diff
@@ -138,10 +139,15 @@ impl SparkthMCPServer {
         client
             .get_token(&username, &password)
             .await
-            .map(|token| {
+            .and_then(|auth_json| {
+                serde_json::from_value::<TokenResponse>(auth_json)
+                    .map_err(|e| LMSError::Other(format!("failed to parse auth payload: {e}")))
```
zaira-bibi, line 144: Instead of doing this here, why not return the `TokenResponse` from `get_token()` directly, instead of `Value`?

src/tools/openedx_tools.rs
```diff
@@ -150,6 +156,41 @@ impl SparkthMCPServer {
...
+        }): Parameters<OpenEdxRefreshTokenPayload>,
+    ) -> Result<String, String> {
+        let mut client = OpenEdxClient::new(&lms_url, None);
+
+        client
+            .refresh_access_token(&refresh_token)
+            .await
+            .and_then(|auth_json| {
+                serde_json::from_value::<TokenResponse>(auth_json)
+                    .map_err(|e| LMSError::Other(format!("failed to parse refresh payload: {e}")))
```
zaira-bibi, line 179: Similarly, we can return a `TokenResponse` from `refresh_access_token()` instead of `Value`.

Cargo.toml
```diff
@@ -26,6 +26,7 @@ chrono = "0.4.41"
 schemars = "1.0.4"
 url = "2.5.4"
 form_urlencoded = "1.2.1"
+bytes = "1.10.1"
```
zaira-bibi, line 29: We can remove the import as it's no longer being used.

zaira-bibi (changes requested):

hamza-56 (approved):

PR #93, opened 2025-11-10
implement database-backed configurations for plugins

### LINKED ISSUE #91: Configuration Parameters Management on the backend and frontend
[Backend] - Implement API Endpoints
[Front End] - Basic POC for end users to view, enable and configure plugins

End Goal: Users will be able to enable a plugin and set its configurations at runtime. Each user will have its own set of configuration settings.

zaira-bibi:
Database:

Add plugins table with metadata (name, version, description, enabled, type)
Add plugin_configs table for key-value configuration storage
Core Plugin System:

Generate config schemas (with type, description, required, default) at compile time from macro definitions.
API Endpoints:

GET /api/plugins - List all plugins with configurations for a specific user
GET /api/plugins/:id - Get specific plugin with config
PATCH /api/plugins/:id/config - Update plugin configuration
PATCH /api/plugins/:id/toggle - Enable/disable plugin

Closes #91.


regisb:
Let's close this now that we are switching to Python.

PR #103, opened 2025-11-19
Setup SQLModel with Alembic and create user table

### LINKED ISSUE #98: Rewrite the User management backend & Authentication APIs in Python
In particular, remember that enabled plugins are user-dependent.

hamza-56:
This PR introduces User management functionality, including:

1. Setup of SQLModel with Alembic for database migrations.
2. User creation via CLI: `python cli.py create-user`.
3. Password reset via CLI: `python cli.py reset-password`. 
4. Updates to the README with relevant instructions.

Closes #98 (a)


migrations/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
regisb: Let's get rid of this file.

hamza-56: Done ✅

PR #108, opened 2025-11-27
feat: add Docker Compose development environment

### LINKED ISSUE #101: Docker image is broken and needs to be modified after the Python rewrite
Currently, the Docker image build is failing: https://github.com/edly-io/sparkth/actions/runs/19459286309
We need to fix the build and make sure that the image is still working on https://sparkth-mcp.edly.space.

hamza-56:
- Add docker-compose.yml with api, db, and migrations services
  - Rewrite Dockerfile to use uv-based Python image
  - Add .dockerignore for optimized builds
  - Update Makefile with Docker commands (up, down, logs, shell, db-shell)
  - Move alembic.ini to project root and fix migration imports
  - Enable tests in CI workflow
  
  Closes #101

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -10,7 +10,6 @@ on:
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
-  BUILD_PROFILE: dist
 
 jobs:
   build-and-publish:
@@ -22,7 +21,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
@@ -36,7 +35,7 @@ jobs:
 
       - name: Extract Docker metadata
         id: meta
-        uses: docker/metadata-action@v4
+        uses: docker/metadata-action@v5
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
@@ -49,5 +48,3 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          build-args: |
-            PROFILE=${{ env.BUILD_PROFILE }}


Dockerfile
```diff
@@ -1,45 +1,33 @@
...
-# Copy project files
-COPY . .
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --frozen --no-install-project --no-dev
 
-# Define build profile (default: dist)
-ARG PROFILE=dist
+COPY . /app
```
regisb: Let's avoid a rebuild of this step when there's a single change in the source code.

docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+    command: >
+      sh -c "
+      echo '⏳ Waiting for DB...' &&
+      alembic upgrade head &&
+      echo '✅ Alembic migrations complete'
+      "
+    depends_on:
+      - db
+
+volumes:
```
regisb: Let's get rid of this.

docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+  db:
+    image: postgres:18-alpine
+    container_name: db
+    restart: always
+    environment:
+      - POSTGRES_USER=sparkth
+      - POSTGRES_PASSWORD=sparkth_password
+      - POSTGRES_DB=sparkth
+    ports:
+      - "5432:5432"
```
regisb: Let's remove this, for security reasons.

docker-compose.yml
```diff
@@ -0,0 +1,54 @@
+services:
+  # ----------------------------------------
+  # Sparkth API Service
+  # ----------------------------------------
+  api:
+    build: .
+    container_name: api
```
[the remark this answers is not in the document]
hamza-56: Removed ✅

docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+    container_name: db
+    restart: always
+    environment:
+      - POSTGRES_USER=sparkth
+      - POSTGRES_PASSWORD=sparkth_password
+      - POSTGRES_DB=sparkth
+    ports:
+      - "5432:5432"
+    volumes:
+      - sparkth_data:/var/lib/postgresql/data
```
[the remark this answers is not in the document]
hamza-56: Updated ✅

Dockerfile
hamza-56: This step is mandatory for `pyproject.toml` CLI entry points to work and allows us to import the app packages correctly. It also byte-compiles the source code for faster startup.

Modifying source code will technically re-run this final sync, but it will not trigger a re-install of external dependencies. Since the heavy dependencies are cached in the previous layer (via `--no-install-project`), this step only installs the local project, making the rebuild extremely fast.

https://github.com/astral-sh/uv-docker-example/blob/main/multistage.Dockerfile

shavaiz-hussain (approved):

Issue #111, opened 2025-12-03
Expand MCP tool testing beyond Claude to include Gemini and OpenAI


zaira-bibi:
MCP tools are currently only tested with Claude, limiting our confidence in cross-LLM compatibility and correctness. We can test with the following tools:

- [Gemini CLI](https://geminicli.com/)
- [Codex](https://openai.com/codex/) for OpenAI

zaira-bibi:
As per Codex [docs](https://developers.openai.com/codex/cli/):

> Sign in with your ChatGPT Plus, Pro, Business, Edu, or Enterprise account for the quickest setup, or provide an API key if you need fine-grained control.

Would need one of these to test with Codex CLI.

PR #117, opened 2025-12-10
Make mcp compatible with Gemini CLI

zaira-bibi:
The `type` field in the `ModuleItem` class conflicted with reserved/ internal usage of "type" in the schema system, causing an `INVALID_ARGUMENT` error for every tool call. See issue #111.

Renaming the field to `module_type` and rebuilding the request payload with the correct field names resolved this issue.

### DIFF
--- a/app/mcp/canvas/types.py
+++ b/app/mcp/canvas/types.py
@@ -15,10 +15,8 @@ class CourseParams(BaseModel):
     auth: AuthenticationPayload
 
 
-class PaginationParams(BaseModel):
+class PaginationParams(CourseParams):
     page: int
-    course_id: int
-    auth: AuthenticationPayload
 
 
 class CourseFormat(str, Enum):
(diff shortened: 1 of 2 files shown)


app/mcp/canvas/types.py
```diff
@@ -11,6 +11,11 @@ class AuthenticationPayload(BaseModel):
 
 
 class CourseParams(BaseModel):
+    course_id: int
+    auth: AuthenticationPayload
+
+
+class PaginationParams(BaseModel):
```
hamza-56: consider using inheritance to reduce duplication:
```
class PaginationParams(CourseParams):
    page: int
```

app/mcp/canvas/types.py
```diff
@@ -100,7 +105,7 @@ class ModuleItemCompletionRequirement(BaseModel):
 
 class ModuleItem(BaseModel):
```
hamza-56: Consider adding a method like:
```
 def to_canvas_payload(self) -> dict:
     data = self.model_dump()
     data['type'] = data.pop('module_type')
     return data
```

PR #127, opened 2026-01-02
Refactor endpoints to return config

zaira-bibi:
In order to display config keys with the user plugin on frontend, the following changes are required:

- refactor endpoints to return config with user-specific plugins
- add CORS middleware to allow frontend requests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,3 +38,4 @@ jobs:
         env:
           DATABASE_URL: sqlite:///./test.db
           SECRET_KEY: secret-key-for-ci
+          FRONTEND_URL: http://frontend_url.com


docker-compose.yml
```diff
@@ -42,4 +42,4 @@ services:
       echo '✅ Alembic migrations complete'
       "
     depends_on:
-      - db
+      - db
```
hamza-56: Revert this change

app/main.py
```diff
@@ -101,6 +102,16 @@ async def lifespan(application: FastAPI) -> AsyncIterator[None]:
...
+
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["http://localhost:3000", "http://127.0.0.1:3000"],
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
+
+
```
hamza-56: Consider:
1. Moving these to environment variables or settings for different environments (dev/staging/prod).
2. For production, avoid allow_methods=["*"] and allow_headers=["*"]
3. CORSMiddleware should typically be added last so it wraps all routes

PR #166, opened 2026-02-23
Chat history with conversation persistence 

### LINKED ISSUE #157: [Sparkth] Chat History
Implement full chat history and display attachments in the correct pill format for previous conversations.

zaira-bibi:
This commit introduces end-to-end implementation of chat history, allowing users to resume previous conversations and correctly render file attachments from past sessions.

Main changes:

- **Routing**
    - chats identified via search param: `/dashboard/chat?id=123`
    - new chats start at /dashboard/chat and redirect to `?id=<id>`

- **Sidebar**
    - renders below nav items in AppSidebar only when on the
    `/dashboard/chat` route
    - fetches and lists recent conversations with title/relative date
    
 Closes #157

### DIFF
--- a/app/api/v1/user_plugins.py
+++ b/app/api/v1/user_plugins.py
@@ -27,10 +27,6 @@ from app.services.plugin import (
 logger = logging.getLogger()
 
 
-# Get the root logger
-logger = logging.getLogger()
-
-
 router: APIRouter = APIRouter()
 
 
--- a/app/core_plugins/chat/service.py
+++ b/app/core_plugins/chat/service.py
@@ -224,7 +224,10 @@ class ChatService:
         self, session: AsyncSession, conversation_id: int, limit: int | None = None, offset: int | None = None
     ) -> list[Message]:
         statement = (
-            select(Message).where(Message.conversation_id == conversation_id).order_by(col(Message.created_at).asc())
+            select(Message)
+            .where(Message.conversation_id == conversation_id)
+            .where(Message.is_error == False)  # noqa: E712
+            .order_by(col(Message.created_at).asc())
         )
 
         if limit:
--- a/app/migrations/versions/53632e4b3fc3_add_attachement_metadata_to_message.py
+++ b/app/migrations/versions/53632e4b3fc3_add_attachement_metadata_to_message.py
@@ -10,7 +10,6 @@ from typing import Sequence, Union
 from alembic import op
 import sqlalchemy as sa
 import sqlmodel
-from sqlalchemy.dialects import postgresql
 
 # revision identifiers, used by Alembic.
 revision: str = '53632e4b3fc3'
@@ -27,7 +26,7 @@ def upgrade() -> None:
             "message_type",
             sa.String(20),
             nullable=False,
-            server_default="text",
+            server_default='text'
         ),
     )
     op.add_column(
--- a/frontend/components/ui/Alert.tsx
+++ b/frontend/components/ui/Alert.tsx
@@ -1,4 +1,5 @@
 import { HTMLAttributes, forwardRef } from "react";
+import { X } from "lucide-react";
 import { cva, type VariantProps } from "class-variance-authority";
 import { cn } from "@/lib/utils";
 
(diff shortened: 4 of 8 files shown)


hamza-56:
1. **Migration adds non-nullable `is_error` column without `server_default` -- will fail on populated tables.** The `is_error` column is added as `nullable=False` with no default value. On PostgreSQL, this raises a `NOT NULL` constraint violation for any table with existing rows. The sibling migration (`53632e4b3fc3`) correctly uses `server_default="text"` for the `message_type` column, but this one does not.

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/app/migrations/versions/12ec22bbfa0d_add_is_error_to_chat_messages.py#L23-L25

2. **`update_api_key` caches plaintext API key, but `get_api_key` now expects encrypted values.** `create_api_key` in `service.py` was updated to store `encrypted_key` in Redis, and `get_api_key` decrypts the cached value. However, `update_api_key` in `routes.py` still stores the raw plaintext `key_update.api_key`. The next `get_api_key` call will try to decrypt plaintext, fail, and fall back to the database on every request until the cache entry expires.

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/app/core_plugins/chat/routes.py#L145-L147

Compare with the correct pattern in service.py:

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/app/core_plugins/chat/service.py#L38-L40

3. **Error messages saved with `is_error=True` are included in subsequent AI provider calls.** When streaming fails, the route saves a synthetic error message with `is_error=True`. However, `get_conversation_messages` does not filter out `is_error` messages. On the next user message, the error text is sent to the AI as a legitimate assistant turn, distorting the conversation context.

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/app/core_plugins/chat/routes.py#L366-L376

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/app/core_plugins/chat/service.py#L223-L228

4. **`ChatHistorySection` sidebar never refreshes after a new conversation is created.** The `useEffect` that fetches the conversation list has `[token]` as its only dependency. After `ChatInterface` creates a new conversation and calls `router.replace`, the token does not change, so the sidebar never re-fetches. The new conversation will not appear in "Recent chats" until a full page reload.

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/frontend/plugins/chat/components/ChatHistorySection.tsx#L44-L52

5. **`historyState` is not reset to loading when navigating between conversations.** When the user clicks a different conversation in the sidebar, `conversationId` changes and the `useEffect` fires, but `historyState` is not reset to `{ loading: true, messages: [] }` before the fetch begins. The previous conversation's messages will flash on screen until the new fetch completes.

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/frontend/plugins/chat/ChatInterface.tsx#L69-L108

hamza-56:
Found 4 additional issues:

1. **Commented-out dead code left in `ChatInterface.tsx`.** Old state variables (`messages`, `loadingHistory`) were commented out instead of deleted after being replaced by the unified `historyState` approach. These are refactoring artifacts that should be removed.

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/frontend/plugins/chat/ChatInterface.tsx#L39-L41

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/frontend/plugins/chat/ChatInterface.tsx#L48

2. **Duplicate logger declaration in `user_plugins.py`.** The PR introduces a second identical `# Get the root logger` + `logger = logging.getLogger()` block immediately after the first one.

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/app/api/v1/user_plugins.py#L26-L31

3. **`str(e)` still leaks internal error details to the client in the streaming SSE path.** Previously flagged in PR #156. The non-streaming paths were correctly fixed in this PR to return generic messages, but the streaming exception handler still sends raw `str(e)` to the browser via SSE. The `logger.error` call above it already logs the original exception server-side, so the SSE payload should use a generic message instead.

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/app/core_plugins/chat/routes.py#L374-L376

4. **Tool iteration limit counts rounds, not individual tool executions.** Previously flagged in PR #156. Both `_send_message_with_tools` and `_stream_message_with_tools` increment `iteration` once per round, but a single LLM response can return many `tool_calls` that all execute before the counter increments. The `max_iterations = 15` guard is effectively `max_rounds`, not `max_tool_executions`.

https://github.com/edly-io/sparkth/blob/1bbcf089581053e489f97410a638b89b65d65c86/app/core_plugins/chat/providers.py#L167-L204

PR #238, opened 2026-04-07
fix(plugins): auto-inject LMS credentials from plugin config into LLM context

zaira-bibi:
## What

Users who have configured an LMS (Open edX or Canvas etc.) via the Plugins page were required to re-type their credentials into every chat message before performing any LMS operation. This PR eliminates that friction by automatically injecting stored credentials into the LLM's system prompt at request time.

### New: `app/core_plugins/chat/lms_credentials.py`
- `build_lms_credentials_message(session, user_id, tools)` — fetches the user's stored plugin config and returns a system prompt extension.
- Credentials present → LLM receives them and uses them automatically.
- Credentials absent → LLM is instructed to direct the user to **My Plugins** instead of asking for credentials in chat (rules 1–4). Rule 5 (use creds automatically) only appears when credentials are actually available.
- `_has_lms_tools(tools)` — short-circuits the DB call when no LMS tools are active in the current request.
- Both helpers are fully generic: LMS plugin names and tool prefixes are derived from `PLUGIN_CONFIG_CLASSES` at runtime — no hardcoded strings.

### Extended: `PluginConfig` base (`app/plugins/config_base.py`)
Two opt-in hooks added (both return `None` on the base class, so non-LMS plugins require zero changes):
- `lms_tool_prefix() -> str | None` — declares the tool-name prefix for this LMS.
- `to_lms_credentials_hint() -> str | None` — returns the formatted credential block to include in the system prompt.

### Extended: `OpenEdxConfig`, `CanvasConfig`
Both config classes override the two hooks above.

### Modified: `app/core_plugins/chat/routes.py`
After tools are resolved, `build_lms_credentials_message` is called and its result is appended to `provider.system_prompt` (not inserted into the message list, so the conversation history stays clean and no extra tokens accumulate per turn).

### Updated: `app/plugins/PLUGIN_GUIDE.md`
The new section sits right after the base config definition and covers:

- When to use the two methods (lms_tool_prefix and to_lms_credentials_hint)
- What each method does and why it exists
- A complete copy-paste example for a new LMS plugin
- A note that non-LMS plugins need no changes (both default to None)

### Tests: `tests/chat/test_lms_credentials.py`
16 unit tests covering all branches: no tools, no LMS tools, credentials configured, credentials missing, both LMS configured, invalid/empty config, rule-5 presence/absence.

## Adding a New LMS in Future
1. Add credentials fields to the new `PluginConfig` subclass.
2. Override `lms_tool_prefix()` and `to_lms_credentials_hint()`.
3. Register in `PLUGIN_CONFIG_CLASSES`.

No changes to `lms_credentials.py`, `routes.py`, or tests are needed.

## Test Plan
- [ ] Configure Open edX credentials via My Plugins → ask the LLM to publish a
  course → confirm it authenticates without prompting for credentials
- [ ] Configure Canvas credentials → same flow
- [ ] Remove credentials → confirm LLM responds with "configure on My Plugins page"
  instead of asking for credentials in chat
- [ ] Run `pytest tests/chat/test_lms_credentials.py` → 16 passed

## Note
Fix Alembic migration failure by resolving multiple head revisions with a merge migration

Closes #233.

### DIFF
--- a/app/core_plugins/canvas/config.py
+++ b/app/core_plugins/canvas/config.py
@@ -6,3 +6,10 @@ from app.plugins.config_base import PluginConfig
 class CanvasConfig(PluginConfig):
     api_url: HttpUrl = Field(..., description="Canvas API URL")
     api_key: str = Field(..., description="Canvas API key", min_length=1)
+
+    @classmethod
+    def lms_tool_prefix(cls) -> str:
+        return "canvas_"
+
+    def to_lms_credentials_hint(self) -> str:
+        return f"Canvas credentials:\n  api_url: {self.api_url}\n  api_token: {self.api_key}"
--- a/app/core_plugins/chat/lms_credentials.py
+++ b/app/core_plugins/chat/lms_credentials.py
@@ -1,0 +1,92 @@
+from typing import Any
+
+from sqlmodel.ext.asyncio.session import AsyncSession
+
+_LMS_RULES = (
+    "If the user asks whether you have their credentials for any LMS (e.g., 'Do you have my <LMS NAME> credentials?') "
+    "OR requests any LMS-related action (such as publishing a course, fetching courses, or creating/updating content), "
+    "you must follow these rules:\n\n"
+    "1. NEVER request, accept, or display LMS credentials in the chat.\n"
+    "2. If the LMS is NOT configured:\n"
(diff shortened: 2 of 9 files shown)


abdulrafey1 (approved):
LGTM!

PR #251, opened 2026-04-13
fix(ci): checkout PR head ref so react-doctor diff mode works

hamza-56:
## What does this PR do?

Fixes react-doctor running a full project scan instead of only checking files changed in the PR.

## Why is this needed?

The default `pull_request` checkout creates a detached HEAD (merge commit). React-doctor's `getCurrentBranch()` returns `null` in detached HEAD state, causing it to skip `diff: main` and fall back to scanning the entire `frontend/` directory.

Adding `ref: ${{ github.event.pull_request.head.ref }}` checks out the actual PR branch head so `git merge-base main HEAD` works correctly.

### DIFF
--- a/.github/workflows/react-doctor.yml
+++ b/.github/workflows/react-doctor.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
         with:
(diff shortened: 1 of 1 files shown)


abdulrafey1 (approved):
LGTM

PR #270, opened 2026-04-20
chore(migrations): merge divergent alembic heads

hamza-56:
## What does this PR do?

Merges divergent Alembic migration heads (`5ae910dbda5d` and `i2b3c4d5e6f7`) into a single lineage so `alembic upgrade head` works without errors.

## Test plan

- [ ] `alembic heads` returns a single head
- [ ] `alembic upgrade head` runs without errors


abdulrafey1 (approved):
lookd good

PR #272, opened 2026-04-20
feat(frontend): add RAG status column and re-sync button to resources tab

hamza-56:
## Summary
- Add per-file RAG status column to the Resources tab (Indexed/Queued/Processing/Failed chips with error tooltip)
- Include `rag_status` and `rag_error` in `DriveFileResponse` API responses
- Restore the re-sync button (lost during resources page redesign) to the ConnectionStatus card
- Rename "Status" header to "Sync Status" for clarity alongside the new "RAG Status" column
- Add icons to all connection action buttons and style Disconnect as destructive (red)


zaira-bibi (approved):
LGTM!

Issue #261, opened 2026-04-16
Centralized LLM config management - BE


zaira-bibi:
### Scope

- [x] Move `providers.py` from `app/core_plugins/chat/` → `app/llm/` 
- [x] Move `encryption.py` and `cache.py` from `app/core_plugins/chat/` → `app/core`
- [x] `app/llm/models.py` — `LLMConfig` SQLModel (`id`, `user_id`, `name`, `provider`, `model`, `encrypted_key`, `masked_key`, `is_active`, `last_used_at`)
- [x] `app/llm/service.py` — `LLMConfigService`: `create`, `list`, `get`, `update`, `rotate_key`, `delete`, `resolve`
- [x] `app/llm/routes.py` — `POST /configs`, `GET /configs`, `PATCH /configs/{id}`, `PUT /configs/{id}/key`, `DELETE /configs/{id}`
- [x] `app/llm/plugin_adapter.py` — `LLMConfigPluginAdapter` base (validates + resolves `llm_config_id`)
- [x] Alembic migration: copy `chat_provider_api_keys` → `llm_configs` (ID-preserving), drop old table
- [x] Chat plugin: remove `/keys` endpoints, update `ChatUserConfig` to `llm_config_id` + `llm_model_override`

### Migration Strategy
A single Alembic migration:

- [x] Creates `llm_configs`
- [x] Copies all rows from `chat_provider_api_keys`, preserving primary keys (existing chat config `provider_api_key_ref` integer references remain valid)
- [x] Auto-generates name as "{provider}-{id}" for migrated rows
- [x] Drops `chat_provider_api_keys`

### Breaking Changes
- [x] `DELETE /api/v1/chat/keys` and its 3 sibling endpoints are removed
- [x] `ChatCompletionRequest.provider: str` → `llm_config_id: int`
- [x] `ChatUserConfig.provider_api_key_ref` → `llm_config_id` (application layer only — no JSON column migration needed due to ID preservation)

PR #275, opened 2026-04-21
feat(api): add LLM config REST endpoints and decouple from chat plugin

zaira-bibi:
## What
As part of #261, add a first-class LLM config management API (`/api/v1/llm/configs`) so users can store, rotate, and activate provider API keys independently of the chat plugin.

## Changes
- `feat(api)`: add CRUD endpoints for LLM configs — create, list, update, rotate key, set active, delete
- `feat(llm)`: add `LLMConfigService` with encrypt/decrypt, cache, soft-delete, and `resolve()` returning (`config`, `api_key`) tuple
- `feat(llm)`: add `LLMConfigPluginAdapter` — validates `llm_config_id` ownership on write, enriches with name/provider/model on read
- `refactor(core)`: promote `ENCRYPTION_KEY`, `REDIS_URL`, `REDIS_KEY_TTL` from `ChatSystemConfig` to core `Settings`
- `refactor(plugins)`: decouple chat plugin from encryption/cache — `ChatService` no longer owns key management
- `refactor(plugins)`: replace `api_key_id` with `llm_config_id` on `Conversation` model; chat routes resolve key via `LLMConfigService.resolve()`
- `chore(migrations)`: add migrations replacing `api_key_id` with `llm_config_id` on conversations table and dropping the `chat_provider_api_keys`.
- `test(llm)`: add route tests covering all 6 endpoints and unauthenticated rejection
- `test(llm)`: add service tests covering encrypt, cache, mask, set_active, rotate_key edge cases
- `test(llm)`: add adapter tests for preprocess ownership check and postprocess enrichment

## How to Test
1. Copy `.env.example `→ `.env`, set `ENCRYPTION_KEY` (Fernet key), `REDIS_URL`, `DATABASE_URL`
2. Run make migrations to apply the `api_key_id` → `llm_config_id` and drop `chat_provider_api_key` table migration.
4. `POST /api/v1/llm/configs` with `{"name": "My Key", "provider": "anthropic", "model": "claude-sonnet-4-20250514", "api_key": "sk-..."}` — expect `201`, masked key in response, no plaintext key
5. `GET /api/v1/llm/configs` — expect list with the new config
6. `PUT /api/v1/llm/configs/{id}/key` with a new key — expect masked key updated
7. `PATCH /api/v1/llm/configs/{id}/active` with `{"is_active": false}` — expect config excluded from list; re-activate and it reappears
8. Configure chat plugin with `llm_config_id` pointing to the created config, start a chat — expect it to work
9. `make test` — all tests pass

## Notes
- **Breaking env var change**: `CHAT_ENCRYPTION_KEY` → `LLM_ENCRYPTION_KEY`, `CHAT_REDIS_URL` → `REDIS_URL` in `.env`
- Migration required: adds `llm_config_id` column, drops `api_key_id` on conversations table and drops `chat_provider_api_keys`
- Existing deployments with `api_key_id` data will need data backfilled before migration if in production use


app/api/v1/llm.py
```diff
@@ -0,0 +1,196 @@
...
+@router.patch("/configs/{config_id}", response_model=LLMConfigResponse)
+async def update_llm_config(
+    config_id: int,
+    body: LLMConfigUpdate,
+    current_user: User = Depends(get_current_user),
+    session: AsyncSession = Depends(get_async_session),
+    service: LLMConfigService = Depends(get_llm_service),
+) -> LLMConfigResponse:
+    if not body.name and not body.model:
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_CONTENT, detail="Empty body. Nothing to update.")
```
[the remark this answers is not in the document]
zaira-bibi: @claude 
only `status.HTTP_422_UNPROCESSABLE_CONTENT` exists in `fastapi.status`.
https://fastapi.tiangolo.com/reference/status/#fastapi.status.HTTP_422_UNPROCESSABLE_CONTENT

app/llm/service.py
```diff
@@ -144,28 +153,47 @@ async def delete(self, session: AsyncSession, user_id: int, config_id: int) -> b
...
-                f"LLMConfig {config_id} has no model set. Update it via PATCH /api/v1/llm-configs/{config_id} before use."
-            )
-
         decrypted = self.encryption.decrypt(config.encrypted_key)
         await self.cache.set(cache_key, config.encrypted_key)
         config.last_used_at = datetime.now(timezone.utc)
         session.add(config)
         await session.flush()
-        return decrypted
+        return config, decrypted
```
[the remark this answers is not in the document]
zaira-bibi: @claude the DB flush only happens on cache miss. the cache-hit path (line 189) returns early without touching the DB. so it's not "per request" in the hot path; it's per-cache-expiry-window, which is already a natural throttle.

app/core/config.py
```diff
@@ -28,6 +28,11 @@ class Settings(BaseSettings):
     SLACK_SIGNING_SECRET: str
     SLACK_REDIRECT_URI: str
 
+    # Encryption & cache (shared across plugins)
+    ENCRYPTION_KEY: str
```
abdulrafey1: if this encryption key is only used to encrypt the API keys, then lets rename it to be more specific.
If this API key is expected to be used for more than just one thing, then we can keep it generic, let me know!

app/core_plugins/chat/routes.py
```diff
@@ -416,21 +290,12 @@ async def chat_completion(
...
-        api_key_record = result.first()
-
         conversation = await service.create_conversation(
             session=session,
             user_id=current_user.id,  # type: ignore
-            api_key_id=api_key_record.id if api_key_record else None,  # type: ignore
-            provider=request.provider,
-            model=request.model,
+            llm_config_id=request.llm_config_id,
+            provider=provider_name,
```
abdulrafey1, line 340: Why do we need the provider name here?
Is the provider not in the LLM config?

app/core_plugins/chat/routes.py
```diff
@@ -619,8 +484,8 @@ async def chat_completion(
                     content=response["content"],
                 ),
                 conversation_id=conversation.uuid,
-                model=request.model,
-                provider=request.provider,
+                model=model,
+                provider=provider_name,
```
abdulrafey1, line 539: same question

app/core/config.py
zaira-bibi: For now, it’s only being used to encrypt API keys, so we can definitely give it a more specific name. If additional use cases come up in the future, we can revisit and generalize it then.

app/core_plugins/chat/routes.py
zaira-bibi: It comes from the LLM config. See line 268:

```python
provider_name = llm_config.provider
```

app/core_plugins/chat/routes.py
zaira-bibi: It comes from the LLM config. See line 268:

```python
provider_name = llm_config.provider
```

.github/workflows/ci.yml
```diff
@@ -68,3 +68,4 @@ jobs:
           SLACK_CLIENT_SECRET: fake-client-secret
           SLACK_SIGNING_SECRET: fake-signing-secret
           SLACK_REDIRECT_URI: http://localhost/callback
+          LLM_ENCRYPTION_KEY: fake-encryption-key-for-ci
```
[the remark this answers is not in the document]
abdulrafey1: @zaira-bibi can we please check if the key is a valid Fernet key or not, if yes then lets comment here and resolve this issue.

app/migrations/versions/2e34759ac1e1_replace_api_key_id_with_llm_config_id_.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+import sqlmodel
+
+
+# revision identifiers, used by Alembic.
+revision: str = '2e34759ac1e1'
+down_revision: Union[str, Sequence[str], None] = 'ed5bc53f5b16'
```
hamza-56: `ed5bc53f5b16` is already an internal node in the chain on main (merged via `9c055c4d542a`, with the chain running through `9566994a20ee` to `e8530a5a5a0e`). Branching off it produces a split head, and `make migrations` will refuse to upgrade.

app/migrations/versions/2e34759ac1e1_replace_api_key_id_with_llm_config_id_.py
```diff
@@ -0,0 +1,43 @@
...
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('chat_conversations', sa.Column('llm_config_id', sa.Integer(), nullable=True))
+    op.drop_index(op.f('ix_chat_conversations_api_key_id'), table_name='chat_conversations')
+    op.create_index(op.f('ix_chat_conversations_llm_config_id'), 'chat_conversations', ['llm_config_id'], unique=False)
+    op.drop_constraint(op.f('chat_conversations_api_key_id_fkey'), 'chat_conversations', type_='foreignkey')
+    op.create_foreign_key(None, 'chat_conversations', 'llm_configs', ['llm_config_id'], ['id'], ondelete='SET NULL')
+    op.drop_column('chat_conversations', 'api_key_id')
```
hamza-56, line 56: There's no backfill from `chat_provider_api_keys` into `llm_configs`, and no translation from old `api_key_id` to new `llm_config_id`. After this runs in production, every existing conversation has `llm_config_id=NULL` and the next chat completion call returns 404 since `llm_config_id` is now required.

app/migrations/versions/5c3e0d64672c_drop_chat_provider_api_key.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(op.f('idx_user_provider_active'), table_name='chat_provider_api_keys')
+    op.drop_index(op.f('ix_chat_provider_api_keys_is_deleted'), table_name='chat_provider_api_keys')
+    op.drop_index(op.f('ix_chat_provider_api_keys_provider'), table_name='chat_provider_api_keys')
+    op.drop_index(op.f('ix_chat_provider_api_keys_user_id'), table_name='chat_provider_api_keys')
+    op.drop_table('chat_provider_api_keys')
```
hamza-56, line 29: This drops every user's encrypted API keys with no recovery path. The `downgrade()` restores the table schema but not the row data.

app/api/v1/llm.py
```diff
@@ -0,0 +1,193 @@
...
+    return LLMConfigResponse(
+        id=config.id,  # type: ignore[arg-type]
+        name=config.name,
+        provider=config.provider,
+        model=config.model,
+        masked_key=config.masked_key,
+        is_active=config.is_active,
+        created_at=config.created_at,
+        last_used_at=config.last_used_at,
+    )
```
hamza-56: The manual field-by-field mapping silently drops any field added to `LLMConfig` later.

app/api/v1/llm.py
```diff
@@ -0,0 +1,193 @@
...
+            user_id=current_user.id,  # type: ignore[arg-type]
+            name=body.name,
+            provider=body.provider,
+            model=body.model,
+            api_key=body.api_key,
+        )
+        await session.commit()
+        return _to_response(config)
+    except ValueError as exc:
+        raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(exc)) from exc
```
hamza-56, line 51: CLAUDE.md exception-handling rule asks for a log on every except block, including when re-raising. The `SQLAlchemyError` branch right below already does this. Same applies at lines 114, 140, and 164.

app/api/v1/llm.py
```diff
@@ -0,0 +1,193 @@
...
+            config_id=config_id,
+            name=body.name,
+            model=body.model,
+        )
+        await session.commit()
+        return _to_response(config)
+    except LLMConfigValidationError as exc:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(exc)) from exc
+    except ValueError as exc:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(exc)) from exc
```
hamza-56, line 101: Two distinct service failures both surface as plain `ValueError` here and both map to 404: config not found (correct), and duplicate name from `service.update()` on `IntegrityError` (should be 409, matching `create_llm_config` above). PATCHing a config's name to an existing one currently returns `404 "An LLM config with name 'X' already exists for this user."`

app/llm/service.py
```diff
@@ -98,6 +101,11 @@ async def update(
         if name is not None:
             config.name = name
         if model is not None:
+            allowed = get_models_for_provider(config.provider)
+            if model not in allowed:
+                raise LLMConfigValidationError(
```
hamza-56, line 113: This raises the typed `LLMConfigValidationError`, but the not-found check earlier in this method (line 100) and at lines 131 and 166 raise plain `ValueError`. That mix is why the route at `app/api/v1/llm.py:114-115` collapses different conditions onto 404, including duplicate-name on PATCH (which should be 409, matching `create_llm_config`).

app/llm/service.py
```diff
@@ -144,28 +152,49 @@ async def delete(self, session: AsyncSession, user_id: int, config_id: int) -> b
...
+        """Return (config, decrypted_api_key). Caches by (user_id, config_id). Updates last_used_at."""
+        config = await self.get(session, user_id, config_id)
+        if config is None:
+            raise LLMConfigNotFoundError(f"LLMConfig {config_id} not found for user {user_id}")
+        if not config.model:
+            raise LLMConfigModelNotSetError(
+                f"LLMConfig {config_id} has no model set. Update it via PATCH /api/v1/llm/configs/{config_id} before use."
+            )
+        if not config.is_active:
+            raise LLMConfigNotFoundError(f"LLMConfig {config_id} is inactive")
```
hamza-56: An inactive config is a precondition failure, not a missing resource. Re-using `LLMConfigNotFoundError` here means the chat route maps this to 404, which hides the actual reason from the client.

app/core_plugins/chat/routes.py
```diff
@@ -18,9 +18,8 @@
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.api.v1.auth import get_current_user
-from app.core.cache import get_cache_service
+from app.api.v1.llm import get_llm_service
```
hamza-56: A plugin module importing a FastAPI dependency factory from `app.api.v1.llm` couples the plugin layer to the API layer.

app/core_plugins/chat/routes.py
```diff
@@ -411,20 +282,23 @@ async def chat_completion(
...
-        raise HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND,
-            detail=f"No API key found for provider: {request.provider}",
+    try:
+        llm_config, api_key = await llm_service.resolve(
+            session=session,
+            user_id=current_user.id,  # type: ignore[arg-type]
+            config_id=request.llm_config_id,
         )
+    except LLMConfigNotFoundError as exc:
```
hamza-56, line 299: Same CLAUDE.md log requirement as the LLM route handlers: log before re-raising.

app/llm/plugin_adapter.py
```diff
@@ -0,0 +1,84 @@
...
+        *,
+        session: AsyncSession,
+        user_id: int,
+        incoming_config: dict[str, Any],
+    ) -> dict[str, Any]:
+        config_id = incoming_config.get("llm_config_id")
+        if config_id is None:
+            return incoming_config
+
+        result = await session.exec(
```
hamza-56: `preprocess_config` and `postprocess_config` issue the same `select(LLMConfig).where(...)` with the same four predicates. The `{"llm_config_name": None, "llm_provider": None, "llm_model": None}` literal is also duplicated in `postprocess_config` (lines 53 and 66).

hamza-56 (approved):

PR #285, opened 2026-04-27
feat(api): allow multi-file attachment in chat plugin

abdulrafey1:
## What
Fixes five interconnected bugs in the chat attachment system where user messages were duplicated or dropped in API payloads, non-drive attachments incorrectly persisted after send, the Drive picker duplicated files on re-open, and a Pill test was verifying the wrong behaviour.

## Changes
- fix(frontend): append user message only to last attachment block in useChatStream
- fix(frontend): prepend user message to plain-text attachment content with separator
- fix(frontend): retain only drive file attachments after send in useChatInput
- fix(frontend): replace drive attachments instead of appending on picker re-open
- fix(frontend): seed DriveFilePicker selectedFilesMap from initialSelectedFiles prop
- fix(frontend): narrow DriveFilePicker prop from selectedFileIds to initialSelectedFiles
- fix(frontend): pass SelectedDriveFile[] to DriveFilePicker from ChatInput
- test(frontend): correct Pill per-file removal test to assert specific driveFileDbId

## How to Test
1. Open a chat session.
2. Attach two Drive files and type a message — send it. Verify the message appears once in the conversation, not once per file.
3. Attach a .txt file and type a message — send it. Verify the message appears above the file content separated by `---`, not dropped.
4. Attach a .txt file and a Drive file together, then send. Verify the .txt attachment is cleared after send while the Drive file attachment persists.
5. Open the Drive picker, select two files, close without confirming. Re-open — verify the previously selected files are still checked without duplicates.
6. Run the Pill test suite: `cd frontend && bun test __tests__/plugins/chat/components/attachment/Pill.test.tsx`

## Notes
**Known limitation — multi-file drive attachment persistence on page reload is deferred.**
The backend API currently returns only a single `active_drive_file_id` / `active_drive_file_name` even though the `conversation_active_drive_files` junction table supports multiple rows. The frontend is already prepared to consume a list; the fix requires the API to return `active_drive_files: [{id, name}]`. No migration is needed on the frontend side; this is tracked as a backend follow-up.

No breaking changes. No new environment variables. No migration required for this PR.


frontend/plugins/chat/hooks/useChatStream.ts
```diff
@@ -16,49 +16,52 @@ interface UseChatStreamOptions {
...
+      ];
+      if (message.trim()) contentBlocks.push({ type: "text", text: message });
       out.push({
         role: "user",
-        content: attachment.text,
+        content: contentBlocks,
         attachment: { name: attachment.name, size: attachment.size },
       });
+    } else {
+      if (attachment.text) {
```
hamza-56, line 55: if a user uploads a .txt and also types a question, the question never gets sent. the text-attachment branch fills `out` with just the file content, so the trailing `out.length === 0` fallback below never fires.

frontend/plugins/chat/hooks/useChatInput.ts
```diff
@@ -47,29 +50,35 @@ export function useChatInput({ token, attachment, setAttachment, onSend }: UseCh
...
       message: message.trim(),
-      attachment,
+      attachments,
     });
 
     setMessage("");
     // Drive file attachments persist across the session until explicitly removed
-    if (!attachment?.driveFileDbId) {
-      setAttachment(null);
+    const hasDriveFiles = attachments.some((a) => a.driveFileDbId);
```
hamza-56: if a non-drive attachment is mixed with a drive file, `!hasDriveFiles` skips the clear — so the non-drive file persists across sends and gets re-sent on every following message until manually removed.

frontend/plugins/chat/components/messages/UserMessage.tsx
```diff
@@ -8,16 +8,15 @@ interface UserMessageProps {
...
+  const openPreview = (attachment: TextAttachment) => {
+    setPreviewAttachment(attachment);
     setPreviewOpen(true);
   };
 
   return (
     <div className="flex justify-end">
       <div className="max-w-[75%] space-y-1">
-        <Pill attachment={message.attachment ?? null} onOpen={openPreview} />
+        <Pill attachments={message.attachments ?? []} onPreview={openPreview} onRemove={() => {}} />
```
hamza-56: `Pill.onRemove` used to be optional and the X was gated on it being defined. it's required now, so this passes `() => {}` to keep typescript happy — but Pill always renders the X, leaving a dead close button on every sent-message pill. same thing in AssistantMessage.tsx for `pillAttachment`.

frontend/plugins/chat/hooks/useConversation.ts
```diff
@@ -110,14 +110,16 @@ export function useConversation(
 
         // Restore persistent drive file attachment (or clear if this conversation has none)
         if (data.active_drive_file_id && data.active_drive_file_name) {
-          setInputAttachment({
-            name: data.active_drive_file_name,
-            size: 0,
-            text: `[File: ${data.active_drive_file_name}]`,
-            driveFileDbId: data.active_drive_file_id,
-          });
+          setInputAttachments([
```
hamza-56, line 109: related: the hydration block above (around lines 94-104, untouched in this PR) still writes `attachment:` (singular) onto loaded messages, but `ChatMessage` was widened to `attachments?: TextAttachment[]` (plural). UserMessage/AssistantMessage read the plural form, so loaded history messages end up rendering with no pill. typescript misses it — the singular field is just an excess property on the object literal.

app/migrations/versions/acc6f4381f88_fix_unexpected_schema_differences.py
```diff
@@ -5,81 +5,120 @@
 Create Date: 2026-04-17 15:04:32.731375
 
 """
+
 from typing import Sequence, Union
```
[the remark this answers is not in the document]
abdulrafey1: @claude this is intentional

hamza-56 (approved):

PR #294, opened 2026-04-29
chore(frontend): proxy /api and /ai to backend in dev mode

hamza-56:
## Summary
- Run the Next dev server in dev mode (no static export) with rewrites that forward `/api/*` and `/ai/*` to `localhost:7727`, so the frontend can call the backend during local development without CORS or hardcoded URLs
- Production build behaviour is unchanged — `output: "export"` still produces the static export consumed by the Docker image

## Test plan
- [ ] Run `make frontend` and confirm requests to `/api/*` reach the local FastAPI on `:7727`
- [ ] Run `make frontend.build` and confirm the static export is still produced under `frontend/out/`
- [ ] Run `make up` and confirm the Docker production flow is unchanged

### DIFF
--- a/frontend/next.config.ts
+++ b/frontend/next.config.ts
@@ -1,8 +1,20 @@
 import type { NextConfig } from "next";
 
-const nextConfig: NextConfig = {
-  output: "export",
-  trailingSlash: true,
-};
+const isDev = process.env.NODE_ENV === "development";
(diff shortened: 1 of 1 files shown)


hamza-56:
Not needed; replacing with a separate frontend PR.

PR #298, opened 2026-04-30
feat(rag): replace vector similarity search with agentic RAG

abdulrafey1:
# [[Sparkth] Replace semantic RAG search with agentic RAG search #301](https://github.com/edly-io/sparkth/issues/301)

## What

Replaces the two-phase semantic similarity RAG (embed query → pgvector search) with a LangGraph ReAct agent that inspects document structure via an MCP server and hand-picks the exact sections to retrieve. Chunks are then fetched directly by section keys — no vector similarity involved.

## Why

Semantic similarity search was finding semantically close but contextually wrong chunks. The agent approach lets the model reason about the document's full structure, understand positional references ("second half", "last chapter"), and select sections with the same intent-awareness as a human reader.

### New: RAG MCP server (`app/rag_mcp/`)
- FastMCP server exposing metadata-only tools: `get_document_structure`, `search_section_by_keyword`, `list_file_sections`, `get_chunk_stats`, `get_file_metadata`, `list_user_files`
- `get_document_structure` returns sections ordered by document position with `chunk_count` and `position_index` for positional reasoning
- Added as a new `rag-mcp` service in `docker-compose.yml`

### New: LangGraph agent (`app/rag/agent.py`)
- ReAct agent connects to the MCP server via `langchain-mcp-adapters`
- Uses the new non-context-manager API (`client.get_tools()`) required by `langchain-mcp-adapters 0.1.0`
- Returns `AgentSearchDecision(source_name, selected_sections)` with exact section keys

### Updated: RAG context service (`app/rag/context_service.py`, `app/rag/store.py`)
- `get_context_via_agent()` replaces the old two-phase method
- `fetch_chunks_by_sections()` does NULL-safe exact matching on `(chapter, section, subsection)` tuples, ordered by chunk `id` (document order)

### Updated: Streaming chat path (`app/core_plugins/chat/routes.py`)
- Streaming path now calls `get_context_via_agent` instead of the old rank+similarity flow
- Fixed `provider._llm` (always `None`) to `provider._create_llm()` — this was causing `AttributeError: 'NoneType' object has no attribute 'bind_tools'`
- `rag_sections` now included in the final SSE `done` payload so the frontend can display them

### Updated: Frontend (`frontend/plugins/chat/hooks/useChatStream.ts`)
- Reads `rag_sections` from the `done` event and populates `message.ragSections` so `AssistantMessage.tsx` renders the "Taking into context" section list

## Test plan

- [ ] Send a chat message with a Drive file attached — confirm the agent calls the MCP server and returns chunks
- [ ] Verify the "Taking into context" section list appears above the response bubble after the stream completes
- [ ] Send a positional query ("summarise the second half") — confirm the agent reasons about `position_index`
- [ ] Send a message without any attachment — confirm no RAG path is triggered
- [ ] Run `make test` — all 824 tests pass

## Migration / breaking changes

None. The `rag-mcp` service must be running (included in `docker-compose.yml`). Requires `RAG_MCP_URL` env var (added to `.env.example`).


app/core/config.py
```diff
@@ -35,6 +35,7 @@ class Settings(BaseSettings):
     RAG_EMBEDDING_MODEL: str = "sentence-transformers/all-MiniLM-L6-v2"  # model name passed to provider
     RAG_DISPLAY_NAME_MAX_CHARS: int = 30  # max chars for section/filename labels in the UI
     MEMORY_PROFILING_ENABLED: bool = False
+    RAG_MCP_URL: str  # = "http://rag-mcp:7728/mcp"
```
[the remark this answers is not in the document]
abdulrafey1: @claude this is intentional, we want the environment to crash if this env is not set.
The crash on a missing env is easier to diagnose and fix than a faulty MCP URL causing unexpected behaviour

app/rag/store.py
```diff
@@ -174,6 +175,49 @@ async def get_distinct_sections(
...
+        stmt = (
+            select(DocumentChunk)
+            .where(
+                col(DocumentChunk.user_id) == user_id,
+                col(DocumentChunk.source_name) == source_name,
+                or_(*[_key_condition(k) for k in section_keys]),
+            )
+            .order_by(col(DocumentChunk.id))
+            .limit(limit)
+        )
```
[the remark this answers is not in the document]
abdulrafey1: @claude this is noted in the codebase in a comment, and will be addressed in a future implementation

app/rag/agent.py
```diff
@@ -0,0 +1,133 @@
...
+    Raises:
+        RAGRetrievalError: If MCP client connection or agent invocation fails
+    """
+    rag_mcp_url = get_settings().RAG_MCP_URL
+
+    try:
+        client = MultiServerMCPClient(
+            {"rag-meta": {"url": rag_mcp_url, "transport": "http"}}  # type: ignore[misc, dict-item]
+        )
+        tools = await client.get_tools()
```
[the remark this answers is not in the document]
abdulrafey1: @claude `MultiServerMCPClient` does not support async context

app/rag/agent.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Raises:
+        RAGRetrievalError: If MCP client connection or agent invocation fails
+    """
+    rag_mcp_url = get_settings().RAG_MCP_URL
+
+    try:
+        client = MultiServerMCPClient(
+            {"rag-meta": {"url": rag_mcp_url, "transport": "http"}}  # type: ignore[misc, dict-item]
+        )
```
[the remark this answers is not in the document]
abdulrafey1: @claude this is not relevant, here's exactly why, from the library source:

`__init__` opens nothing `client.py:107-112`: it only stores the connection config dict in `self.connections`. No HTTP connections are established at construction time.

`get_tools()` manages sessions internally `client.py:152`: it calls `load_mcp_tools(None, connection=..., ...)` — passing `None` as the session and letting `load_mcp_tools` create and tear down a session per call. The docstring even says it explicitly: "A new session will be created for each tool call."

`async with` was intentionally removed `client.py:33-42`: `__aenter__` raises `NotImplementedError` with the message `"As of langchain-mcp-adapters 0.1.0, MultiServerMCPClient cannot be used as a context manager"`. The library authors deliberately moved away from a persistent-connection model.

There is no `aclose()` on the client — the library doesn't expose one because there's nothing to close on the client object itself.

So the reviewer's `try/finally`: `await client.aclose()` suggestion would throw `AttributeError` if attempted. The current usage is exactly the pattern the library intends.

app/rag/types.py
```diff
@@ -36,3 +40,56 @@ def to_dict(self) -> dict[str, str | None]:
...
     metadata: ChunkMetadata
+
+
+@dataclass
+class RAGContext:
+    """Retrieved context ready for injection into an LLM prompt."""
+
+    file_db_id: int
+    source_name: str
+    chunks: list[SimilarityResult]
```
hamza-56, line 59: `SimilarityResult` is referenced here but defined further down in the same file. Reads a bit out of order.

app/rag_mcp/tools.py
```diff
@@ -0,0 +1,245 @@
...
+                    DriveFile.user_id == user_id,
+                    DriveFile.is_deleted == False,  # noqa: E712
+                )
+            )
+            file = file_result.scalars().first()
+            if not file:
+                return []
+
+            source_name = resolve_source_name(file)
+            keyword_safe = keyword.replace("%", r"\%").replace("_", r"\_")
```
hamza-56: Looks like backslashes are not escaped here, only `%` and `_`. If a keyword contains a literal `\`, Postgres rejects the resulting LIKE pattern with error 22025.

app/rag/agent.py
```diff
@@ -0,0 +1,132 @@
...
+        RAGSearchAgentResponse with source_name and selected_sections
+
+    Raises:
+        RAGRetrievalError: If MCP client connection or agent invocation fails
+    """
+    rag_mcp_url = get_settings().RAG_MCP_URL
+
+    try:
+        client = MultiServerMCPClient(
+            {"rag-meta": {"url": rag_mcp_url, "transport": "http"}}  # type: ignore[misc, dict-item]
```
hamza-56: The `# type: ignore` here feels like it could be avoided pretty easily.

app/rag/agent.py
```diff
@@ -0,0 +1,132 @@
...
+        result = await agent.ainvoke({"messages": [system_message, human_message]})
+        decision: RAGSearchAgentResponse | None = result.get("structured_response")
+        if decision is None:
+            logger.warning("Agent produced no structured_response for user %d, file %d", user_id, file_id)
+            return RAGSearchAgentResponse(source_name="", selected_sections=[])
+        return decision
+
+    except ValidationError as exc:
+        logger.warning("Agent response validation failed for user %d, file %d: %s", user_id, file_id, exc)
+        return RAGSearchAgentResponse(source_name="", selected_sections=[])
```
hamza-56: When validation fails this returns an empty response, which from the caller perspective looks identical to the agent legitimately deciding no sections are relevant. Makes downstream debugging tricky.

app/rag/agent.py
```diff
@@ -0,0 +1,132 @@
...
+    except (ConnectError, HTTPStatusError) as exc:
+        logger.exception("MCP client connection error for user %s, file %s", user_id, file_id)
+        raise RAGRetrievalError(f"Failed to connect to RAG metadata server: {exc}") from exc
+    except LangChainException as exc:
+        logger.exception("Agent invocation error for user %s, file %s", user_id, file_id)
+        raise RAGRetrievalError(f"Agent failed to process query: {exc}") from exc
+    except ValueError as exc:
+        logger.exception("Error parsing agent response for user %s, file %s", user_id, file_id)
+        raise RAGRetrievalError(f"Invalid agent response format: {exc}") from exc
```
hamza-56: Plain `ValueError` is pretty broad here, and it is not obvious which call inside the try block is actually meant to raise it.

app/core_plugins/chat/routes.py
```diff
@@ -60,12 +59,18 @@
...
-from app.rag.context_service import RAGContextService, format_chunks_as_context
+from app.rag.context_service import RAGContextService
 from app.rag.exceptions import DriveFileNotFoundError, RAGNotReadyError, RAGRetrievalError
 from app.rag.provider import get_provider as get_rag_provider
+from app.rag.types import RAGContext
+from app.rag.utils import get_asset
 
 logger = get_logger(__name__)
 
+_RAG_CONTEXT_PROMPT = get_asset("rag_context_replacement_prompt", "txt")
```
hamza-56, line 70: Doing a disk read at import time means any test that imports this module needs the asset on disk, even ones unrelated to RAG.

app/rag/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+def get_asset(file_name: str, file_extension: str) -> str | dict[str, Any]:
+    """Read a file from app/rag/assets and return its contents.
+
+    Returns a str for .txt files and a dict for .json files.
+    """
+    rag_assets_dir = Path(__file__).parent / "assets"
+    path = rag_assets_dir / f"{file_name}.{file_extension.lstrip('.')}"
+
+    with open(path) as f:
```
hamza-56: `open(path)` here, but `path` is already a `Path`. Also `lstrip(".")` ends up computed three times in this function.

tests/rag_mcp/test_tools.py
```diff
@@ -0,0 +1,301 @@
...
+            mock_get_session.return_value.__aenter__.return_value = mock_session
+
+            result = await get_chunk_stats(user_id=1, file_id=1)
+
+            assert result is not None
+            assert result.chunk_count == 42
+            assert result.avg_token_count == 128.5
+
+
+class TestSearchSectionByKeyword:
```
hamza-56, line 322: Noticed there are no tests for `get_document_structure` in this file, even though it is the tool the agent leans on most heavily.

app/core_plugins/chat/routes.py
abdulrafey1: The concern is valid in general, but not a problem in this specific case. The asset file (rag_context_replacement_prompt.txt) is checked into the repository under app/rag/assets/, so it's always present wherever the source tree is — including CI. Tests that import this module (e.g. test_provider_model.py, test_persistent_attachment.py) collect and run fine without any mocking or asset setup.

The pattern would be a real issue if the asset were generated at build time, fetched from a remote, or only present in certain environments. Since it's a static file committed alongside the code, the import-time read is essentially as safe as importing a constant. Happy to revisit if that ever changes.

hamza-56 (approved):

PR #302, opened 2026-05-04
Add email verification on registration

### LINKED ISSUE #289: [Sparkth] Email Verification
mplement a user registration flow that creates a new user account, sends an email with a verification link, and marks the account as verified upon link activation. 

Acceptance Criteria
- [ ]  Registration form accepts email, password, and optional full name
- [ ]  Password validation enforces minimum 8 characters, at least one uppercase letter, one number, and one special character
- [ ]  System generates a unique, time-limited verification token (valid for 24 hours)
- [ ]  Verification email is sent asynchronously (non-blocking) to the user's email address
- [ ]  Email contains a clickable verification link with embedded token
- [ ]  Clicking the link marks the user account as verified and expires the token
- [ ]  Attempting to use an expired token shows a user-friendly error with resend option
- [ ]  Verified users can log in; unverified users receive a prompt to verify before access
- [ ]  Resend verification email flow works and generates a new token, invalidating the old one
- [ ]  Rate limiting prevents email spam (e.g., max 3 verification emails per hour per user)

hamza-56:
## What

Adds email confirmation on account registration. New email/password signups create unverified accounts; users must click a tokenized link before they can log in. Existing users are auto-verified via migration backfill, and Google OAuth signups are auto-verified on creation/link. Also enforces password complexity at registration.

## Summary

**Backend**
- New `email_verification_token` table; sha256-hashed tokens (raw never persisted), 24h expiry, single-use, prior-token invalidation on resend.
- `EmailVerificationService` mints / verifies / resends; `send_verification_email` composes a multipart text + HTML body (HTML-escaped name to prevent injection).
- `app/core/email.py` wraps `aiosmtplib` — works with Amazon SES SMTP, Mailgun, Gmail, MailHog locally.
- Endpoints: `POST /auth/verify-email` (400 invalid/expired/used), `POST /auth/verify-email/resend` (Redis-cooldowned via `SET NX EX`, async-context-manager lifecycle).
- `POST /auth/register` schedules the email via `BackgroundTasks` (atomic flush + single commit).
- `POST /auth/login` returns `403 {code:"email_not_verified", email}` for unverified accounts (only reachable post-password-verification).
- `validate_password_complexity` in `app/core/security.py` enforces ≥8 chars + uppercase + digit + special; runs via `UserCreate.field_validator`.
- Email is normalized (`strip().lower()`) at register and Google OAuth callback to prevent case-sensitive lookup mismatches and duplicate accounts.
- Alembic migration adds the schema and backfills existing users as verified (cross-DB safe via `sa.func.now()`).

**Frontend**
- New `/verify-email` page with success / expired / invalid states and a resend form.
- Login form renders an inline alert + resend button when login returns the unverified 403 (idle / sending / sent / rate-limited).
- Register form shows "check your email — we've sent a confirmation link to {email}" instead of auto-redirecting; password input shows the complexity rules as inline hint text.

## How to Test

1. `make test` — 45 new/updated tests cover token lifecycle, all four endpoints, Google OAuth verification, SMTP composition, resend rate limiting, and password complexity.
2. `make frontend.build` — clean Next.js build with `/verify-email` route generated.
3. End-to-end with MailHog: register with whitelisted email → "check your email" → click link from MailHog → "Email confirmed" → log in → success.
4. Try logging in unverified → resend CTA → resend → second email → resend immediately → "Please wait before resending" (429).
5. Submit weak passwords (`abc`, `alllowercase1!`, `NoDigits!!`, `NoSpecial1A`) → 422 with field-level error; submit `Sup3rSecret!` → 200.
6. Sign up via Google OAuth → confirm `email_verified=true` on the resulting user row.
7. Apply the migration on a non-empty user table (Postgres + SQLite) → confirm existing rows are marked verified.

Closes #289

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,14 +78,13 @@ Copy `.env.example` → `.env`. Required variables:
 | `DATABASE_URL` | PostgreSQL connection string |
 | `SECRET_KEY` | JWT signing key |
 | `LLM_ENCRYPTION_KEY` | Fernet key for encrypting stored LLM API keys |
-| `REDIS_URL` | Redis for session caching, used in chat plugin |
+| `REDIS_URL` | Redis for chat session caching and the email-verification resend rate-limit bucket |
 | `GOOGLE_CLIENT_ID/SECRET` | Google OAuth |
 | `SMTP_HOST` / `SMTP_PORT` / `SMTP_USERNAME` / `SMTP_PASSWORD` / `SMTP_USE_TLS` | Outbound SMTP (Amazon SES, Mailgun, MailHog, …) |
 | `SMTP_FROM_EMAIL` / `SMTP_FROM_NAME` | From-header for verification + other transactional emails |
 | `EMAIL_VERIFICATION_TOKEN_TTL_HOURS` | Lifetime of an email-verification token (default 24) |
 | `EMAIL_VERIFICATION_RESEND_COOLDOWN_SECONDS` | Per-email cooldown on the resend endpoint (default 60) |
 | `FRONTEND_BASE_URL` | Base URL used in verification email links |
-| `REDIS_URL` | Redis used for the resend rate-limit bucket |
 
 CI uses `DATABASE_URL=sqlite+aiosqlite:///./test.db`. Tests always run against SQLite.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -80,6 +80,12 @@ Copy `.env.example` → `.env`. Required variables:
 | `LLM_ENCRYPTION_KEY` | Fernet key for encrypting stored LLM API keys |
 | `REDIS_URL` | Redis for session caching, used in chat plugin |
 | `GOOGLE_CLIENT_ID/SECRET` | Google OAuth |
+| `SMTP_HOST` / `SMTP_PORT` / `SMTP_USERNAME` / `SMTP_PASSWORD` / `SMTP_USE_TLS` | Outbound SMTP (Amazon SES, Mailgun, MailHog, …) |
+| `SMTP_FROM_EMAIL` / `SMTP_FROM_NAME` | From-header for verification + other transactional emails |
+| `EMAIL_VERIFICATION_TOKEN_TTL_HOURS` | Lifetime of an email-verification token (default 24) |
+| `EMAIL_VERIFICATION_RESEND_COOLDOWN_SECONDS` | Per-email cooldown on the resend endpoint (default 60) |
+| `FRONTEND_BASE_URL` | Base URL used in verification email links |
+| `REDIS_URL` | Redis used for the resend rate-limit bucket |
 
 CI uses `DATABASE_URL=sqlite+aiosqlite:///./test.db`. Tests always run against SQLite.
 
--- a/app/api/v1/auth.py
+++ b/app/api/v1/auth.py
@@ -2,7 +2,7 @@ from datetime import datetime, timedelta
 from urllib.parse import quote
 
 import jwt
-from fastapi import APIRouter, Depends, HTTPException, Query, status
+from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, Query, status
 from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 from sqlmodel import select
@@ -20,6 +20,10 @@ from app.models.base import utc_now
 from app.models.user import User
 from app.schemas import GoogleAuthUrl, Token, UserCreate, UserLogin
 from app.schemas import User as UserSchema
+from app.services.email_verification import (
+    EmailVerificationService,
+    send_verification_email,
+)
 from app.services.whitelist import WhitelistService
 
 settings = get_settings()
@@ -63,7 +67,11 @@ async def get_current_user(
 
 
 @router.post("/register", response_model=UserSchema)
-async def register_user(user: UserCreate, session: AsyncSession = Depends(get_async_session)) -> User:
+async def register_user(
+    user: UserCreate,
+    background_tasks: BackgroundTasks,
+    session: AsyncSession = Depends(get_async_session),
+) -> User:
     if not settings.REGISTRATION_ENABLED:
         raise HTTPException(status_code=403, detail="Registration is currently disabled")
     if not await WhitelistService.is_email_allowed(session, user.email):
@@ -91,6 +99,22 @@ async def register_user(user: UserCreate, session: AsyncSession = Depends(get_as
     session.add(db_user)
     await session.commit()
     await session.refresh(db_user)
+
+    assert db_user.id is not None
+    user_id = db_user.id
+    user_email = db_user.email
+    user_name = db_user.name
+
+    raw_token = await EmailVerificationService.create_token(session, user_id=user_id)
+    await session.commit()
+    await session.refresh(db_user)
+
+    background_tasks.add_task(
+        send_verification_email,
+        to=user_email,
+        name=user_name,
+        raw_token=raw_token,
+    )
     return db_user
 
 
--- a/app/api/v1/auth.py
+++ b/app/api/v1/auth.py
@@ -217,6 +217,9 @@ async def google_callback(
             if user:
                 # Link Google ID to existing account
                 user.google_id = google_id
+                if not user.email_verified:
+                    user.email_verified = True
+                    user.email_verified_at = utc_now()
                 session.add(user)
                 await session.commit()
                 await session.refresh(user)
@@ -243,6 +246,8 @@ async def google_callback(
                     email=email,
                     google_id=google_id,
                     hashed_password=None,
+                    email_verified=True,
+                    email_verified_at=utc_now(),
                 )
                 session.add(user)
                 await session.commit()
--- a/app/api/v1/auth.py
+++ b/app/api/v1/auth.py
@@ -1,4 +1,6 @@
+import hashlib
 from datetime import datetime, timedelta
+from typing import Any
 from urllib.parse import quote
 
 import jwt
@@ -9,6 +11,7 @@ from sqlmodel import select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.core import security
+from app.core.cache import get_cache_service
 from app.core.config import get_settings
 from app.core.db import get_async_session
 from app.core.google_auth import (
@@ -20,6 +23,7 @@ from app.models.base import utc_now
 from app.models.user import User
 from app.schemas import (
     GoogleAuthUrl,
+    ResendVerificationRequest,
     Token,
     UserCreate,
     UserLogin,
@@ -273,6 +277,51 @@ async def verify_email(
     await session.commit()
     await session.refresh(user)
     return user
+
+
+async def _get_resend_redis() -> Any:
+    """Indirection so tests can override the rate-limit backend."""
+    cache = get_cache_service(settings.REDIS_URL)
+    await cache.connect()
+    return cache._redis  # noqa: SLF001
+
+
+@router.post("/verify-email/resend", status_code=status.HTTP_202_ACCEPTED)
+async def resend_verification_email(
+    body: ResendVerificationRequest,
+    background_tasks: BackgroundTasks,
+    session: AsyncSession = Depends(get_async_session),
+) -> dict[str, str]:
+    email_lower = body.email.lower()
+    key = f"email_verify_resend:{hashlib.sha256(email_lower.encode()).hexdigest()}"
+    redis = await _get_resend_redis()
+    accepted = await redis.set(
+        key,
+        "1",
+        ex=settings.EMAIL_VERIFICATION_RESEND_COOLDOWN_SECONDS,
+        nx=True,
+    )
+    if not accepted:
+        raise HTTPException(status_code=429, detail="rate_limited")
+
+    raw_token = await EmailVerificationService.request_resend(session, email=email_lower)
+    if raw_token is None:
+        return {}
+
+    await session.commit()
+
+    user_result = await session.exec(select(User).where(User.email == email_lower))
+    user = user_result.one()
+    user_email = user.email
+    user_name = user.name
+
+    background_tasks.add_task(
+        send_verification_email,
+        to=user_email,
+        name=user_name,
+        raw_token=raw_token,
+    )
+    return {}
 
 
 async def require_superuser(
--- a/app/api/v1/auth.py
+++ b/app/api/v1/auth.py
@@ -18,10 +18,18 @@ from app.core.google_auth import (
 )
 from app.models.base import utc_now
 from app.models.user import User
-from app.schemas import GoogleAuthUrl, Token, UserCreate, UserLogin
+from app.schemas import (
+    GoogleAuthUrl,
+    Token,
+    UserCreate,
+    UserLogin,
+    VerifyEmailRequest,
+)
 from app.schemas import User as UserSchema
 from app.services.email_verification import (
     EmailVerificationService,
+    TokenExpiredError,
+    TokenInvalidError,
     send_verification_email,
 )
 from app.services.whitelist import WhitelistService
@@ -249,6 +257,22 @@ async def google_callback(
     except ValueError as e:
         # Redirect to login page with error
         return RedirectResponse(url=f"/login?error={quote(str(e))}", status_code=302)
+
+
+@router.post("/verify-email", response_model=UserSchema)
+async def verify_email(
+    body: VerifyEmailRequest,
+    session: AsyncSession = Depends(get_async_session),
+) -> User:
+    try:
+        user = await EmailVerificationService.verify_token(session, raw_token=body.token)
+    except TokenExpiredError:
+        raise HTTPException(status_code=400, detail="expired_token")
+    except TokenInvalidError:
+        raise HTTPException(status_code=400, detail="invalid_token")
+    await session.commit()
+    await session.refresh(user)
+    return user
 
 
 async def require_superuser(
--- a/app/api/v1/auth.py
+++ b/app/api/v1/auth.py
@@ -1,4 +1,6 @@
 import hashlib
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
 from datetime import datetime, timedelta
 from typing import Any
 from urllib.parse import quote
@@ -281,21 +283,26 @@ async def verify_email(
     return user
 
 
-async def _get_resend_redis() -> Any:
-    """Indirection so tests can override the rate-limit backend.
+@asynccontextmanager
+async def _get_resend_redis() -> AsyncIterator[Any]:
+    """Yield a Redis client for the resend cooldown; close on exit.
 
-    Returns a fresh Redis client. The caller is responsible for closing it
-    via `aclose()` (see the resend endpoint's try/finally). The endpoint
-    is rate-limited to roughly one request per email per cooldown window,
-    so per-request pool setup is fine and avoids the shared-mutable-state
-    cost of a long-lived singleton. Not routed through `CacheService` —
+    The endpoint is rate-limited to roughly one request per email per
+    cooldown window, so per-request client + pool setup is acceptable
+    and avoids any shared mutable state. Not routed through `CacheService` —
     that wraps a different API (string get/set with default ttl).
+
+    Tests override this with a fake context manager.
     """
-    return aioredis.from_url(  # type: ignore[no-untyped-call]
+    redis = aioredis.from_url(  # type: ignore[no-untyped-call]
         settings.REDIS_URL,
         encoding="utf-8",
         decode_responses=True,
     )
+    try:
+        yield redis
+    finally:
+        await redis.aclose()
 
 
 @router.post("/verify-email/resend", status_code=status.HTTP_202_ACCEPTED)
@@ -306,16 +313,13 @@ async def resend_verification_email(
 ) -> dict[str, str]:
     email_lower = body.email.lower()
     key = f"email_verify_resend:{hashlib.sha256(email_lower.encode()).hexdigest()}"
-    redis = await _get_resend_redis()
-    try:
+    async with _get_resend_redis() as redis:
         accepted = await redis.set(
             key,
             "1",
             ex=settings.EMAIL_VERIFICATION_RESEND_COOLDOWN_SECONDS,
             nx=True,
         )
-    finally:
-        await redis.aclose()
     if not accepted:
         raise HTTPException(status_code=429, detail="rate_limited")
 
--- a/app/api/v1/auth.py
+++ b/app/api/v1/auth.py
@@ -1,5 +1,6 @@
 import hashlib
 from datetime import datetime, timedelta
+from functools import lru_cache
 from typing import Any
 from urllib.parse import quote
 
@@ -281,18 +282,26 @@ async def verify_email(
     return user
 
 
-async def _get_resend_redis() -> Any:
-    """Indirection so tests can override the rate-limit backend.
+@lru_cache(maxsize=1)
+def _resend_redis_client() -> aioredis.Redis:
+    """Process-level singleton Redis client for the resend cooldown.
 
-    Returns a Redis client used only for the resend cooldown (`SET NX EX`).
-    Not routed through `CacheService` because that wraps a different API
-    (string get/set with ttl) and we'd be reaching into private state.
+    `aioredis.from_url` returns immediately; the connection pool is
+    initialized lazily on first command, so the cached client is safe to
+    build at import-adjacent time. Not routed through `CacheService` —
+    that wraps a different API (string get/set with default ttl) and
+    we'd be reaching into private state.
     """
-    return aioredis.from_url(  # type: ignore[no-untyped-call]
+    return aioredis.from_url(  # type: ignore[no-untyped-call,no-any-return]
         settings.REDIS_URL,
         encoding="utf-8",
         decode_responses=True,
     )
+
+
+async def _get_resend_redis() -> Any:
+    """Indirection so tests can override the rate-limit backend."""
+    return _resend_redis_client()
 
 
 @router.post("/verify-email/resend", status_code=status.HTTP_202_ACCEPTED)
--- a/app/api/v1/auth.py
+++ b/app/api/v1/auth.py
@@ -4,6 +4,7 @@ from typing import Any
 from urllib.parse import quote
 
 import jwt
+import redis.asyncio as aioredis
 from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, Query, status
 from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
@@ -11,7 +12,6 @@ from sqlmodel import select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.core import security
-from app.core.cache import get_cache_service
 from app.core.config import get_settings
 from app.core.db import get_async_session
 from app.core.google_auth import (
@@ -86,7 +86,9 @@ async def register_user(
 ) -> User:
     if not settings.REGISTRATION_ENABLED:
         raise HTTPException(status_code=403, detail="Registration is currently disabled")
-    if not await WhitelistService.is_email_allowed(session, user.email):
+
+    normalized_email = user.email.strip().lower()
+    if not await WhitelistService.is_email_allowed(session, normalized_email):
         raise HTTPException(
             status_code=status.HTTP_403_FORBIDDEN,
             detail="This email address is not authorized to register. Contact an administrator.",
@@ -96,7 +98,7 @@ async def register_user(
     if db_user:
         raise HTTPException(status_code=400, detail="Username already registered")
 
-    result = await session.exec(select(User).where(User.email == user.email))
+    result = await session.exec(select(User).where(User.email == normalized_email))
     db_user_email = result.one_or_none()
     if db_user_email:
         raise HTTPException(status_code=400, detail="Email already registered")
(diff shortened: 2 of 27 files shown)


app/services/email_verification.py
```diff
@@ -0,0 +1,134 @@
...
+        f"This link expires in {ttl} hours.\n\n"
+        "If you didn't sign up for Sparkth, you can ignore this email.\n"
+    )
+    html = (
+        f"<p>Hi {name},</p>"
+        "<p>Click the link below to confirm your email address:</p>"
+        f'<p><a href="{url}">Confirm my email</a></p>'
+        f"<p>This link expires in {ttl} hours.</p>"
+        "<p>If you didn't sign up for Sparkth, you can ignore this email.</p>"
+    )
```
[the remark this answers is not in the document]
hamza-56: Fixed in 263ead3. `name` is now passed through `html.escape()` and the URL through `html.escape(quote=True)`. Added `test_escapes_name_in_html_body` that registers with `<script>alert("xss")</script><a href="http://evil">click</a>` as the name and asserts the raw tags are absent / escaped form is present in the HTML body.

app/api/v1/auth.py
```diff
@@ -221,6 +268,67 @@ async def google_callback(
...
+    return cache._redis  # noqa: SLF001
+
+
+@router.post("/verify-email/resend", status_code=status.HTTP_202_ACCEPTED)
+async def resend_verification_email(
+    body: ResendVerificationRequest,
+    background_tasks: BackgroundTasks,
+    session: AsyncSession = Depends(get_async_session),
+) -> dict[str, str]:
+    email_lower = body.email.lower()
```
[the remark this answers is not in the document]
hamza-56: Fixed in 55d29dd via option 1: registration and the Google OAuth callback both now `email.strip().lower()` before storing, so resend (which already lowercases) lines up. This also closes a duplicate-account door where `Alice@Example.com` and `alice@example.com` could both register. Added `test_email_is_normalized_to_lowercase` to lock the behavior in.

app/api/v1/auth.py
```diff
@@ -91,6 +111,22 @@ async def register_user(user: UserCreate, session: AsyncSession = Depends(get_as
...
     await session.refresh(db_user)
+
+    assert db_user.id is not None
+    user_id = db_user.id
+    user_email = db_user.email
+    user_name = db_user.name
+
+    raw_token = await EmailVerificationService.create_token(session, user_id=user_id)
+    await session.commit()
+    await session.refresh(db_user)
```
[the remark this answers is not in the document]
hamza-56: Fixed in 55d29dd. `register_user` now does `flush()` to populate `db_user.id`, then `create_token`, then a single `commit()` — user + token are committed atomically. Dropped the redundant intermediate `refresh()` along the way.

app/api/v1/auth.py
```diff
@@ -221,6 +268,67 @@ async def google_callback(
...
+    await session.commit()
+    await session.refresh(user)
+    return user
+
+
+async def _get_resend_redis() -> Any:
+    """Indirection so tests can override the rate-limit backend."""
+    cache = get_cache_service(settings.REDIS_URL)
+    await cache.connect()
+    return cache._redis  # noqa: SLF001
```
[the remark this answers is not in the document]
hamza-56: Fixed in 55d29dd. `_get_resend_redis` now calls `redis.asyncio.from_url(settings.REDIS_URL, ...)` directly. CacheService wraps a different API (`get`/`set` with default TTL), and the resend cooldown only needs raw `SET NX EX`, so going through it added an indirection without a payoff.

app/api/v1/auth.py
```diff
@@ -221,6 +268,67 @@ async def google_callback(
...
+    if raw_token is None:
+        return {}
+
+    await session.commit()
+
+    user_result = await session.exec(select(User).where(User.email == email_lower))
+    user = user_result.one()
+    user_email = user.email
+    user_name = user.name
+
```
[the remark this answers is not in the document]
hamza-56: Fixed in 263ead3. `request_resend` now returns `tuple[str, str] | None` — `(raw_token, user.name)` — and the endpoint unpacks that directly. The post-commit `select(User)` is gone.

app/api/v1/auth.py
```diff
@@ -116,6 +152,12 @@ async def login_for_access_token(
             headers={"WWW-Authenticate": "Bearer"},
         )
 
+    if not user.email_verified:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail={"code": "email_not_verified", "email": user.email},
+        )
```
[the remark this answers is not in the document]
hamza-56: Keeping as-is. The 403 only fires *after* `verify_password` succeeds, so the only request that ever sees the email back is one that already has the correct password — i.e. the legitimate user. Username + correct password being a precondition makes this disclosure equivalent to "the user can see their own email," which is fine. Masking would also break the resend form (frontend uses `detail.email` to pre-fill) without a real security gain.

frontend/app/verify-email/page.tsx
```diff
@@ -0,0 +1,16 @@
...
+import VerifyEmailClient from "./VerifyEmailClient";
+
+export const metadata: Metadata = {
+  title: "Confirm your email | Sparkth",
+  description: "Verify your Sparkth account",
+};
+
+export default function VerifyEmailPage() {
+  return (
+    <Suspense>
```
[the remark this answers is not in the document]
hamza-56: Keeping as-is for consistency — `app/login/page.tsx` uses the same `<Suspense>` without a fallback, and matching that convention here avoids an inconsistency. Happy to make a follow-up that adds `fallback={null}` (or a real skeleton) to both pages at once if you prefer.

app/api/v1/auth.py
```diff
@@ -221,6 +265,70 @@ async def google_callback(
...
+) -> dict[str, str]:
+    email_lower = body.email.lower()
+    key = f"email_verify_resend:{hashlib.sha256(email_lower.encode()).hexdigest()}"
+    redis = await _get_resend_redis()
+    accepted = await redis.set(
+        key,
+        "1",
+        ex=settings.EMAIL_VERIFICATION_RESEND_COOLDOWN_SECONDS,
+        nx=True,
+    )
```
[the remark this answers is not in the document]
hamza-56: Fixed in 4067fa4. `_resend_redis_client` is now wrapped in `functools.lru_cache(maxsize=1)`, so we build one client per process instead of one per request. This matches the existing `get_cache_service` singleton pattern in `app/core/cache.py`. The connection pool is set up lazily on first command, so the cached client is cheap to keep around. `_get_resend_redis` stays as the test-override seam (tests patch it to return a `FakeRedis` shim, unchanged).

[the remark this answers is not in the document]
hamza-56: Switched to the try/finally + aclose() approach in d6b639c (per offline feedback that the singleton was overkill). The resend endpoint is rate-limited to ~1/email/cooldown, so per-request pool setup is negligible, and this removes the shared mutable module state from the previous fix. Added a no-op `aclose()` to the FakeRedis test shim to match the real client lifecycle.

app/core/security.py
```diff
@@ -19,6 +20,23 @@ def get_password_hash(password: str) -> str:
...
+
+    Rules: at least 8 chars, at least one uppercase letter, at least one
+    digit, and at least one non-alphanumeric character.
+    """
+    if (
+        len(password) < 8
+        or not re.search(r"[A-Z]", password)
+        or not re.search(r"\d", password)
+        or not re.search(r"[^A-Za-z0-9]", password)
+    ):
```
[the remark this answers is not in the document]
hamza-56: Keeping as-is. The frontend already shows the rules as inline hint text under the password input, so by the time a user submits a weak password they have the full ruleset visible. Splitting into per-rule messages would mean the user sees the same information in two places phrased two different ways — net negative for clarity. Happy to revisit if a UX requirement comes up that benefits from "you missed exactly *X*" feedback.

app/api/v1/auth.py
```diff
@@ -221,6 +265,76 @@ async def google_callback(
...
+            ex=settings.EMAIL_VERIFICATION_RESEND_COOLDOWN_SECONDS,
+            nx=True,
+        )
+    finally:
+        await redis.aclose()
+    if not accepted:
+        raise HTTPException(status_code=429, detail="rate_limited")
+
+    result = await EmailVerificationService.request_resend(session, email=email_lower)
+    if result is None:
```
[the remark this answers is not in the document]
hamza-56: Keeping as-is. With the 60s default cooldown the worst case is "user waits one extra minute," which is below the threshold worth a compensating delete + the failure-mode complexity that introduces (e.g. the delete itself can fail and now we have two paths leaking state). If we ever raise the cooldown to, say, an hour, this would be worth revisiting — flagging in the follow-up that addresses the rate-limit shape mentioned in #289.

app/core/security.py
```diff
@@ -19,6 +20,23 @@ def get_password_hash(password: str) -> str:
...
+    """Raise ValueError if `password` doesn't meet the complexity rules.
+
+    Rules: at least 8 chars, at least one uppercase letter, at least one
+    digit, and at least one non-alphanumeric character.
+    """
+    if (
+        len(password) < 8
+        or not re.search(r"[A-Z]", password)
+        or not re.search(r"\d", password)
+        or not re.search(r"[^A-Za-z0-9]", password)
```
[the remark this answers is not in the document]
hamza-56: Fixed in f74cd0c. `validate_password_complexity` now rejects any password longer than `MAX_PASSWORD_LENGTH = 128` chars before running the complexity regexes. Added a parametrized case (`"A1!" + "a"*200`) that hits the cap. Defense-in-depth even though we already have the registration whitelist + auth rate limits.

app/core/email.py
```diff
@@ -0,0 +1,39 @@
...
+    """Send a multipart (text + HTML) email via SMTP.
+
+    Raises RuntimeError if SMTP is not configured. Logs and re-raises
+    aiosmtplib.SMTPException on send failure.
+    """
+    if not settings.SMTP_HOST:
+        raise RuntimeError("SMTP not configured: set SMTP_HOST")
+
+    message = EmailMessage()
+    message["From"] = f'"{settings.SMTP_FROM_NAME}" <{settings.SMTP_FROM_EMAIL}>'
```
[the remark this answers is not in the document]
hamza-56: Fixed in f74cd0c. `From` is now built with `email.utils.formataddr((SMTP_FROM_NAME, SMTP_FROM_EMAIL))`, which handles RFC 5322 quoting and RFC 2047 encoding for non-ASCII names. Existing email-composition test still passes (the simple `"Test" <…>` case round-trips identically through `formataddr`).

app/services/email_verification.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        now = datetime.now(timezone.utc)
+
+        result = await session.exec(
+            select(EmailVerificationToken).where(
+                (EmailVerificationToken.user_id == user_id) & (EmailVerificationToken.used_at.is_(None))  # type: ignore[union-attr]
+            )
+        )
+        for prior in result.all():
+            prior.used_at = now
```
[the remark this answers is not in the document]
hamza-56: Keeping as-is. The reviewer notes this is non-blocking, and the design intent is "any token row not currently active." For audit, the workflow is to find the token whose `used_at` is closest to the user record's `email_verified_at` — they are written together inside `verify_token`. Adding an `invalidated_at` column or a status enum would be cheap if a future flow needs to distinguish (e.g. analytics on resend churn), but YAGNI today.

app/services/email_verification.py
```diff
@@ -0,0 +1,137 @@
...
+        result = await session.exec(select(User).where(User.email == email))
+        user = result.one_or_none()
+        if user is None or user.email_verified or user.id is None:
+            return None
+        token = await EmailVerificationService.create_token(session, user_id=user.id)
+        return token, user.name
+
+
+def _build_verify_url(raw_token: str) -> str:
+    return f"{settings.FRONTEND_BASE_URL}/verify-email?token={raw_token}"
```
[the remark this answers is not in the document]
hamza-56: Fixed in f74cd0c. `_build_verify_url` now does `settings.FRONTEND_BASE_URL.rstrip("/")` so configs ending in `/` no longer produce `//verify-email`. Added `test_strips_trailing_slash_in_frontend_base_url` to lock it.

app/migrations/versions/f1e914de8460_add_email_verification.py
```diff
@@ -0,0 +1,86 @@
...
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("token_hash"),
+    )
+    op.create_index(
+        op.f("ix_email_verification_token_user_id"),
+        "email_verification_token",
+        ["user_id"],
+        unique=False,
+    )
+    op.create_index(
```
abdulrafey1: Redundant index: `sa.UniqueConstraint("token_hash")` on line 44 already causes Postgres (and SQLite) to create a unique index on `token_hash`. This `op.create_index` call adds a second, non-unique index on the same column — redundant lookup overhead with no benefit.

The `user_id` index below is correct and needed. Only this block should be removed:
```python
op.create_index(
    op.f("ix_email_verification_token_token_hash"),
    "email_verification_token",
    ["token_hash"],
    unique=False,
)
```
And its matching `op.drop_index` in `downgrade()`.

app/services/email_verification.py
```diff
@@ -0,0 +1,137 @@
...
+        f"<p>Hi {html_escape(name)},</p>"
+        "<p>Click the link below to confirm your email address:</p>"
+        f'<p><a href="{html_escape(url, quote=True)}">Confirm my email</a></p>'
+        f"<p>This link expires in {ttl} hours.</p>"
+        "<p>If you didn't sign up for Sparkth, you can ignore this email.</p>"
+    )
+    return text, html
+
+
+async def send_verification_email(*, to: str, name: str, raw_token: str) -> None:
```
abdulrafey1, line 133: Background tasks swallow exceptions silently. If `SMTP_HOST` is blank (the default for local dev), `send_email` raises `RuntimeError("SMTP not configured")` — this propagates through the background task and is only logged at FastAPI's generic ERROR level with no context (user, email address). `aiosmtplib.SMTPException` is logged inside `send_email` but also re-raised, so it has the same problem here.

Wrap the call so failures are diagnosable:
```python
async def send_verification_email(*, to: str, name: str, raw_token: str) -> None:
    try:
        text_body, html_body = _render_email(name, raw_token)
        await send_email(
            to=to,
            subject="Confirm your Sparkth account",
            html_body=html_body,
            text_body=text_body,
        )
    except (RuntimeError, aiosmtplib.SMTPException):
        logger.exception("Failed to send verification email to %s", to)
```
The user already received a 201 by this point, so re-raising would serve no purpose — logging and absorbing is the right call here.

app/api/v1/auth.py
```diff
@@ -221,6 +267,78 @@ async def google_callback(
...
+        await redis.aclose()
+
+
+@router.post("/verify-email/resend", status_code=status.HTTP_202_ACCEPTED)
+async def resend_verification_email(
+    body: ResendVerificationRequest,
+    background_tasks: BackgroundTasks,
+    session: AsyncSession = Depends(get_async_session),
+) -> dict[str, str]:
+    email_lower = body.email.lower()
```
abdulrafey1: Inconsistent email normalisation: `register_user` (line 92) and `google_callback` (line 204) both use `strip().lower()`, but this endpoint only calls `.lower()`. Pydantic's `EmailStr` does strip some whitespace, but relying on that is an implementation detail. Use the same pattern throughout:
```python
email_lower = body.email.strip().lower()
```

app/api/v1/auth.py
```diff
@@ -85,12 +109,23 @@ async def register_user(user: UserCreate, session: AsyncSession = Depends(get_as
...
         name=user.name,
         username=user.username,
-        email=user.email,
+        email=normalized_email,
         hashed_password=hashed_password,
     )
     session.add(db_user)
+    await session.flush()
+
+    assert db_user.id is not None
```
abdulrafey1: `assert` is compiled away in optimised mode (`python -O`), so this guard disappears in some production setups. The invariant is real (flush above should always populate `id`), but the enforcement should survive optimisation:
```python
if db_user.id is None:
    raise RuntimeError("user.id unexpectedly None after flush")
```

app/api/v1/auth.py
```diff
@@ -221,6 +267,78 @@ async def google_callback(
...
+
+
+@router.post("/verify-email/resend", status_code=status.HTTP_202_ACCEPTED)
+async def resend_verification_email(
+    body: ResendVerificationRequest,
+    background_tasks: BackgroundTasks,
+    session: AsyncSession = Depends(get_async_session),
+) -> dict[str, str]:
+    email_lower = body.email.lower()
+    key = f"email_verify_resend:{hashlib.sha256(email_lower.encode()).hexdigest()}"
```
abdulrafey1, line 316: Rate limiting is per-email only. An attacker can probe many different email addresses in parallel — each gets its own 60-second bucket with no global request budget. This is lower priority given the endpoint already returns 202 for unknown emails (anti-enumeration), but consider pairing with an IP-based bucket if enumeration is a concern:
```python
# IP bucket (pseudo-code — needs request: Request injected)
ip_key = f"email_verify_resend_ip:{hashlib.sha256(request.client.host.encode()).hexdigest()}"
await redis.set(ip_key, "1", ex=10, nx=True)  # coarser window per IP
```

app/api/v1/auth.py
```diff
@@ -221,6 +267,78 @@ async def google_callback(
         return RedirectResponse(url=f"/login?error={quote(str(e))}", status_code=302)
 
 
+@router.post("/verify-email", response_model=UserSchema)
```
abdulrafey1: This endpoint returns the full `UserSchema` (id, name, is_superuser, email_verified, …) but `VerifyEmailClient.tsx` doesn't use the response at all — it only distinguishes success from failure. Returning a 204 or a minimal `{"ok": true}` avoids over-sharing user data on an unauthenticated endpoint.

frontend/app/verify-email/VerifyEmailClient.tsx
```diff
@@ -0,0 +1,161 @@
...
+              name="email"
+              type="email"
+              required
+              placeholder="you@example.com"
+              value={resendEmail}
+              onChange={(e) => setResendEmail(e.target.value)}
+            />
+            <Button
+              onClick={handleResend}
+              disabled={!resendEmail || resendStatus === "sending" || resendStatus === "sent"}
```
abdulrafey1, line 147: Once `resendStatus` reaches `"sent"`, the button is disabled permanently for the lifetime of the component. If the user didn't receive the email and wants to request it again (after the server's 60-second cooldown), they'd have to refresh the page. Consider auto-resetting to `"idle"` after the cooldown window:
```ts
setResendStatus("sent");
setTimeout(() => setResendStatus("idle"), 90_000); // slightly longer than server cooldown
```

frontend/app/login/LoginForm.tsx
```diff
@@ -88,6 +144,24 @@ export default function LoginPage() {
...
+                Please confirm your email before logging in. We sent a link to{" "}
+                <strong>{unverifiedEmail}</strong>.
+              </p>
+              <Button
+                type="button"
+                variant="outline"
+                size="sm"
+                className="mt-2"
+                onClick={handleResend}
+                disabled={resendStatus === "sending" || resendStatus === "sent"}
```
abdulrafey1, line 165: Same permanent-disable issue as in `VerifyEmailClient.tsx`: once the email is sent the button is locked for the rest of the component's lifetime. The server cooldown is 60 seconds — the user has no in-page way to retry after it expires. Auto-reset to `"idle"` after ~90 s.

frontend/app/verify-email/VerifyEmailClient.tsx
```diff
@@ -0,0 +1,161 @@
...
+    } catch (err) {
+      if (err instanceof ApiRequestError && err.status === 429) {
+        setResendStatus("rate_limited");
+      } else {
+        setResendStatus("error");
+      }
+    }
+  };
+
+  const resendButtonLabel = (() => {
```
abdulrafey1: The IIFE `(() => { switch ... })()` re-runs on every render and is an unusual pattern here. A plain lookup object is cleaner and evaluates in constant time:
```ts
const RESEND_LABELS: Record<typeof resendStatus, string> = {
  idle: "Send new link",
  sending: "Sending…",
  sent: "Email sent — check your inbox",
  rate_limited: "Please wait before resending",
  error: "Try again",
};
const resendButtonLabel = RESEND_LABELS[resendStatus];
```

frontend/app/login/LoginForm.tsx
```diff
@@ -45,6 +71,36 @@ export default function LoginPage() {
...
+    } catch (err) {
+      if (err instanceof ApiRequestError && err.status === 429) {
+        setResendStatus("rate_limited");
+      } else {
+        setResendStatus("error");
+      }
+    }
+  };
+
+  const resendButtonLabel = (() => {
```
abdulrafey1: Same IIFE pattern as in `VerifyEmailClient.tsx` — prefer a `Record<ResendStatus, string>` lookup object for clarity and consistency.

app/core/config.py
```diff
@@ -28,6 +28,21 @@ class Settings(BaseSettings):
...
 
+    # Email / SMTP
+    SMTP_HOST: str = ""
+    SMTP_PORT: int = 587
+    SMTP_USERNAME: str = ""
+    SMTP_PASSWORD: str = ""
+    SMTP_USE_TLS: bool = True
+    SMTP_FROM_EMAIL: str = ""
+    SMTP_FROM_NAME: str = "Sparkth"
+
```
zaira-bibi, line 39: These should probably be documented in both `.env.example` and `CLAUDE.md`, since they’re required environment setup variables.

zaira-bibi (changes requested):

app/migrations/versions/f1e914de8460_add_email_verification.py
hamza-56: Fixed in a93703e. Dropped the redundant `op.create_index` on `token_hash` (and its matching `op.drop_index` in `downgrade`); the `UniqueConstraint` already produces the unique index. Migration hasn't shipped to main yet, so editing the file in place is OK per CLAUDE.md (the "never edit migrations" rule is about migrations that have been applied).

app/services/email_verification.py
hamza-56: Fixed in 9600b0e. Wrapped `send_email` in `send_verification_email` with a `try/except (RuntimeError, aiosmtplib.SMTPException)` that logs via `logger.exception('Failed to send verification email to %s', to)` and absorbs — re-raising made no sense since the user already received the 201. Added two regression tests (`test_swallows_and_logs_runtime_error_when_smtp_unconfigured`, `test_swallows_and_logs_smtp_exception`) that verify the function does not raise and that `logger.exception` is called with the recipient.

app/api/v1/auth.py
hamza-56: Fixed in 30fde88. Resend now uses `body.email.strip().lower()` so all three entry points (register, Google callback, resend) normalize identically.

app/api/v1/auth.py
hamza-56: Fixed in 85bfd0b. Replaced the `assert` with `if db_user.id is None: raise RuntimeError(...)` so the invariant survives `python -O`.

app/api/v1/auth.py
hamza-56: Keeping as-is for now. Two reasons:

1. The threat model an IP bucket actually mitigates is narrow: the endpoint already returns 202 silently for unknown/verified emails (so it's not a useful enumeration oracle), and the per-email bucket already caps inbox-flooding of any single victim to one email per cooldown. The remaining concern is SMTP-quota exhaustion via many distinct emails — and `request.client.host` is trivially bypassed by a botnet or proxy network, which is the population that would actually attempt that.
2. False positives from shared NAT (corporate networks, mobile carriers) are non-trivial when the bucket is small enough to be meaningful.

If we want a real budget here I'd rather put it at a layer that sees true client identity (Cloudflare / WAF / our load balancer), or set a global send budget per minute against the SMTP provider's quota. Happy to file a follow-up if you'd still like the in-app bucket as a defense-in-depth layer.

app/api/v1/auth.py
hamza-56: Fixed in a5f6256. Endpoint now returns 204 (no `response_model`), the FastAPI handler returns `None`, the test asserts `status_code == 204` and re-queries the user instead of reading the body, and `verifyEmail()` in `frontend/lib/api.ts` is now `Promise<void>` (drops the `response.json()` call which would have failed on the empty body). The `VerifyEmailContent` caller already only branched on success vs error, so no UI change needed.

frontend/app/verify-email/VerifyEmailClient.tsx
hamza-56: Fixed in 68715ad. Added a `useEffect` that, when `resendStatus` becomes `sent` or `rate_limited`, sets a 90s timer to reset to `idle` (and clears the timer on cleanup / state change). 90s = server cooldown (60s) + small grace. `rate_limited` is included so an early retry doesn't get locked out of the page either.

frontend/app/login/LoginForm.tsx
hamza-56: Fixed in 68715ad alongside the verify-email page — same `useEffect`-with-cleanup that resets `resendStatus` to `idle` after 90s.

frontend/app/verify-email/VerifyEmailClient.tsx
hamza-56: Fixed in 68715ad. Extracted a module-level `RESEND_LABELS: Record<ResendStatus, string>` and switched to `const resendButtonLabel = RESEND_LABELS[resendStatus]`. Also lifted the literal status-string union to a named `ResendStatus` type so the `useState` declaration is no longer a multi-line type literal.

frontend/app/login/LoginForm.tsx
hamza-56: Fixed in 68715ad — same `Record<ResendStatus, string>` lookup + named `ResendStatus` type as in `VerifyEmailClient.tsx`. (Same map shape, different idle label since this lives in the login flow.)

app/core/config.py
hamza-56: `.env.example` was already updated as part of the SMTP/email-verification commit on this branch (the `# SMTP …`, `# Email verification`, and `# Redis (rate limiting; …)` blocks at the bottom of the file). The CLAUDE.md env-var table hadn't caught up — added the SMTP / email-verification / FRONTEND_BASE_URL / REDIS_URL rows in bae2aff.

CLAUDE.md
```diff
@@ -80,6 +80,12 @@ Copy `.env.example` → `.env`. Required variables:
 | `LLM_ENCRYPTION_KEY` | Fernet key for encrypting stored LLM API keys |
 | `REDIS_URL` | Redis for session caching, used in chat plugin |
 | `GOOGLE_CLIENT_ID/SECRET` | Google OAuth |
+| `SMTP_HOST` / `SMTP_PORT` / `SMTP_USERNAME` / `SMTP_PASSWORD` / `SMTP_USE_TLS` | Outbound SMTP (Amazon SES, Mailgun, MailHog, …) |
+| `SMTP_FROM_EMAIL` / `SMTP_FROM_NAME` | From-header for verification + other transactional emails |
+| `EMAIL_VERIFICATION_TOKEN_TTL_HOURS` | Lifetime of an email-verification token (default 24) |
+| `EMAIL_VERIFICATION_RESEND_COOLDOWN_SECONDS` | Per-email cooldown on the resend endpoint (default 60) |
+| `FRONTEND_BASE_URL` | Base URL used in verification email links |
+| `REDIS_URL` | Redis used for the resend rate-limit bucket |
```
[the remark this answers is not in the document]
hamza-56: Fixed in 60e4f61. Merged the two `REDIS_URL` rows into one that names both uses (chat session caching + email-verification resend rate-limit bucket).

CLAUDE.md
```diff
@@ -80,6 +80,12 @@ Copy `.env.example` → `.env`. Required variables:
 | `LLM_ENCRYPTION_KEY` | Fernet key for encrypting stored LLM API keys |
 | `REDIS_URL` | Redis for session caching, used in chat plugin |
 | `GOOGLE_CLIENT_ID/SECRET` | Google OAuth |
+| `SMTP_HOST` / `SMTP_PORT` / `SMTP_USERNAME` / `SMTP_PASSWORD` / `SMTP_USE_TLS` | Outbound SMTP (Amazon SES, Mailgun, MailHog, …) |
+| `SMTP_FROM_EMAIL` / `SMTP_FROM_NAME` | From-header for verification + other transactional emails |
+| `EMAIL_VERIFICATION_TOKEN_TTL_HOURS` | Lifetime of an email-verification token (default 24) |
+| `EMAIL_VERIFICATION_RESEND_COOLDOWN_SECONDS` | Per-email cooldown on the resend endpoint (default 60) |
```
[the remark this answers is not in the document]
hamza-56: Fixed in 60e4f61 — merged the duplicate `REDIS_URL` rows into one combined entry.

app/models/email_verification.py
```diff
@@ -0,0 +1,29 @@
...
+    id: int | None = Field(default=None, primary_key=True)
+    user_id: int = Field(
+        sa_column=Column(
+            Integer,
+            ForeignKey("user.id", ondelete="CASCADE"),
+            nullable=False,
+            index=True,
+        )
+    )
+    token_hash: str = Field(max_length=64, unique=True, index=True)
```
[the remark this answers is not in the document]
hamza-56: Fixed in 60e4f61. Dropped `index=True` from `token_hash` on `EmailVerificationToken`; `unique=True` already produces the index, so the test-DB schema (`SQLModel.metadata.create_all`) now matches the migration.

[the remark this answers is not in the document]
hamza-56: Fixed in 60e4f61 — dropped `index=True` so the model matches the migration's deduped state.

zaira-bibi (approved):
LGTM!

abdulrafey1 (approved):

PR #304, opened 2026-05-05
feat(slack): add LLM synthesis to Slack TA Bot RAG pipeline

### LINKED ISSUE #305: Slack TA Bot: LLM-synthesized answers for Slack TA Bot
The Slack TA Bot currently returns raw document chunks concatenated together when a student asks a question. The response is hard to read, lacks coherence across multiple chunks, and gives no indication of which source each piece of information came from.

Allow a teacher to connect their Slack bot to one of their configured LLM providers so that retrieved chunks are synthesized into a single, natural-language answer before being posted in Slack. When no LLM is configured, or when synthesis fails, the bot falls back to the existing formatted-chunks response so there is no regression in availability.

**Acceptance Criteria**

- [x] A bot with `llm_config_id` set returns a synthesized answer drawn from the top-ranked RAG chunks
- [x] A bot with `llm_config_id = null` continues to return the formatted raw-chunks response (backward compatible)
- [x] LLM synthesis failures (rate limits, network errors, validation errors) are caught and the bot falls back to raw chunks without returning an error to the student
- [x] `llm_temperature` is configurable per bot, defaulting to `0.3` for factual Q&A
- [x] The bot replies with a clear, actionable message when the plugin is missing, disabled, or misconfigured — rather than silently failing
- [x] All existing Slack bot tests continue to pass

zaira-bibi:
## What
Upgrades the Slack TA Bot from raw chunk concatenation to a section-ranked RAG pipeline with optional LLM answer synthesis, backed by the shared `RAGContextService`. Closes #305 

## Changes
- `feat(plugins)`: add `llm_config_id` and `llm_temperature` to `SlackConfig` for per-user LLM synthesis
- `feat(plugins)`: replace ad-hoc store singletons in `rag.py` with `RAGContextService` (section ranking + per-source search + re-rank)
- `feat(plugins)`: add `synthesis.py` with `synthesize_answer` and graceful fallback to raw chunks on LLM errors
- `feat(plugins)`: `add _build_llm_provider` and `_post_slack_message` helpers; harden `_dispatch_event` with plugin presence/enabled/config guards
- `refactor(plugins)`: rename `SlackBotConfig` → `SlackConfig`, `SlackBotPlugin` → `Slack`
- `chore(migrations)`: add migration to delete stale `slack_bot_plugin` registry record and merge heads
- `test(plugins)`: expand Slack test suite with synthesis, config, and route coverage

## How to Test
1. Configure a Slack workspace and set `llm_config_id` in the user's Slack plugin config to a valid LLM config row.
2. Send a question to the bot in Slack — verify the reply is a synthesized natural-language answer, not raw chunk text.
3. Set `llm_config_id` to an invalid ID — verify the bot falls back to the prefixed raw-chunks message without erroring.
4. Disable the Slack plugin for a user and send a message — verify the bot replies with the "currently disabled" message.
5. Remove the plugin config entirely — verify the bot replies with the "configuration incomplete" message.
6. Set `allowed_sources` to two document names — verify `rank_sections` and `search_with_embedding` are called once per source and results are merged and trimmed to limit.
6. Run `pytest tests/slack/` — all tests should pass.

## Notes
Migration required: `7f8a52663a6b` deletes the old `slack_bot_plugin` registry record (plugin was registered under the old name); run alembic upgrade head before deploying.

### DIFF
--- a/app/core_plugins/slack/rag.py
+++ b/app/core_plugins/slack/rag.py
@@ -1,54 +1,136 @@
 """RAG dispatch for the Slack TA Bot plugin."""
 
+import httpx
+from langchain_core.exceptions import LangChainException
+from pydantic import ValidationError
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.core.logger import get_logger
-from app.core_plugins.slack.config import SlackBotConfig
+from app.core_plugins.slack.config import SlackConfig
+from app.core_plugins.slack.synthesis import synthesize_answer
+from app.llm.providers import BaseChatProvider
+from app.rag.context_service import RAGContextService, format_chunks_as_context
 from app.rag.embeddings import BaseEmbeddingProvider
 from app.rag.provider import get_provider
-from app.rag.store import VectorStoreService
+from app.rag.store import SimilarityResult, VectorStoreService
 
 logger = get_logger(__name__)
 
 # Module-level singleton
 _store: VectorStoreService = VectorStoreService()
 
+logger = get_logger(__name__)
+
+# Module-level singleton — embedding model and vector store loaded once per process
+_rag_service: RAGContextService = RAGContextService()
+
+
+_SIMILARITY_THRESHOLD = 0.5
+
 
 async def answer_question(
     session: AsyncSession,
     user_id: int,
     question: str,
-    config: SlackBotConfig,
-    similarity_threshold: float = 0.3,
+    config: SlackConfig,
+    similarity_threshold: float = _SIMILARITY_THRESHOLD,
     limit: int = 5,
     provider: BaseEmbeddingProvider | None = None,
+    llm_provider: BaseChatProvider | None = None,
 ) -> tuple[str, bool]:
-    """Embed question, search vector store, return (answer, rag_matched).
+    """Embed question, rank sections, search vector store, return (answer, rag_matched).
+
+    Mirrors the chat RAG flow when allowed_sources are configured:
+      1. Embed the query.
+      2. Rank document sections by title similarity per source.
+      3. Run similarity search filtered to the top-ranked sections.
+      4. Merge and re-rank results across all sources.
+      5. (Optional) Synthesize a natural-language answer via LLM.
+
+    Falls back to a broad similarity search across all user content when
+    no allowed_sources are configured.
 
     Returns (config.fallback_message, False) when no chunks meet the threshold.
-    provider defaults to HuggingFaceEmbeddingProvider; pass a mock for testing.
+    provider defaults to the RAGContextService embedding provider; pass a mock for testing.
+    llm_provider, when set, sends the question + chunks to an LLM for synthesis.
     """
     embedding_provider = provider or get_provider()
     query_embedding = await embedding_provider.embed_query(question)
 
-    results = await _store.similarity_search(
-        session=session,
-        user_id=user_id,
-        query_embedding=query_embedding,
-        limit=limit,
-        similarity_threshold=similarity_threshold,
-        source_names=config.allowed_sources or None,
-    )
+    all_results: list[SimilarityResult] = []
+    source_names = config.allowed_sources
+
+    if source_names:
+        for source_name in source_names:
+            ranked_sections = await _rag_service.rank_sections(
+                session=session,
+                user_id=user_id,
+                source_name=source_name,
+                query_embedding=query_embedding,
+            )
+            section_filter = list({s["section"] for s in ranked_sections if s["section"]}) or None
+
+            results = await _rag_service.search_with_embedding(
+                session=session,
+                user_id=user_id,
+                source_name=source_name,
+                query_embedding=query_embedding,
+                limit=limit,
+                similarity_threshold=similarity_threshold,
+                sections=section_filter,
+            )
+            all_results.extend(results)
+
+        # Re-rank merged results and keep the top limit
+        all_results.sort(key=lambda r: r.similarity, reverse=True)
+        all_results = all_results[:limit]
+    else:
+        # No source filter: broad search across all user content
+        all_results = await _rag_service._store.similarity_search(
+            session=session,
+            user_id=user_id,
+            query_embedding=query_embedding,
+            limit=limit,
+            similarity_threshold=similarity_threshold,
+        )
 
     logger.info(
-        "RAG search for user_id=%d found %d chunks (threshold=%.2f)",
+        "RAG search for user_id=%d found %d chunks (threshold=%.2f, sources=%s)",
         user_id,
-        len(results),
+        len(all_results),
         similarity_threshold,
+        source_names or "all",
     )
 
-    if not results:
+    if not all_results:
         return config.fallback_message, False
 
-    answer = "\n\n".join(r.chunk.content for r in results)
-    return answer, True
+    # Group results by source and format
+    sources_seen: list[str] = []
+    results_by_source: dict[str, list[SimilarityResult]] = {}
+    for r in all_results:
+        sname = r.chunk.source_name
+        if sname not in results_by_source:
+            sources_seen.append(sname)
+            results_by_source[sname] = []
+        results_by_source[sname].append(r)
+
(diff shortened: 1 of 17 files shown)


abdulrafey1 (approved):
LGTM

app/migrations/versions/ed5bc53f5b16_create_llm_configs_table.py
```diff
@@ -68,7 +68,7 @@ def upgrade() -> None:
     op.execute("""
         SELECT setval(
             pg_get_serial_sequence('llm_configs', 'id'),
-            COALESCE((SELECT MAX(id) FROM llm_configs), 0)
+            COALESCE((SELECT MAX(id) FROM llm_configs), 1)
```
hamza-56: This migration is already on main from commit 47d8728, so editing it here won't run on environments that already applied it. CLAUDE.md says no edits to existing migrations. Probably worth doing this in a fresh autogenerated migration instead.

app/core_plugins/slack/routes.py
```diff
@@ -159,59 +164,169 @@ def get_connection_status(
...
+
+    Returns None if the config cannot be resolved or the provider cannot be built.
+    """
+    try:
+        settings = get_settings()
+        llm_service = LLMConfigService(
+            encryption=get_encryption_service(settings.LLM_ENCRYPTION_KEY),
+            cache=get_cache_service(settings.REDIS_URL, settings.REDIS_KEY_TTL),
+        )
+    except Exception as exc:  # noqa: BLE001
```
hamza-56: Catching plain Exception with a noqa goes against the project rule about specific exception types. The realistic failures here are pretty narrow (ValueError on a bad Fernet key, RedisError, ValidationError), so we could just name them.

app/api/v1/llm.py
```diff
@@ -0,0 +1,193 @@
...
+            config_id=config_id,
+            name=body.name,
+            model=body.model,
+        )
+        await session.commit()
+        return _to_response(config)
+    except LLMConfigValidationError as exc:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(exc)) from exc
+    except ValueError as exc:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(exc)) from exc
```
hamza-56, line 101: service.update raises a plain ValueError for both 'not found' and 'duplicate name' (via IntegrityError), so a name collision on PATCH ends up here as a 404 with an 'already exists' body. The create handler returns 409 for the same case, so behavior is inconsistent across endpoints.

app/llm/service.py
```diff
@@ -144,28 +152,49 @@ async def delete(self, session: AsyncSession, user_id: int, config_id: int) -> b
         logger.info("Soft-deleted LLMConfig id=%s for user_id=%s", config_id, user_id)
         return True
 
-    async def resolve(self, session: AsyncSession, user_id: int, config_id: int) -> str:
-        """Return decrypted API key. Caches by (user_id, config_id). Updates last_used_at."""
+    async def set_active(self, session: AsyncSession, user_id: int, config_id: int, is_active: bool) -> LLMConfig:
```
hamza-56, line 162: rotate_key and delete both invalidate the cached encrypted key, but set_active doesn't. Deactivating a config leaves the key cached for REDIS_KEY_TTL. Worth keeping these in sync.

app/llm/service.py
```diff
@@ -144,28 +152,49 @@ async def delete(self, session: AsyncSession, user_id: int, config_id: int) -> b
...
+        return config
+
+    async def resolve(self, session: AsyncSession, user_id: int, config_id: int) -> tuple[LLMConfig, str]:
+        """Return (config, decrypted_api_key). Caches by (user_id, config_id). Updates last_used_at."""
+        config = await self.get(session, user_id, config_id)
+        if config is None:
+            raise LLMConfigNotFoundError(f"LLMConfig {config_id} not found for user {user_id}")
+        if not config.model:
+            raise LLMConfigModelNotSetError(
+                f"LLMConfig {config_id} has no model set. Update it via PATCH /api/v1/llm/{config_id} before use."
```
hamza-56: The route is mounted as /api/v1/llm/configs/{id}, so this error string sends users to a 404. Looks like the /configs/ segment got dropped.

app/core_plugins/slack/rag.py
```diff
@@ -1,54 +1,136 @@
...
 from app.rag.provider import get_provider
-from app.rag.store import VectorStoreService
+from app.rag.store import SimilarityResult, VectorStoreService
 
 logger = get_logger(__name__)
 
 # Module-level singleton
 _store: VectorStoreService = VectorStoreService()
 
+logger = get_logger(__name__)
```
hamza-56, line 17: logger gets reassigned here even though it was already set on line 17, and the _store on line 20 is never used after this (the code below goes through _rag_service._store instead). Looks like leftovers from a refactor.

app/core_plugins/slack/rag.py
```diff
@@ -1,54 +1,136 @@
...
+
+# Module-level singleton — embedding model and vector store loaded once per process
+_rag_service: RAGContextService = RAGContextService()
+
+
+_SIMILARITY_THRESHOLD = 0.5
```
hamza-56, line 23: Threshold went from 0.3 on main to 0.5 here. Slack doesn't have the retry fallback that chat has, so this will surface fallback_message quite a bit more often. Was the bump intentional?

app/core_plugins/slack/rag.py
```diff
@@ -1,54 +1,136 @@
...
+                sections=section_filter,
+            )
+            all_results.extend(results)
+
+        # Re-rank merged results and keep the top limit
+        all_results.sort(key=lambda r: r.similarity, reverse=True)
+        all_results = all_results[:limit]
+    else:
+        # No source filter: broad search across all user content
+        all_results = await _rag_service._store.similarity_search(
```
hamza-56: Reaching into _rag_service._store breaks encapsulation a bit. If RAGContextService renames this internal field, Slack search will break silently. A public method on RAGContextService would probably be more durable.

app/core_plugins/slack/rag.py
```diff
@@ -1,54 +1,136 @@
...
+            return answer, True
+        except (LangChainException, ValidationError, ValueError, RuntimeError, httpx.RemoteProtocolError) as exc:
+            logger.warning(
+                "LLM synthesis failed for user_id=%d, falling back to raw chunks: %s: %s",
+                user_id,
+                type(exc).__name__,
+                exc,
+            )
+
+    return f"AI summary is not available at the moment, but here is what RAG found:\n\n{formatted_context}", True
```
hamza-56: This makes config.fallback_message unreachable whenever llm_config_id is set, since synthesis failures (or a None llm_provider getting past the if) always show this English preamble instead. Probably want to fall back to the configured message or just the raw chunks here.

app/core_plugins/slack/rag.py
zaira-bibi: It was 0.7 on main at the time of this commit. I brought it down to 0.5.

app/core_plugins/slack/rag.py
```diff
@@ -1,54 +1,132 @@
...
+        except (LangChainException, ValidationError, ValueError, RuntimeError, httpx.RemoteProtocolError) as exc:
+            logger.warning(
+                "LLM synthesis failed for user_id=%d, falling back to raw chunks: %s: %s",
+                user_id,
+                type(exc).__name__,
+                exc,
+            )
+            return f"Could not generate an AI summary, but here is what RAG found:\n\n{formatted_context}", True
+
+    return f"AI summary is not available at the moment, but here is what RAG found:\n\n{formatted_context}", True
```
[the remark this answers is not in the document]
zaira-bibi: @claude `config.fallback_message` semantically means "we couldn't help you at all," while line 130 means "we found relevant content, but could not generate an AI summary due to an error on the AI provider's end".

The intent behind line 132 is, when students are expecting AI (the platform implies it) but the teacher hasn't configured `llm_config_id`, so they should see "AI summary is not available" in that case.

hamza-56 (approved):

PR #307, opened 2026-05-06
fix(chat): pass conversation history to scope classifier

### LINKED ISSUE #306: Chat: Scope classifier incorrectly refuses in-context follow-up messages
**Problem**

The scope classifier evaluates each user message in isolation, without considering conversation history. This causes false refusals when a user answers a clarifying question the assistant asked.

**Example**:
```
User:  create a refresher course on this doc
Bot:   Who is the primary audience for this course?
User:  AI developers and engineers at Arbisoft   ← flagged as out-of-scope ❌
Bot:   I'm a course creation assistant and can only help with designing and building courses...
```
The user then has to explicitly say "my previous message is related to course generation" to continue.

**Root Cause**

In `app/core_plugins/chat/routes.py`, the tiered scope check extracts only the latest user message (`_extract_query_text`) and passes it to the classifier with no surrounding context. The LLM classifier in `app/llm/classifier.py` also received only a single `HumanMessage`, so a short reply like "AI developers and engineers at Arbisoft" had no signal that it was a direct answer to an assistant question.

**Acceptance Criteria**

 - [ ] Answering the assistant's clarifying questions does not trigger a refusal
  - [ ] Genuinely out-of-scope first messages (e.g. "What is the capital of France?") are still refused
  - [ ] Existing classifier tests pass; new tests cover the history-aware path

zaira-bibi:
## What
Pass conversation history to the scope classifier so that short follow-up answers to the assistant's own clarifying questions are no longer incorrectly refused. Closes #306 

## Changes
- `fix(api)`: update `ScopeClassifier.classify()` to accept optional history and include last 6 turns as context
- `fix(api)`: update classifier system prompt to evaluate scope using full conversation, not latest message in isolation
- `fix(api)`: pass `db_messages` as history to `classifier.classify()` in chat completions route
- `fix(api)`: update `system_prompt.txt` with context-awareness instruction as secondary safety net

## How to Test
1. Start a new conversation and send: `create a refresher course on this doc`
2. The assistant asks a clarifying question (e.g. "Who is the primary audience?")
3. Reply with a short in-context answer: `AI developers and engineers`.
4. Verify the assistant continues the course creation workflow instead of returning the refusal message
5. In a fresh conversation, send a clearly out-of-scope message: `What is the capital of France?`
6. Verify the refusal message is still returned

## Notes
None

### DIFF
--- a/app/core_plugins/chat/routes.py
+++ b/app/core_plugins/chat/routes.py
@@ -2,7 +2,7 @@ import asyncio
 import json
 import re
 from functools import lru_cache
-from typing import Any, AsyncGenerator
+from typing import Any, AsyncGenerator, Literal, cast
 from uuid import UUID
 
 import anthropic
@@ -49,7 +49,7 @@ from app.core_plugins.chat.schemas import (
 )
 from app.core_plugins.chat.service import ChatService
 from app.core_plugins.chat.tools import get_tool_registry
-from app.llm.classifier import ScopeClassifier
+from app.llm.classifier import HistoryTurn, ScopeClassifier
 from app.llm.prompt import REFUSAL_MESSAGE, is_query_in_scope
 from app.llm.providers import (
     DEFAULT_MODEL,
@@ -525,7 +525,11 @@ async def chat_completion(
                 else:
                     # Tier 2: LLM classifier for nuanced cases keywords can't handle
                     classifier = ScopeClassifier(provider_name=request.provider, api_key=api_key)
-                    prior_history = [{"role": m.role, "content": m.content} for m in db_messages]
+                    prior_history: list[HistoryTurn] = [
+                        {"role": cast(Literal["user", "assistant"], m.role), "content": m.content}
+                        for m in db_messages
+                        if m is not db_messages[-1] or not (m.role == "user" and m.content == query_text)
+                    ]
                     _in_scope = await classifier.classify(query_text, history=prior_history)
             if not _in_scope:
                 await service.add_message(
--- a/app/core_plugins/chat/routes.py
+++ b/app/core_plugins/chat/routes.py
@@ -525,7 +525,8 @@ async def chat_completion(
                 else:
                     # Tier 2: LLM classifier for nuanced cases keywords can't handle
                     classifier = ScopeClassifier(provider_name=request.provider, api_key=api_key)
-                    _in_scope = await classifier.classify(query_text)
+                    prior_history = [{"role": m.role, "content": m.content} for m in db_messages]
+                    _in_scope = await classifier.classify(query_text, history=prior_history)
             if not _in_scope:
                 await service.add_message(
                     session=session,
--- a/app/llm/classifier.py
+++ b/app/llm/classifier.py
@@ -1,10 +1,10 @@
 """LLM-based scope classifier for the learning design assistant."""
 
-from typing import Any
+from typing import Any, Literal, TypedDict
 
 from langchain_anthropic import ChatAnthropic
 from langchain_core.exceptions import LangChainException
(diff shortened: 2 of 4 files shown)


abdulrafey1 (approved):
LGTM

app/core_plugins/chat/routes.py
```diff
@@ -525,7 +525,8 @@ async def chat_completion(
                 else:
                     # Tier 2: LLM classifier for nuanced cases keywords can't handle
                     classifier = ScopeClassifier(provider_name=request.provider, api_key=api_key)
-                    _in_scope = await classifier.classify(query_text)
+                    prior_history = [{"role": m.role, "content": m.content} for m in db_messages]
```
hamza-56: heads up, by this point `db_messages` already contains the message we just persisted (see the `add_message` loop above and the fetch on line 511), so `prior_history` ends with this same `query_text`. The classifier then appends `query_text` on its own, so the current message ends up in the input twice.

app/llm/classifier.py
```diff
@@ -73,22 +76,33 @@ def __init__(self, provider_name: str, api_key: str) -> None:
 
         self._chain: Any = llm.with_structured_output(_ScopeResult)
 
-    async def classify(self, query: str) -> bool:
+    async def classify(self, query: str, history: list[dict[str, str]] | None = None) -> bool:
```
hamza-56: noticed there are no tests for the new `history` arg. role mapping, the slice cap, and the case where history already ends with the current query all feel worth covering.

also `list[dict[str, str]]` is pretty loose for what's really a structured role/content shape.

app/llm/classifier.py
```diff
@@ -73,22 +76,33 @@ def __init__(self, provider_name: str, api_key: str) -> None:
...
+        so the classifier can determine if the current message is a reply to an assistant question.
+
         Fails open on any error — the main LLM's system prompt handles
         out-of-scope requests as a fallback.
         """
         if not query.strip():
             return True
 
+        # Build messages: system prompt + up to last 6 history turns + current query
+        messages: list[Any] = [SystemMessage(content=_CLASSIFIER_SYSTEM_PROMPT)]
```
hamza-56: `list[Any]` turns off type checking here, when these are all `BaseMessage` subclasses.

app/llm/classifier.py
```diff
@@ -73,22 +76,33 @@ def __init__(self, provider_name: str, api_key: str) -> None:
...
+        for turn in (history or [])[-6:]:
```
hamza-56: why 6 specifically? not jumping out from the surrounding context.

app/llm/classifier.py
```diff
@@ -73,22 +76,33 @@ def __init__(self, provider_name: str, api_key: str) -> None:
...
+            role = turn.get("role", "")
+            content = turn.get("content", "")
+            if not isinstance(content, str):
+                continue
+            if role == "assistant":
+                messages.append(AIMessage(content=content))
+            elif role == "user":
+                messages.append(HumanMessage(content=content))
+        messages.append(HumanMessage(content=query))
```
hamza-56, line 109: heads up, if a caller passes history that ends with the current user message (which `routes.py` is doing today), this line ends up duplicating it in the input.

app/llm/classifier.py
zaira-bibi: It doesn't come from the surrounding context. It's an arbitrary number picked to give enough context for the classifier to recognise follow-up answers, like "python" being a reply to "what language?", without blowing up the token budget.

app/llm/classifier.py
zaira-bibi: the `routes.py` fix handles the only caller so the classifier has no deduplication.

hamza-56 (approved):

Issue #291, opened 2026-04-29
[Sparkth] MD extraction in batches


MahnoorArbisoft:
Current implementation processes entire document that uses a lot of RAM. To optimize this, we'll start extracting in batches (more technical details to be added by @abdulrafey1 )

abdulrafey1:
## Investigation findings: batched MD extraction

I instrumented `_extract_pdf` to process pages in batches of 10 (calling `pymupdf4llm.to_markdown` with `pages=[...]` per batch, and `fitz.TOOLS.store_shrink(100)` between batches) and profiled RSS across 6 ingestions of the same 3 PDFs (~278, 471, 261 pages).

### What batching fixed
- **Within a single file**, peak RSS is bounded. After the first 1–2 batches load fonts, subsequent batches add ~0–5 MB each and the curve plateaus. No unbounded growth per file.

### What batching did NOT fix
RSS still climbs **across files** and never recovers:

| Run | File | RSS Δ (extraction) |
|---|---|---|
| 1 | Vol 1 (cold) | **+260 MB** |
| 2 | Vol 2 | +155 MB |
| 3 | Vol 3 | +66 MB |
| 4 | Vol 1 (re-run) | +41 MB |
| 5 | Vol 2 (re-run) | +124 MB |
| 6 | Vol 3 (re-run) | +40 MB |

Process RSS went from **1162 MB → 1852 MB** over the run with no decreases between files.

### Root cause
`fitz.TOOLS.store_shrink` only releases unreferenced entries from MuPDF's process-global font/CMap store. Font subsets are deduplicated and pinned across documents, so they accumulate for the lifetime of the process. Glibc's allocator also retains freed arenas rather than returning them to the OS, compounding the effect.

### Conclusion
Batching solves the per-file memory ceiling, but it does **not** solve the cross-file accumulation that causes pod OOMs.

PR #316, opened 2026-05-11
feat(rag): batch PDF extraction and reject scanned PDFs

abdulrafey1:
## What
Bound peak memory during PDF extraction by processing pages in batches, and reject image-only/scanned PDFs at ingestion with a clear error so they no longer pollute the vector store or get stuck in PROCESSING. Addresses #291.

## Changes
- feat(rag): extract PDFs in page batches to bound peak memory
- feat(rag): reject scanned PDFs with a clear user-facing error

### Batched extraction
1. Ingest a large multi-page PDF (e.g. 200+ pages) through the Google Drive plugin.
2. Watch the API logs and confirm the file is processed without OOM and ends in `READY`.
3. Optionally enable `MEMORY_PROFILING_ENABLED=true` and verify the per-batch RSS pattern in `logs/ram-logs.txt` shows a bounded ceiling rather than unbounded growth within a single file.

### Scanned-PDF rejection
1. Upload a scanned/image-only PDF (every page is a photo of text, no text layer) via Google Drive.
2. Confirm the file transitions to `FAILED` (not stuck in `PROCESSING`).
3. Confirm the error message surfaced to the user reads: *"<filename> appears to be a scanned PDF (image-only pages with no extractable text). OCR-based ingestion is not yet supported — please upload a text-based PDF."*
4. Upload a normal text PDF and confirm it still extracts and indexes normally.

### Tuning
- Override the detection threshold via env: `RAG_SCANNED_PDF_MIN_CHARS_PER_PAGE=50` (default 100).

## Notes
- New env var: `RAG_SCANNED_PDF_MIN_CHARS_PER_PAGE` (int, default 100). Optional — no migration required.
- Investigation findings on cross-file RSS accumulation (which batching alone does **not** fix) posted as a comment on #291; a follow-up issue for process-level isolation is recommended.

### DIFF
--- a/app/rag/extraction.py
+++ b/app/rag/extraction.py
@@ -138,11 +138,11 @@ def _extract_pdf(data: bytes, source_name: str, *, batch_size: Optional[int] = N
     settings = get_settings()
     effective_batch_size = batch_size if batch_size is not None else settings.RAG_PDF_EXTRACTION_BATCH_SIZE
     with fitz.open(stream=data, filetype="pdf") as doc:
-        if _is_scanned_pdf(doc, min_chars_per_page=settings.RAG_SCANNED_PDF_MIN_CHARS_PER_PAGE):
-            raise ScannedPDFError(source_name)
-        page_count = len(doc)
-        parts: list[str] = []
         try:
+            if _is_scanned_pdf(doc, min_chars_per_page=settings.RAG_SCANNED_PDF_MIN_CHARS_PER_PAGE):
+                raise ScannedPDFError(source_name)
+            page_count = len(doc)
+            parts: list[str] = []
             for start in range(0, page_count, effective_batch_size):
                 pages = list(range(start, min(start + effective_batch_size, page_count)))
                 batch_md = pymupdf4llm.to_markdown(doc=doc, pages=pages, page_chunks=False)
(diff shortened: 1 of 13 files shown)


app/rag/extraction.py
```diff
@@ -100,22 +102,63 @@ def __repr__(self) -> str:  # pragma: no cover
...
+    Sampling ~10% of pages at random spreads coverage so a few atypical
+    pages (covers, blank inserts, watermarks) cannot dominate the decision.
+    """
+    page_count = len(doc)
+    if page_count == 0:
+        return False
+
+    sample_size = max(1, round(page_count * sample_ratio))
+    sample_size = min(sample_size, page_count)
+    sampled = random.sample(range(page_count), sample_size)
```
hamza-56: Worth flagging: there's no seed in play here, so the same PDF can come out accepted on one upload and rejected as "scanned" on the next. For most documents this never matters — born-digital pages have thousands of chars, scanned pages have basically none — but mixed-content files sitting near the threshold will flip-flop, and that's pretty ugly to debug when a user reports "it worked yesterday." Small PDFs are the roughest case: a 5-page doc samples exactly one page, so the docstring's promise that "atypical pages cannot dominate" doesn't really hold there.

zaira-bibi (approved):
LGTM!

hamza-56 (approved):

PR #322, opened 2026-05-13
feat(chat): show and persist tool calls in assistant messages

zaira-bibi:
## What
Show tool call names during LLM execution, collapse to a persistent "X operations completed" summary after, and save them in message metadata so the summary survives page refresh.

## Changes
- `feat(chat)`: collect `tool_end` events during streaming and save tool call names in message metadata alongside `rag_sections`
- `feat(api)`: add `tool_calls` field to `MessageResponse` and expose via conversation history endpoint
- `feat(frontend)`: display live tool names during streaming, then collapse to a dotted-underline summary with a hover tooltip listing each call
- `feat(frontend)`: restore `toolCalls` from API response when loading conversation history
- `test(chat)`: add tests for `_parse_tool_calls`, metadata persistence, done payload, and `MessageResponse.tool_calls`

## How to Test
1. Open a conversation that uses tools (e.g. ask a question that triggers RAG or an OpenedX tool)
2. During generation, verify tool names appear above the response bubble with a pulsing indicator
3. After generation completes, verify the list collapses to "X operations completed" with a dotted underline
4. Hover over "X operations completed" — confirm a tooltip lists each tool call in order
5. Refresh the page — confirm "X operations completed" and the tooltip are still present

## Notes
No migration required — `tool_calls` is stored in the existing `model_metadata` JSON column alongside `rag_sections`.

### DIFF
--- a/app/core_plugins/chat/routes.py
+++ b/app/core_plugins/chat/routes.py
@@ -112,25 +112,15 @@ async def _stream_out_of_scope_refusal() -> AsyncGenerator[str, None]:
     yield f"data: {json.dumps({'done': True, 'content': REFUSAL_MESSAGE})}\n\n"
 
 
-def _parse_rag_sections(model_metadata: str | None) -> list[dict[str, Any]] | None:
+def _parse_metadata_list(model_metadata: str | None, key: str) -> list[dict[str, Any]] | None:
     if not model_metadata:
         return None
     try:
         meta = json.loads(model_metadata)
-        sections = meta.get("rag_sections")
-        return sections if isinstance(sections, list) else None
-    except (json.JSONDecodeError, AttributeError):
-        return None
-
-
-def _parse_tool_calls(model_metadata: str | None) -> list[dict[str, Any]] | None:
-    if not model_metadata:
-        return None
-    try:
-        meta = json.loads(model_metadata)
-        calls = meta.get("tool_calls")
-        return calls if isinstance(calls, list) else None
-    except (json.JSONDecodeError, AttributeError):
+        value = meta.get(key)
+        return value if isinstance(value, list) else None
+    except (json.JSONDecodeError, AttributeError) as exc:
+        logger.error("Failed to parse model_metadata for key %r: %s", key, exc)
         return None
 
 
@@ -804,7 +794,7 @@ async def stream_chat_response(
     # --- Phase 2: LLM streaming ---
     full_response = ""
     conversation_id: int = conversation.id  # type: ignore[assignment]
-    completed_tool_calls: list[dict[str, str]] = []
+    completed_tool_calls: list[dict[str, Any]] = []
 
     try:
         async for event in provider.stream_message(messages, tools=tools):
@@ -1037,8 +1027,8 @@ async def get_conversation(
                 message_type=msg.message_type,
                 attachment_name=msg.attachment_name,
                 attachment_size=msg.attachment_size,
-                rag_sections=_parse_rag_sections(msg.model_metadata),
-                tool_calls=_parse_tool_calls(msg.model_metadata),
+                rag_sections=_parse_metadata_list(msg.model_metadata, "rag_sections"),
+                tool_calls=_parse_metadata_list(msg.model_metadata, "tool_calls"),
                 is_error=msg.is_error,
             )
             for msg in messages
--- a/app/core_plugins/chat/routes.py
+++ b/app/core_plugins/chat/routes.py
@@ -123,6 +123,17 @@ def _parse_rag_sections(model_metadata: str | None) -> list[dict[str, Any]] | No
         return None
 
 
+def _parse_tool_calls(model_metadata: str | None) -> list[dict[str, Any]] | None:
+    if not model_metadata:
+        return None
+    try:
+        meta = json.loads(model_metadata)
+        calls = meta.get("tool_calls")
+        return calls if isinstance(calls, list) else None
+    except (json.JSONDecodeError, AttributeError):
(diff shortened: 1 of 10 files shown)


app/llm/providers.py
```diff
@@ -333,55 +337,26 @@ async def _stream_message_simple(self, messages: list[dict[str, Any]]) -> AsyncI
...
 
-    async def _stream_message_with_tools(self, messages: list[dict[str, Any]], tools: list[Any]) -> AsyncIterator[str]:
-        """Stream a message with tool support."""
-        llm = self._create_llm(streaming=True)
+    async def _stream_message_with_tools(
+        self, messages: list[dict[str, Any]], tools: list[Any]
+    ) -> AsyncIterator[dict[str, Any]]:
+        """Tool-loop streaming: silences intermediate LLM chatter, emits tool events,
+        and yields only the final response as token events."""
+        llm = self._create_llm(streaming=False)
```
hamza-56, line 349: Switching from `astream` + `streaming=True` to `ainvoke` + `streaming=False` looks aimed at silencing the LLM's intermediate chatter between tool calls, which is a real improvement. But it also drops streaming for the *final* answer — that reply now arrives as a single event after the model has fully finished generating (see the `yield {"type": "token", "content": content}` near the bottom of the function). So when a tool is involved, the bubble will sit on the thinking dots and then the whole response will pop in at once, instead of streaming in token-by-token like it does today for non-tool chats. Was the intent to suppress only the intermediate text, or also the final-response streaming? Worth calling this out explicitly either way, since it's a noticeable UX shift for any chat that uses tools.

app/core_plugins/chat/routes.py
```diff
@@ -126,6 +126,17 @@ def _parse_rag_sections(model_metadata: str | None) -> list[dict[str, Any]] | No
...
 
+def _parse_tool_calls(model_metadata: str | None) -> list[dict[str, Any]] | None:
+    if not model_metadata:
+        return None
+    try:
+        meta = json.loads(model_metadata)
+        calls = meta.get("tool_calls")
+        return calls if isinstance(calls, list) else None
+    except (json.JSONDecodeError, AttributeError):
+        return None
```
hamza-56: The exception-handling section of the project's CLAUDE.md asks for every caught exception to be logged with enough context to diagnose the failure. Here, if the metadata JSON stored on a message is ever corrupted (bad upgrade, manual DB edit, etc.), this helper quietly returns `None` and a message that *did* have tool calls will look like it had none — with no log trail at all to point at the bad row. The neighboring `_parse_rag_sections` has the same gap, so this is consistent with what's already there, but worth surfacing in case you want to address both together.

app/core_plugins/chat/routes.py
```diff
@@ -725,19 +736,32 @@ async def stream_chat_response(
     # --- Phase 2: LLM streaming ---
     full_response = ""
     conversation_id: int = conversation.id  # type: ignore[assignment]
+    completed_tool_calls: list[dict[str, str]] = []
```
hamza-56: Small inconsistency to flag: this is annotated `list[dict[str, str]]`, but the matching schema field on `MessageResponse` is `list[dict[str, Any]] | None`. They agree today because every entry is just a `{"name": ...}` dict, but if a future event ever carries a non-string value (numeric tool id, structured args, etc.), the two types will silently disagree and mypy won't catch it on this side.

frontend/plugins/chat/hooks/useChatStream.ts
```diff
@@ -141,6 +141,31 @@ function applyStatusEvent(
...
+        // Mark the first running entry for this tool name as done
+        let marked = false;
+        const updated = existing.map((t) => {
+          if (!marked && t.name === toolName && t.status === "running") {
+            marked = true;
+            return { ...t, status: "done" as const };
+          }
+          return t;
+        });
+        return { ...msg, toolCalls: updated };
```
hamza-56: This pairs `tool_start` and `tool_end` events by tool *name* alone. Today the backend yields them in strict sequential order so it works fine, but if the model is ever allowed to call the same tool in parallel (e.g., two `search_web` calls firing at once), there's no way for this code to tell which `tool_end` belongs to which running entry — whichever running card the loop hits first will flip to "done", even if it isn't the one that actually finished. Not a today-bug, but worth being aware of as tool-use grows.

frontend/plugins/chat/components/messages/AssistantMessage.tsx
```diff
@@ -24,14 +24,22 @@ interface AssistantMessageProps {
...
   setPreviewAttachment,
   onOptionClick,
 }: AssistantMessageProps) {
   const displayText = message.streamedContent ?? message.content;
-  const isThinking = message.isTyping && !displayText;
+  const toolCalls = message.toolCalls ?? [];
+  const hasRunningTools = toolCalls.some((t) => t.status === "running");
+  const isThinking = message.isTyping && !displayText && !hasRunningTools;
+  // While tools are running and no response text yet, suppress the card entirely
+  const showCard = !hasRunningTools || !!displayText || message.isError || !message.isTyping;
```
hamza-56, line 42: Between `toolCalls`, `hasRunningTools`, `isThinking`, `showCard`, and the inline tool-call JSX further down, what's actually visible inside the assistant bubble now depends on four or five interacting booleans. It reads okay today if you follow it line-by-line, but every future state added here (error while tools are still running, tools-only response with no text, etc.) will have to be reasoned about against all the others. Worth keeping an eye on before this grows further.

hamza-56 (approved):

PR #324, opened 2026-05-13
feat(chat): add RAG intent router with per-conversation attachment management

abdulrafey1:
## What
Automates the RAG retrieval decision in the chat plugin so teachers are no longer forced to manually re-attach Drive files on each message turn. An intent router (a small, fast LLM step) inspects the user's latest message and lightweight section-title metadata from their conversation attachments, then decides whether RAG retrieval should run for that turn (binary: run / skip). RAG now activates automatically based on message intent rather than requiring per-message inline file selection.

## Changes
- `feat(chat)`: add `RAGIntentRouter` class (`app/core_plugins/chat/intent_router.py`) with `RAGIntentRouterError`; fails closed on `LangChainException`/`ValidationError`
- `feat(chat)`: add `rag_intent_router_system_prompt.txt` prompt asset
- `feat(chat)`: add `ConversationAttachment` SQLModel join table + `attachments` relationship on `Conversation`; legacy columns untouched
- `feat(chat)`: add `RAGRoutingDecision`, `ConversationAttachmentCreate`, `ConversationAttachmentResponse` schemas
- `feat(chat)`: add `attach_drive_file`, `detach_drive_file`, `list_conversation_attachments` service methods
- `feat(chat)`: wire intent router into `chat_completion`; emit `scanning_attachments` / `skipping_rag` SSE events; add `POST`/`DELETE /conversations/{uuid}/attachments` endpoints; 502 fail-closed handler
- `feat(migrations)`: add Alembic migration for `chat_conversation_attachments` join table with back-fill from legacy `active_drive_file_id(s)` columns
- `fix(core)`: add conditional pool kwargs in `app/core/db.py` for SQLite test compatibility
- `test(chat)`: 9 unit tests for intent router, 14 tests for attach/detach service + endpoints, 5 integration tests driving `/api/v1/chat/completions`

## How to Test
1. Apply the migration: `make migrations`
2. Start the stack: `make dev.up`
3. Run the full test suite: `make test` — expect 969 tests passing
4. Run lint and type-check: `make lint && make mypy`
5. In the chat UI, attach a Drive file to a conversation via `POST /api/v1/chat/conversations/{uuid}/attachments`
6. Send a message that clearly relates to the attached document — confirm RAG retrieval fires (look for `scanning_attachments` SSE event followed by retrieved context in the completion)
7. Send a message that is clearly off-topic from the document — confirm RAG is skipped (look for `skipping_rag` SSE event and no retrieval overhead)
8. Remove the attachment via `DELETE /api/v1/chat/conversations/{uuid}/attachments/{id}` and confirm subsequent messages skip RAG
9. Test that an intent router failure returns HTTP 502 (simulate by providing an invalid LLM key)

## Notes
**Migration required:** Run `make migrations` to apply the new `chat_conversation_attachments` join table. The migration includes a back-fill step from the legacy `active_drive_file_id`/`active_drive_file_ids` columns; those columns remain intact so no rollback risk.

**Breaking changes:** None. Legacy columns are preserved.

**No new environment variables required.** The intent router reuses the same LLM configured for the main chat.


app/core_plugins/chat/intent_router.py
```diff
@@ -0,0 +1,104 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from app.core.logger import get_logger
+from app.core_plugins.chat.schemas import RAGRoutingDecision
+from app.models.drive import DriveFile
+
+logger = get_logger(__name__)
+
+# Load system prompt at module import time
+_SYSTEM_PROMPT = (Path(__file__).parent / "assets" / "rag_intent_router_system_prompt.txt").read_text()
```
hamza-56, line 19: Reading the prompt at import time means a missing or unreadable asset will crash app startup rather than just failing at chat-time. Easy to miss in a deploy.

app/core_plugins/chat/intent_router.py
```diff
@@ -0,0 +1,104 @@
...
+                *[get_document_structure(user_id=user_id, file_id=cast(int, f.id)) for f in attached_files],
+                return_exceptions=True,
+            )
+
+            attachment_summary = "The user has attached the following files:\n"
+            for file, sections_or_exc in zip(attached_files, results):
+                attachment_summary += f"\n- {file.name}:\n"
+                if isinstance(sections_or_exc, BaseException):
+                    logger.warning(
+                        "Failed to get document structure for file %d: %s",
```
hamza-56: `file.id` is typed `int | None`, so this `%d` format will raise `TypeError` if it's ever None, awkward when you're already inside an error path.

app/core_plugins/chat/service.py
```diff
@@ -190,4 +202,85 @@ async def update_conversation_title(
...
-            logger.warning(f"Conversation {conversation_id} not found for title update")
+            logger.warning("Conversation %s not found for title update", conversation_id)
+
+    async def attach_drive_file(
+        self,
+        session: AsyncSession,
+        conversation_id: int,
+        drive_file_id: int,
+    ) -> ConversationAttachment:
+        """Attach a drive file to a conversation (upsert-safe)."""
```
hamza-56, line 222: This helper calls `session.commit()` and `session.rollback()` directly, so it owns the outer request-scoped transaction. When `chat_completion` invokes it mid-handler, a rollback here will quietly drop any other unsaved state on the session.

app/core_plugins/chat/service.py
```diff
@@ -190,4 +202,85 @@ async def update_conversation_title(
...
+        if attachment:
+            await session.delete(attachment)
+            await session.commit()
+
+    async def list_conversation_attachments(
+        self,
+        session: AsyncSession,
+        conversation_id: int,
+    ) -> list[DriveFile]:
+        """List ready drive files attached to a conversation."""
```
hamza-56: No `user_id` filter on this query, it relies entirely on the caller to have validated conversation ownership. Fine today, but easy for a future call site to misuse.

app/core_plugins/chat/routes.py
```diff
@@ -353,6 +356,13 @@ async def chat_completion(
                 drive_file_id=active_drive_file_id,
                 drive_file_ids=active_drive_file_ids,
             )
+            # Sync inline drive_file blocks into the join table so the router can find them.
+            for fid in active_drive_file_ids:
```
hamza-56: This same write-through loop already lives inside `ChatService.create_conversation`. Having it in two places means the next change has to remember to touch both spots.

app/core_plugins/chat/routes.py
```diff
@@ -463,18 +495,43 @@ async def chat_completion(
...
+                resolved_messages = await _resolve_drive_file_blocks(
+                    messages=unresolved_messages,
+                    session=session,
+                    user_id=current_user.id,  # type: ignore[arg-type]
+                    rag_service=rag_service,
+                    llm=provider.create_llm(),
+                )
+            else:
+                # Otherwise use request messages as-is (no RAG)
+                resolved_messages = request.messages
```
hamza-56, line 541: When the router decides to skip RAG and the request is non-streaming, `request.messages` flows through as-is. If the user sent inline `drive_file` content blocks this turn, they'll reach the LLM unredacted, the safety strip further down in `stream_chat_response` only runs on the streaming path.

app/core_plugins/chat/routes.py
```diff
@@ -1029,6 +1123,95 @@ async def clear_active_drive_file(
...
+        user_id=current_user.id,  # type: ignore
+    )
+    if not conversation:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Conversation not found")
+
+    df_result = await session.exec(
+        select(DriveFileModel).where(
+            DriveFileModel.id == drive_file_id,
+            DriveFileModel.user_id == current_user.id,
+            DriveFileModel.is_deleted == False,  # noqa: E712
```
hamza-56: Filtering on `is_deleted == False` here means once the underlying drive file gets soft-deleted, the user can no longer detach the attachment, and the orphan row in the join table becomes unreachable from the API.

app/core_plugins/chat/intent_router.py
abdulrafey1: Since the file is committed to the repo alongside the code, a missing file would mean a broken deployment regardless. Reading at import time actually gives you faster, clearer failure (startup crash with a readable traceback) rather than a silent issue that surfaces mid-conversation.

abdulrafey1:
@claude Thanks for the thorough review! Here's where things stand:

**Addressed:**
- `%d` → `%s` for `file.id` in intent router warning log
- `attach_drive_file` now uses `begin_nested()` savepoint so a race-condition rollback only affects the insert, not the caller's outer transaction
- `list_conversation_attachments` now enforces `user_id` ownership at the query level via a `Conversation` join
- Extracted `sync_drive_file_attachments` to deduplicate the write-through loop — both call sites (new and existing conversation) now delegate to the helper
- `_strip_drive_file_blocks` extracted and applied on the non-streaming path to fix `drive_file` blocks leaking to the LLM when RAG is skipped
- Removed `is_deleted` filter from the detach ownership check so soft-deleted files can be detached without leaving orphan join-table rows

**Still open / intentionally deferred:**

- **`session.commit()` in `attach_drive_file` owning the outer transaction** — mitigated by the `begin_nested()` savepoint. A full refactor to flush-only with caller-owned commits would require touching every `ChatService` method; noted for a future cleanup pass.
- **No `user_id` filter in `detach_drive_file`** — agree it's a defense-in-depth gap. Happy to add it in this PR if preferred, or track it as a follow-up.
- **`sync_drive_file_attachments` called from both `create_conversation` and the route** — these serve genuinely different code paths (new vs. existing conversation), so the call sites are intentional. The duplication of the raw loop is gone; what remains is two legitimate callers of the same helper.

zaira-bibi (approved):
Please resolve the remaining open comments. Otherwise, looks good to me!

hamza-56 (approved):

PR #326, opened 2026-05-14
feat(chat): join-table attachments, PersistedFilesInfo, docker HF cache

abdulrafey1:
## What
Completes the drive-file attachment model migration by removing the
denormalised `active_drive_file_id / active_drive_file_ids` columns,
replacing them with the authoritative `ConversationAttachment` join
table. Adds frontend components and hooks to make attach/detach
immediate, restores persisted attachments on conversation load, and
fixes several UX issues in the Resources tab. Also fixes a Docker
startup hang caused by the HuggingFace XET download protocol.

## Changes
- `chore(docker)`: add named `hf_cache` volume and set `HF_HUB_DISABLE_XET=1` on api, rag-cleanup, and rag-mcp services
- `feat(chat)`: drop `active_drive_file_id / active_drive_file_ids` columns; new `GET /conversations/{id}/attachments` endpoint; `drive_file_ids` field on `ChatCompletionRequest`; scope classifier receives file names; section SSE emitted before LLM stream; migration `87e049122b83`
- `feat(frontend)`: `PersistedFilesInfo` component (read-only list of persisted files); immediate `POST/DELETE /attachments` on file select/deselect; remove `drive_file` inline content blocks from `buildUserMessages`; load persisted attachments from new endpoint on conversation load
- `fix(frontend)`: Queued chip colour changed from amber to grey; Failed tooltip always rendered with fallback message; RAG status polling every 5 s for folders with in-progress files, stops on terminal state

## How to Test
1. `make clean && make up` — confirm the api container starts without hanging on embedding model download; confirm the `hf_cache` volume is created.
2. Open a new conversation, attach a Google Drive file via the file picker — confirm `POST /conversations/{id}/attachments` fires immediately.
3. Deselect the file — confirm `DELETE /conversations/{id}/attachments/{file_id}` fires immediately.
4. Send a message that references the attached document — confirm the completion request includes `drive_file_ids` and the response uses retrieved sections.
5. Reload the conversation — confirm the attached file list is restored from `GET /conversations/{id}/attachments`.
6. Navigate to the Resources tab with a folder that has files in Queued state — confirm the chip is grey, not amber.
7. Leave a file in Failed state — confirm the tooltip renders even with no `ragError` message.
8. Index a file that takes time — confirm the status chip updates to Ready without a page refresh (polling).
9. Run `make test` — all tests pass including `tests/chat/test_conversation_attachments.py` and `tests/chat/test_persistent_attachment.py`.

## Notes
**Migration required:** `87e049122b83_remove_active_drive_file_columns_from_.py` drops `active_drive_file_id` and `active_drive_file_ids` from the `conversation` table. Run `make migrations` after deploying.

**Breaking change:** `GET /conversations/{id}` no longer returns `active_drive_file_id` or `active_drive_file_ids` fields. Any client reading those fields must switch to `GET /conversations/{id}/attachments`. The old `DELETE /conversations/{id}/active-file` endpoint is removed; use `DELETE /conversations/{id}/attachments/{file_id}`.

**Docker:** The `hf_cache` named volume is new — existing `docker-compose up` will create it automatically. No manual steps required.


app/core_plugins/chat/routes.py
```diff
@@ -391,6 +329,32 @@ async def chat_completion(
...
+                current_user.id,
+                skipped,
+            )
+        for file_id in owned_ids:
+            await service.attach_drive_file(
+                session,
+                conversation_id=conversation.id,  # type: ignore
+                drive_file_id=file_id,
+            )
+
```
hamza-56, line 357: Heads up, there's no test covering the new `drive_file_ids` ownership-check branch added here. The flow validates IDs against `DriveFileModel.user_id == current_user.id AND is_deleted == False`, logs and skips unowned or deleted IDs, then calls `service.attach_drive_file` for each owned one.

A grep for `drive_file_ids` in `tests/` comes back empty. The new attachment tests in this PR cover the `/attachments` CRUD endpoints but never POST a chat completion with this field populated, so the ownership filter, the skipped-IDs log path, and the attach loop all go untested.

Since this is a security-relevant path and CLAUDE.md is pretty firm about writing tests before implementation, it would be good to add an integration test that POSTs to `/chat/completions` with `drive_file_ids` covering:
- all-owned IDs
- a mix of owned and unowned
- all unowned (should log a warning, not 4xx)

CLAUDE.md reference: https://github.com/edly-io/sparkth/blob/1739da6ab2584f5ccd53c70e1ceb872aa077e9a4/CLAUDE.md#L86-L99

app/llm/classifier.py
```diff
@@ -107,7 +115,13 @@ async def classify(self, query: str, history: list[HistoryTurn] | None = None) -
                 messages.append(AIMessage(content=content))
             elif role == "user":
                 messages.append(HumanMessage(content=content))
-        messages.append(HumanMessage(content=query))
+
+        # Prepend attachment context so the classifier knows files are in play
+        user_message = query
+        if attached_file_names:
+            file_list = ", ".join(f'"{n}"' for n in attached_file_names)
+            user_message = f"[The user has attached the following files to this conversation: {file_list}]\n\n{query}"
```
hamza-56, line 123: This adds a new `attached_file_names` parameter to `classify()` plus the message-prepending logic on lines 121-122, but `tests/llm/test_classifier.py` wasn't touched in this PR and has zero references to the parameter. The `f'"{n}"'` join formatting and the `[The user has attached the following files...]` envelope aren't exercised anywhere.

Quick fix: add two tests to `tests/llm/test_classifier.py`. One asserts the prepended-message format when `attached_file_names` is non-empty, and one regression test confirms `user_message` stays unchanged when the parameter is omitted or `None`.

Per the TDD rule in CLAUDE.md: https://github.com/edly-io/sparkth/blob/1739da6ab2584f5ccd53c70e1ceb872aa077e9a4/CLAUDE.md#L86-L99

app/rag/agent.py
```diff
@@ -125,6 +135,14 @@ async def run_agentic_rag_search(llm: Any, user_id: int, file_id: int, user_quer
...
         logger.exception("MCP client connection error for user %s, file %s", user_id, file_id)
         raise RAGRetrievalError(f"Failed to connect to RAG metadata server: {exc}") from exc
+    except GraphRecursionError as exc:
+        logger.error(
+            "RAG agent exceeded recursion limit (%d steps) for user %d, file %d",
+            _MAX_AGENT_RECURSION,
+            user_id,
+            file_id,
+        )
+        raise RAGRetrievalError("RAG agent exceeded maximum steps; query may be too complex.") from exc
```
hamza-56, line 145: The new `except GraphRecursionError` block (along with the `_MAX_AGENT_RECURSION` constant and the `config={"recursion_limit": ...}` on `agent.ainvoke`) doesn't have a test. Every other exception handler in `run_agentic_rag_search` (`ValidationError`, `ConnectError`, `HTTPStatusError`, `LangChainException`, `ValueError`) is covered in `test_agent.py`, so this one is the odd one out.

This is observable error-handling behavior, not just a config tweak, so it falls under the TDD rule. Easiest fix: mirror one of the existing `TestRunAgenticRAGSearch` exception-path tests, mock `agent.ainvoke` to raise `GraphRecursionError`, and assert the `RAGRetrievalError` plus the `logger.error` call.

CLAUDE.md: https://github.com/edly-io/sparkth/blob/1739da6ab2584f5ccd53c70e1ceb872aa077e9a4/CLAUDE.md#L86-L99

frontend/plugins/google-drive/GoogleDrive.tsx
```diff
@@ -657,6 +673,60 @@ export default function GoogleDrive() {
...
+        }
+      }
+    };
+
+    timerId = setTimeout(poll, BASE_DELAY);
+    return () => {
+      cancelled = true;
+      if (timerId) clearTimeout(timerId);
+    };
+  }, [token, folderIdsWithNonTerminal]);
```
hamza-56: This polling loop is doing roughly the same thing as `frontend/lib/useRagStatusPolling.ts`, which is already used by `FolderDetail` and `DriveFilePicker` and ships with a 12-test suite in `useRagStatusPolling.test.ts`.

If you reuse the existing hook you get:
- About 50 fewer lines of inline polling and backoff logic to maintain
- Existing coverage for cancellation, retry-on-error, and terminal-state stop (also the easiest way to satisfy CLAUDE.md TDD for this new behavior, since the hook is already tested)
- A cleaner `GoogleDrive.tsx` focused on UI and reducer concerns

If the existing hook's API doesn't quite fit (e.g. it polls per-file but this caller needs per-folder), extending the shared hook is probably better than forking it inline.

References:
- https://github.com/edly-io/sparkth/blob/1739da6ab2584f5ccd53c70e1ceb872aa077e9a4/frontend/lib/useRagStatusPolling.ts
- https://github.com/edly-io/sparkth/blob/1739da6ab2584f5ccd53c70e1ceb872aa077e9a4/frontend/lib/useRagStatusPolling.test.ts

frontend/plugins/chat/hooks/useChatInput.ts
```diff
@@ -53,35 +64,116 @@ export function useChatInput({ token, attachments, setAttachments, onSend }: Use
...
+            setAttachments((prev) => prev.filter((a) => a.driveFileDbId !== file.id));
+          });
+      }
+
+      const newIds = new Set(driveFiles.map((f) => f.id));
+      const removedAttachments = attachments.filter(
+        (a) => a.driveFileDbId !== undefined && !newIds.has(a.driveFileDbId!),
+      );
+      for (const att of removedAttachments) {
+        fetch(`/api/v1/chat/conversations/${conversationId}/attachments/${att.driveFileDbId}`, {
```
hamza-56, line 108: Small thing worth flagging: the `existingIds` set on line 80 and the `removedAttachments` filter on line 100 both read from the closed-over `attachments`, while the state update right above uses `setAttachments((prev) => ...)`. `attachments` reflects whatever the value was when this callback was last recreated, not the value at the moment the picker resolves.

If anything else mutates `attachments` between render and the picker close (a parallel upload, a programmatic clear, etc.), `existingIds` and `removedAttachments` end up working off stale state. The fallout would be spurious POSTs (re-attaching files already on the server) or spurious DELETEs (detaching files the user is still selecting).

The picker being modal keeps the window pretty tight in practice, but you can close the gap by deriving both sets from `prev` inside a `setAttachments((prev) => ...)` updater, or by tracking the latest value with a `useRef`.

frontend/__tests__/plugins/chat/components/attachment/PersistedFilesInfo.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+    expect(container.firstChild).toBeNull();
+  });
+
+  it("renders null when attachments array is empty", () => {
+    const { container } = render(
+      <PersistedFilesInfo attachments={[]} onDetachFile={mockOnDetachFile} />,
+    );
+
+    expect(container.firstChild).toBeNull();
+  });
```
hamza-56: These two tests are functionally identical: both render `<PersistedFilesInfo attachments={[]} ... />` and assert `container.firstChild` is null.

The first name suggests it's about filtering out non-Drive attachments, but the body doesn't actually construct any non-Drive entries. That scenario is already covered by the test on line 130 (`'does not render when only non-Drive attachments are present'`).

Either drop one of them or rewrite the first to actually pass non-Drive `TextAttachment`s.

```suggestion
  it("renders null when attachments array is empty", () => {
    const { container } = render(
      <PersistedFilesInfo attachments={[]} onDetachFile={mockOnDetachFile} />,
    );

    expect(container.firstChild).toBeNull();
  });
```

hamza-56 (approved):

app/llm/classifier.py
```diff
@@ -29,6 +29,8 @@
 - Analyzing a target audience for a course
 - Short answers, clarifications, or confirmations that are replies to the assistant's own course-creation questions
 - Publishing, managing, or retrieving courses on LMS platforms (Canvas, Open edX) — including creating, reading, listing, updating, or deleting courses, modules, pages, quizzes, and other course components via available tools
+- Loading, reading, summarising, or referencing attached documents or files — users attach documents to inform course design, so any message referencing "these documents", "this file", "the attached material", "load into context", etc. is IN SCOPE
+- Any message where the user has attached files to the conversation and is asking about their content, structure, or how to use them for course creation
```
zaira-bibi, line 33: I'm assuming the classifier only evaluates the user’s prompt when determining whether something is IN SCOPE. Interested in understanding how it handles cases where a user attaches a file containing malicious or prompt-injection-style instructions.

zaira-bibi (approved):

app/llm/classifier.py
abdulrafey1: yes

PR #327, opened 2026-05-15
fix(chat): restore streaming response after browser refresh or conversation switch

### LINKED ISSUE #297: [Sparkth] Response disappears on refresh
If page is refreshed while response is in progress (for eg during course publishing on Open edX), the entire message disappears

abdulrafey1:
## What
When the browser was refreshed or a conversation was switched mid-stream, the in-progress assistant response was permanently lost because the backend generator's lifecycle was tied directly to the SSE connection — a client disconnect cancelled the LLM call and nothing was written to the DB.

Additionally, provider API errors (e.g. Anthropic 529 Overloaded) and RAG intent router errors were not persisted to the DB, so refreshing a conversation after such an error showed a blank placeholder instead of the error card.

Finally, out-of-scope first messages were creating orphan conversation stubs in the DB (a conversation record with no useful content), causing the sidebar to fill with dead single-refusal entries.

### Response persistence on refresh / disconnect
- `fix(frontend)`: restore streaming message on page refresh — sessionStorage token buffer + polling until DB message lands
- `fix(chat)`: keep backend processing alive after browser refresh or conversation switch — asyncio background task decouples LLM/RAG work from SSE connection; RAG error messages fixed (DriveFileNotFoundError, RAGNotReadyError, RAGRetrievalError all produced empty str(exc)); attachment-aware placeholder text; frontend error guard changed from `if (parsed.error)` to `if (parsed.error !== undefined)` to stop falsy errors being swallowed
- `test(chat)`: add 4 new backend tests for RAG error messages and background task persistence

### Error card persistence
- `fix(chat)`: persist RAG phase errors to DB so reload shows error card directly — `RAGIntentRouterError` handler now calls `service.add_message(..., is_error=True)` before raising `HTTPException`
- `fix(chat)`: persist unexpected stream errors to DB so reload shows error card directly — `_PROVIDER_API_ERRORS` handler (covers Anthropic 529, OpenAI, Google errors) does the same

### Out-of-scope first message — no orphan conversations
- `fix(chat)`: make `ChatCompletionResponse.conversation_id` optional (`UUID | None = None`) for the stateless refusal path
- `fix(chat)`: skip conversation creation for out-of-scope first messages — pre-flight scope check (keyword tier-1 + LLM tier-2) runs before any DB write for new conversations; if out-of-scope, refusal is returned immediately with zero DB side effects; existing conversations are completely unaffected; `_scope_already_checked` flag prevents double LLM classifier call for new in-scope messages

## How to Test
1. Start a conversation with a file attachment and send a message that triggers RAG.
2. While the assistant is streaming, refresh the browser.
3. Verify the partial response is still visible (or a context-aware placeholder — "Still scanning your attached files..." — if no tokens arrived yet).
4. Wait up to ~3 s; verify the placeholder is replaced by the complete assistant message once the DB write completes.
5. Repeat steps 2–4 by switching to a different conversation mid-stream instead of refreshing.
6. Trigger a RAG error (e.g. detach a file mid-query) and confirm the error message is displayed on refresh rather than a blank placeholder.
7. Trigger a provider API error (e.g. use an invalid API key to force a 401/429) and confirm refreshing the conversation shows the error card.
8. Send a message that is clearly out-of-scope (e.g. "what is 2+2?") as the first message of a new conversation. Verify the refusal is shown and **no new conversation appears in the sidebar**.
9. Send an out-of-scope message in an **existing** conversation and verify the refusal is saved and shown on refresh.
10. Run `make test` — all 1057 tests must pass.

## Notes
This branch targets `rafey/rag-intent-router-frontend` (not `main`) because it depends on DB migrations and backend changes introduced in that branch.

Closes #297


app/core_plugins/chat/routes.py
```diff
@@ -629,237 +630,257 @@ async def stream_chat_response(
...
+                    conversation_id=conversation_id,
+                    role="assistant",
+                    content=user_message,
+                    is_error=True,
+                )
+            except SQLAlchemyError:
+                logger.exception("Failed to persist unexpected error for conversation %s", conversation_id)
+            await queue.put(json.dumps({"error": user_message, "done": True}))
+
+    asyncio.ensure_future(_process_and_stream())
```
zaira-bibi: `ensure_future` is no longer the recommended way to schedule coroutines. Replace with:

```python
asyncio.create_task(_process_and_stream())
```

app/core_plugins/chat/routes.py
```diff
@@ -629,237 +630,257 @@ async def stream_chat_response(
...
+    # refresh) does not cancel the work. The generator only drains a queue.
+    conversation_id: int = conversation.id  # type: ignore[assignment]
+    conversation_uuid: str = str(conversation.uuid)
+    queue: asyncio.Queue[str | None] = asyncio.Queue()
 
-    # Emit skipping_rag event if router said to skip but attachments existed
-    if not should_run_rag and rag_routing_reason is not None:
-        yield f"data: {json.dumps({'status': 'skipping_rag', 'reason': rag_routing_reason, 'done': False})}\n\n"
+    async def _process_and_stream() -> None:
+        async with AsyncSession(async_engine, expire_on_commit=False) as bg_session:
```
zaira-bibi, line 696: Every database call in `_run` uses `bg_session`, and the injected session is silently ignored. Either remove the `session` parameter in `stream_chat_response` and update call sites, or document it explicitly as deprecated/unused.

frontend/plugins/chat/hooks/useChatStream.ts
```diff
@@ -169,23 +225,33 @@ async function readStream(
...
+          if (parsed.status === "scanning_attachments" || parsed.status === "searching_document") {
+            saveStreamProgress(conversationId, assistantText, parsed.status as string);
+          }
           applyStatusEvent(parsed, assistantId, setMessages);
           continue;
         }
 
         if (parsed.token) {
           assistantText += parsed.token;
+          saveStreamProgress(conversationId, assistantText);
```
zaira-bibi: `saveStreamProgress` is called on every token. `sessionStorage.setItem` is synchronous and serializes the entire accumulated string on each call. For a 2000-token response that's 2000 blocking serialization writes, which will cause jank on slower devices. Maybe throttle it? For example, every 50 tokens or every 500ms, and flush once unconditionally at stream end.

frontend/plugins/chat/hooks/useConversation.ts
```diff
@@ -176,6 +221,59 @@ export function useConversation(
...
+    if (!hasPendingPlaceholder || !conversationId || !token) return;
+
+    const POLL_INTERVAL_MS = 3_000;
+    const POLL_TIMEOUT_MS = 5 * 60 * 1000; // stop after 5 min regardless
+    const startedAt = Date.now();
+
+    const intervalId = setInterval(async () => {
+      if (Date.now() - startedAt > POLL_TIMEOUT_MS) {
+        clearInterval(intervalId);
+        return;
```
zaira-bibi, line 272: Polling timeout silently leaves a stale placeholder. After 5 minutes of failed polling (e.g. the backend task crashed), the user is still looking at "Your response is still being generated. It will appear here automatically once ready." with no indication it failed. Replace the placeholder with an error message on timeout:

```python
setMessages((prev) => prev.map((m) =>
  RECOVERY_PLACEHOLDER_IDS.has(m.id)
    ? { ...m, content: "The response could not be recovered. Please try again.", isError: true }
    : m
));
```

frontend/plugins/chat/hooks/useConversation.ts
```diff
@@ -176,6 +221,59 @@ export function useConversation(
...
+      }
+      try {
+        const r = await fetch(`/api/v1/chat/conversations/${conversationId}`, {
```
zaira-bibi: This returns all messages in the conversation. At `POLL_INTERVAL_MS = 3_000` for up to 5 minutes, that's up to ~100 full-conversation fetches per recovery attempt. For long conversations this is significant wasted bandwidth and server load. A lightweight endpoint returning only the last message (or a `?last_only=true` query param) would be sufficient here.

zaira-bibi (changes requested):

frontend/plugins/chat/hooks/useConversation.ts
```diff
@@ -176,6 +221,59 @@ export function useConversation(
...
+          ];
+        });
+        clearInterval(intervalId);
+      } catch {
+        // network error — will retry on next tick
+      }
+    }, POLL_INTERVAL_MS);
+
+    return () => clearInterval(intervalId);
+  }, [hasPendingPlaceholder, conversationId, token, setMessages]);
```
hamza-56, line 316: This polling fetch is bound only to the interval timer, not to the lifetime of the effect. If the user switches conversations while a poll is in flight, the cleanup clears the interval but the in-flight `fetch(/api/v1/chat/conversations/${conversationId})` keeps running.

When it resolves, the captured `conversationId` is still the old conversation, but `setMessages` writes into the parent's current `historyState`, which is now displaying a different conversation. The placeholder filter is a no-op for the new conversation, so the old conversation's last assistant message gets appended to the new conversation's view.

Worse, `clearStreamProgress(conversationId)` on line 252 will wipe any partial-stream buffer for the original conversation while the user is somewhere else.

Easiest fix: use an `AbortController` for the polling fetch and abort it in the cleanup (same pattern `loadConversation` already uses). Or compare the captured `conversationId` against a ref of the currently active one inside the resolve handler before touching state.

app/core_plugins/chat/routes.py
hamza-56: A couple of related concerns about how this background task is owned:

1. The result of `asyncio.ensure_future(_process_and_stream())` is dropped on the floor. The asyncio docs are explicit: "the event loop only keeps weak references to tasks. Save a reference to the result of this function, to avoid a task disappearing mid-execution." Today it works because the consumer (this generator) is parked on `queue.get()` and that frame keeps things alive indirectly, but the link is fragile. Worth assigning to a variable: `task = asyncio.create_task(_process_and_stream())`.

2. When the SSE client disconnects (browser refresh), FastAPI cancels this generator at `await queue.get()`. The background task keeps running, which is the whole point of the refactor, but the queue is now consumer-less. The task keeps `queue.put()`-ing the rest of the LLM stream into an unbounded `asyncio.Queue` until it finishes, holding the entire response in memory per abandoned stream. For long responses this is a transient memory spike that scales with concurrent abandonment.

For the queue accumulation, a couple of options: use a bounded queue with `put_nowait` plus drop-on-full once we detect no consumer, or skip the queue puts after a disconnect signal (e.g. a `connected` flag flipped from `False` when the generator catches `CancelledError`).

Minor: the rest of the codebase uses `asyncio.create_task` rather than `ensure_future` (see `app/main.py`, `app/llm/providers.py`, `app/rag/memory_profiler.py`). Switching here would be more consistent.

frontend/plugins/chat/hooks/useChatStream.ts
hamza-56: `saveStreamProgress` runs on every single token with the full accumulated `assistantText`. Each call does a `JSON.stringify` plus a synchronous `sessionStorage.setItem`. For an N-token response, total bytes written is roughly O(N²). A 10k-token response can write hundreds of MB synchronously on the main thread.

Browsers also cap sessionStorage at around 5-10 MB. Once `assistantText` exceeds the quota, every subsequent `setItem` throws `QuotaExceededError`, which is silently swallowed by the empty `catch {}` block in [`saveStreamProgress`](https://github.com/edly-io/sparkth/blob/a98e49ad2765c6a7599d77845d0a87a6a12ea79e/frontend/plugins/chat/hooks/useChatStream.ts#L147-L158). The user gets no warning and the refresh-recovery feature quietly stops working for the rest of the response.

Two fixes that compose well:
- Throttle or debounce the writes (every 250 ms or every Nth token instead of every one).
- Store the delta (or just the most recent chunk plus a count) rather than the full accumulated text.

frontend/plugins/chat/hooks/useConversation.ts
```diff
@@ -176,6 +221,59 @@ export function useConversation(
...
+
+    const POLL_INTERVAL_MS = 3_000;
+    const POLL_TIMEOUT_MS = 5 * 60 * 1000; // stop after 5 min regardless
+    const startedAt = Date.now();
+
+    const intervalId = setInterval(async () => {
+      if (Date.now() - startedAt > POLL_TIMEOUT_MS) {
+        clearInterval(intervalId);
+        return;
+      }
```
hamza-56, line 273: When polling times out after 5 minutes without ever seeing an assistant message, the interval is cleared but the recovery placeholder ("Your response is still being generated...") stays in the message list indefinitely. The user is left staring at a misleading "still generating" message with no signal that the poller gave up.

This is exactly the failure mode this feature was designed to surface (background task killed by OOM, worker restart, infra outage, anything that prevents the assistant message from ever landing in the DB). Right now those failures look identical to "still streaming" forever from the user's perspective.

On timeout the placeholder should be replaced with a real error message ("Response generation timed out, please try sending your message again") and `clearStreamProgress(conversationId)` should be called to flush the stale sessionStorage buffer.

app/core_plugins/chat/routes.py
```diff
@@ -629,237 +630,257 @@ async def stream_chat_response(
...
+            try:
+                await _run(bg_session)
+            except Exception:
+                logger.exception("Unhandled error in stream task for conversation %s", conversation_id)
+                await queue.put(json.dumps({"error": "An unexpected error occurred. Please try again.", "done": True}))
+            finally:
+                await queue.put(None)
```
hamza-56: Heads up, CLAUDE.md is pretty firm about this: _"Never use bare `except Exception` blocks. Always catch specific exception types."_ ([CLAUDE.md#L138-L150](https://github.com/edly-io/sparkth/blob/a98e49ad2765c6a7599d77845d0a87a6a12ea79e/CLAUDE.md#L138-L150))

That said, this catch is doing real work as a fail-safe before the queue's `None` sentinel. Without something to catch escapes from `_run`, the consumer would block forever on `await queue.get()` if anything unexpected propagates out, so removing the catch entirely isn't the move. The cleaner options:

- Enumerate the remaining types you actually expect to leak out of `_run` (anything `_run` doesn't already handle specifically). Tedious but explicit.
- Switch to `except BaseException` with an inline comment noting this is the fire-and-forget task boundary. `BaseException` also catches `CancelledError`, which matters for asyncio tasks: if the task is cancelled here you currently won't run the `finally` cleanly because `CancelledError` is a `BaseException`, not an `Exception`.

At minimum, please add a short comment so future readers understand the broad catch is deliberate and not an oversight.

tests/chat/test_rag_status_events.py
```diff
@@ -344,3 +347,129 @@ def test_message_response_rag_sections_defaults_to_none(self) -> None:
...
+        session=AsyncMock(),
+        tools=None,
+    )
+
+    # Consume only the first event then abandon the generator
+    async for _ in gen:
+        break
+
+    # Give the background task time to finish
+    await asyncio.sleep(0.1)
```
hamza-56: This `asyncio.sleep(0.1)` is timing-race synchronization. On a loaded CI machine the background task may not finish in 100 ms and the assertion below will fail intermittently with no useful signal as to why.

The deterministic alternative is to await the task directly. If `stream_chat_response` (or the `_process_and_stream` helper) is refactored to expose the Task handle (or to accept a "done" event/future), this test could `await task` instead of sleeping. That refactor also helps in production: storing the Task handle is part of the fix for the weak-ref concern flagged separately on `routes.py:877`.

hamza-56 (approved):

zaira-bibi (approved):
LGTM!

PR #328, opened 2026-05-18
feat(slack): slack UI and comprehensive log improvements

### LINKED ISSUE #246: Slack TA Bot Frontend: Plugin UI (connect, configure, logs)
Implements the frontend plugin UI for the Slack TA Bot, following the existing plugin UI pattern in `frontend/plugins/`.

Depends on the OAuth connection issue and the Events/RAG issue.

### What's included:

- New plugin UI at `frontend/plugins/slack/` registered in `frontend/lib/plugins/registry.ts`
- **OAuth connect/disconnect flow**: "Connect to Slack" button → redirects to install URL → shows connected workspace name on return
- After OAuth connection, show a document picker (sourced from `GET /rag/sources`) so the teacher can select which docs this bot can access
- **Bot configuration form**: `bot_name` and `fallback_message` fields, saved via the existing user-plugin config API. Above selection is also saved as `allowed_sources` in the bot's plugin config
- **Response logs table**: Paginated view of `GET /logs`, showing question, answer, RAG match status, timestamp
- **Connection status banner**: Shows workspace name when connected, prompt to connect when not

### Acceptance Criteria

 - [ ] Plugin appears in the plugins dashboard under "Slack TA Bot"
 - [ ] "Connect to Slack" button redirects to the Slack install URL returned by `GET /oauth/authorize`
 - [ ] After OAuth callback (`?connected=true`), UI shows connected workspace name and hides the connect button
 - [ ] A document picker (sourced from `GET /rag/sources`) enables the teacher to select which docs this bot can access
 - [ ] "Disconnect" button calls `DELETE /oauth/disconnect` and resets the UI to disconnected state
 - [ ] Bot name and fallback message fields are pre-populated from saved config
 - [ ] Saving config calls the `user-plugin` config API and shows success/error feedback
 - [ ] Logs table shows question, answer (or "No match"), RAG match badge, and timestamp
 - [ ] Logs table paginates correctly (next/prev)
 - [ ] Logs table shows empty state when no responses yet
 - [ ] All API errors surface as user-facing error messages (no silent failures)

zaira-bibi:
## What

Completes the Slack plugin with two sets of changes:

**Slack UI** — config modal now has a proper textarea for system prompts and a doc source picker for selecting RAG knowledge bases. LogStream is fully implemented with real-time polling and pagination.

**Log improvements** — three gaps in the logging system fixed:
- Bot replies for greeting, config-incomplete, and plugin-disabled paths were silently dropped. Every `_dispatch_event` path now logs a `BotResponseLog` row with a `response_type` enum value.
- OAuth connect and disconnect events now produce `SlackConnectionLog` rows, giving instructors a full connection audit trail.
- Logs displayed raw Slack IDs. Slack user and channel names are now resolved at log time. Timestamps include timezone context.

Closes #246.

### Frontend
| File | Change |
|---|---|
| `frontend/components/ui/Textarea.tsx` | New reusable textarea component |
| `frontend/plugins/slack/components/DocSourcePicker.tsx` | New component for selecting RAG knowledge sources in Slack config |
| `frontend/plugins/slack/components/LogStream.tsx` | Full implementation: merged log stream, `MessageRow` / `ConnectionRow`, `responseTypeLabel`, timezone timestamps, `stateRef` polling pattern |
| `frontend/plugins/slack/SlackPlugin.tsx` | Config UI updates |
| `frontend/plugins/slack/components/SlackConfigModal.tsx` | Uses Textarea and DocSourcePicker |
| `frontend/lib/slack-api.ts` | `BotResponseLogItem` gains `type`, name fields, `response_type`; `ConnectionLogItem` and `LogItem` union added; `next_cursor` / `cursor` changed to `string` |

### Backend
| File | Change |
|---|---|
| `app/core_plugins/slack/models.py` | `ResponseType` (5 values) and `ConnectionEventType` enums; `SlackConnectionLog` model; three new columns on `BotResponseLog` |
| `app/core_plugins/slack/client.py` | `get_user_display_name()` and `get_channel_name()` — both return `None` on any failure |
| `app/core_plugins/slack/routes.py` | `_resolve_names` helper; all `_dispatch_event` branches log; connect/disconnect insert `SlackConnectionLog`; `/logs` rewritten as merged timestamp-paginated stream |
| `app/core_plugins/slack/types.py` | `BotResponseLogItem` updated; `ConnectionLogItem` and `LogItem` union added; `LogsResponse.next_cursor` → `str \| None` |
| `app/migrations/versions/d300f558a4b4_slack_log_improvements.py` | Adds `response_type NOT NULL DEFAULT 'rag_match'`, `slack_user_name`, `slack_channel_name` columns; creates `slack_connection_logs` table |
| `tests/slack/test_models.py` | New — 9 tests for new enums and models |
| `tests/slack/test_client.py` | +10 tests for `get_user_display_name` and `get_channel_name` |
| `tests/slack/test_routes.py` | `TestDispatchEventLogging` (5 tests), `TestConnectionEventLogging` (2 tests), `TestGetLogs` rewritten (11 tests) |

## Test plan

- [ ] `make test` — 982 tests, all pass
- [ ] `make mypy` — clean (204 source files)
- [ ] `make lint` — clean
- [ ] `make frontend.lint` — 0 warnings / errors
- [ ] `make migrations` before deploying (migration required)
- [ ] Manual: connect a workspace → "Connected to …" row appears in log stream
- [ ] Manual: send a bot mention → log entry shows resolved display name, channel name, response type label, and timezone in timestamp
- [ ] Manual: send a message when plugin is disabled / config incomplete → log entry appears with amber label
- [ ] Manual: disconnect → "Disconnected from …" row appears in log stream
- [ ] Manual: paginate logs with "Load older" — cursor advances correctly

## Migration notes

Run `make migrations` before deploying. Existing `BotResponseLog` rows receive `response_type = 'rag_match'` via server default. `slack_user_name` and `slack_channel_name` will be `NULL` for historical rows — the frontend falls back to raw Slack IDs for those entries.


frontend/plugins/slack/components/LogStream.tsx
```diff
@@ -0,0 +1,364 @@
...
+      };
+    case "INITIAL_FAILURE":
+      return { ...state, initialLoading: false, error: action.error };
+    case "POLL_START":
+      return { ...state, polling: true };
+    case "POLL_SUCCESS":
+      return {
+        ...state,
+        polling: false,
+        entries: [...[...action.newItems].reverse(), ...state.entries],
```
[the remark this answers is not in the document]
zaira-bibi: @claude The inner `[...action.newItems]` before `.reverse()` is intentional: reducers must be pure and shouldn't mutate action objects. Removing it would make `action.newItems` reversed as a side effect, which violates the reducer purity contract even if harmless today.

app/core_plugins/slack/routes.py
```diff
@@ -251,6 +300,8 @@ async def _dispatch_event(
     slack_user = event.get("user", "")
     thread_ts: str | None = event.get("thread_ts") or event.get("ts")
 
+    slack_user_name, slack_channel_name = await _resolve_names(bot_token_encrypted, slack_user, channel)
```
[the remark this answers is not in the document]
zaira-bibi: @claude Deferring `_resolve_names` past the plugin checks would mean either calling it separately in each early-exit branch (more duplication) or passing `None` names into logs for those paths. The trade-off isn't worth the added complexity right now.

app/core_plugins/slack/routes.py
```diff
@@ -432,44 +545,88 @@ def get_response_logs(
...
+            ) from exc
+
+    msg_stmt = (
+        select(BotResponseLog)
+        .where(col(BotResponseLog.workspace_id) == workspace.id)
+        .order_by(col(BotResponseLog.created_at).desc())
+        .limit(limit + 1)
+    )
+    if cursor_dt is not None:
+        msg_stmt = msg_stmt.where(col(BotResponseLog.created_at) < cursor_dt)
```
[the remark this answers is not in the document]
zaira-bibi: @claude Composite cursor is a reasonable future improvement but "extremely unlikely" with microsecond timestamps - skipping for now.

app/core_plugins/slack/routes.py
```diff
@@ -238,6 +270,57 @@ async def _post_slack_message(
...
+    bot_token_encrypted: str,
+    slack_user: str,
+    channel: str,
+) -> tuple[str | None, str | None]:
+    """Resolve Slack user and channel display names. Returns (None, None) on any failure."""
+    try:
+        bot_token = decrypt_token(bot_token_encrypted)
+        async with SlackClient(bot_token) as slack:
+            user_name = await slack.get_user_display_name(slack_user)
+            channel_name = await slack.get_channel_name(channel)
```
hamza-56: These two Slack API calls run serially. Each has a 10s timeout in `SlackClient`, so worst-case `_resolve_names` blocks `_dispatch_event` for ~20s before the bot can post any reply (greeting, plugin-disabled, config-incomplete, RAG, anything).

Wrapping both in `asyncio.gather` halves the worst case to ~10s without any other change:

```python
async with SlackClient(bot_token) as slack:
    user_name, channel_name = await asyncio.gather(
        slack.get_user_display_name(slack_user),
        slack.get_channel_name(channel),
    )
```

This matters because `_resolve_names` is called at the top of `_dispatch_event` (line 337) before any plugin or config check, so every event pays the full cost even on fast-path early-return branches where we just want to tell the user the bot isn't configured.

app/core_plugins/slack/routes.py
```diff
@@ -432,44 +554,88 @@ def get_response_logs(
...
-    has_more = len(rows) > limit
+    has_more = len(all_items) > limit
     if has_more:
-        rows = rows[:limit]
+        all_items = all_items[:limit]
 
-    items = [_to_log_item(r) for r in rows]
-    next_cursor = rows[-1].id if rows else None
-    return LogsResponse(items=items, total=len(items), next_cursor=next_cursor, has_more=has_more)
+    next_cursor_val = all_items[-1].created_at.isoformat() if all_items else None
```
hamza-56: Pagination boundary issue: the next-page filter is `col(...).created_at < cursor_dt` (strict less-than) and `next_cursor` is just `all_items[-1].created_at.isoformat()` with no secondary tiebreaker. If two rows in either table share a `created_at` value (microsecond collision under load, or a connect event and a fast-following message log written in the same instant), they can land on different sides of the boundary and the strict `<` will silently drop the ones at exactly `cursor_dt`.

There's also no `id` tiebreaker in the `order_by` clauses, so the order between tied rows on the same page is non-deterministic across requests.

Recommended pattern: order by `(created_at desc, id desc)` in each query, tie-break the post-merge sort by `(type, id)` as well, and emit a composite cursor like `f"{created_at.isoformat()}|{id}|{type}"`, then filter with a tuple comparison `(created_at, id) < (cursor_ts, cursor_id)`.

Likelihood is low on PostgreSQL `timestamptz` (microsecond resolution) but it does happen under load, and the resulting dropped rows would never reappear in any subsequent page.

app/migrations/versions/d300f558a4b4_slack_log_improvements.py
```diff
@@ -0,0 +1,54 @@
...
+    sa.Column('workspace_id', sa.Integer(), nullable=False),
+    sa.Column('event_type', sa.Enum('connected', 'disconnected', name='connectioneventtype'), nullable=False),
+    sa.Column('team_name', sqlmodel.sql.sqltypes.AutoString(length=255), nullable=True),
+    sa.ForeignKeyConstraint(['workspace_id'], ['slack_workspaces.id'], ),
+    sa.PrimaryKeyConstraint('id')
+    )
+    op.create_index(op.f('ix_slack_connection_logs_workspace_id'), 'slack_connection_logs', ['workspace_id'], unique=False)
+    responsetype = sa.Enum('rag_match', 'fallback', 'greeting', 'config_incomplete', 'plugin_disabled', name='responsetype')
+    responsetype.create(op.get_bind(), checkfirst=True)
+    op.add_column('slack_bot_response_logs', sa.Column('response_type', sa.Enum('rag_match', 'fallback', 'greeting', 'config_incomplete', 'plugin_disabled', name='responsetype'), server_default='rag_match', nullable=False))
```
hamza-56: The `server_default='rag_match'` will label every existing row in `slack_bot_response_logs` as a RAG match after the migration runs, including historical rows where `rag_matched=False` (fallback replies, no-match responses).

In the frontend, [`LogStream.tsx:109`](https://github.com/edly-io/sparkth/blob/e8743e8d35764bd1eb9326fccd510eece7d0975d/frontend/plugins/slack/components/LogStream.tsx#L109) translates `response_type === "rag_match"` into a green "matched" pill. So after deploy, every pre-migration fallback reply will be falsely shown as a successful RAG match.

Quick fix is to add a one-line backfill right after the `add_column` and before the column becomes the source of truth:

```python
op.execute("UPDATE slack_bot_response_logs SET response_type='fallback' WHERE rag_matched = false")
```

Alternatively, set the server default to a neutral value like `'legacy'` and surface that neutrally in the UI.

app/core_plugins/slack/routes.py
```diff
@@ -261,56 +346,94 @@ async def _dispatch_event(
...
+                    slack_user=slack_user,
+                    posted_ts=posted_ts,
+                    question=question,
+                    answer=reply,
+                    rag_matched=False,
+                    response_type=ResponseType.plugin_disabled,
+                    slack_user_name=slack_user_name,
+                    slack_channel_name=slack_channel_name,
+                )
             return
```
hamza-56, line 401: The plugin-not-configured, plugin-disabled, config-empty, and config-invalid branches in `_dispatch_event` all follow the same shape: build a reply string, call `_post_slack_message`, and if posting succeeded log a `BotResponseLog` row with a specific `response_type`. The only things that vary are the reply text and the `response_type`.

Extracting that into a small helper would cut a lot of repeated structure and make the dispatch flow much easier to scan:

```python
async def _reply_and_log(
    session, *, reply, response_type, workspace_id, channel,
    slack_user, thread_ts, bot_token_encrypted, question,
    slack_user_name, slack_channel_name,
) -> None:
    posted_ts = await _post_slack_message(
        bot_token_encrypted, channel, reply, thread_ts, workspace_id
    )
    if posted_ts is not None:
        await _log_response(
            session,
            workspace_id=workspace_id,
            channel=channel,
            slack_user=slack_user,
            posted_ts=posted_ts,
            question=question,
            answer=reply,
            rag_matched=False,
            response_type=response_type,
            slack_user_name=slack_user_name,
            slack_channel_name=slack_channel_name,
        )
```

Each early-return branch then becomes a one-liner call. Reviewers no longer have to diff four ~12-line blocks to confirm they're structurally identical.

tests/slack/test_routes.py
```diff
@@ -289,130 +301,230 @@ async def test_returns_existing_log_entry(
...
+            slack_ts="100.0",
+            question="q1",
+            answer="a1",
+            rag_matched=False,
+            response_type=ResponseType.fallback,
+        )
+        sync_session.add(msg_log)
+        sync_session.commit()
+
+        time.sleep(0.01)
```
hamza-56: These `time.sleep(0.01)` calls are using wall-clock to force `created_at` ordering between rows inserted close together. That's fragile on a loaded CI runner where 10 ms can easily get squeezed by the scheduler, and the tests then start failing for reasons totally unrelated to what they're verifying.

The deterministic alternative is to pass explicit `created_at` values to each model constructor and rely on the known ordering:

```python
from datetime import datetime, timedelta, timezone

base = datetime.now(timezone.utc)
msg_log = BotResponseLog(..., created_at=base - timedelta(seconds=1))
conn_log = SlackConnectionLog(..., created_at=base)
```

The same pattern works for all five sleeps in this file. It also makes the tests roughly 50 ms faster overall, which adds up across the suite.

frontend/plugins/slack/components/LogStream.tsx
hamza-56: The double-spread plus inline `.reverse()` is doing more than it looks at a glance: `[...action.newItems]` clones the array (so the reverse doesn't mutate the action payload), `.reverse()` flips it in place on the clone, then the outer spread spreads the result. A named intermediate makes the intent self-documenting:

```ts
const newestFirst = [...action.newItems].reverse();
return { ...state, entries: [...newestFirst, ...state.entries] };
```

Or at least a one-line comment explaining why we reverse: the backend's `since_id` endpoint returns items ascending by id, and we need them prepended newest-first to match the rest of the list.

frontend/plugins/slack/components/LogStream.tsx
```diff
@@ -0,0 +1,364 @@
...
+    if (!token) return;
+    const { initialLoading, polling, loadingOlder, entries } = stateRef.current;
+    if (initialLoading || polling || loadingOlder) return;
+
+    const sinceId = entries
+      .filter((e): e is BotResponseLogItem => e.type === "message")
+      .reduce<number | undefined>(
+        (max, e) => (max === undefined || e.id > max ? e.id : max),
+        undefined,
+      );
```
hamza-56: The `reduce<number | undefined>` exists only to distinguish "no messages" from "max id is 0", which is valid but obscure. Since `entries` comes from the server sorted newest-first (highest id at the front), filtering to messages and taking the first element is both shorter and self-documenting:

```ts
const msgEntries = entries.filter(
  (e): e is BotResponseLogItem => e.type === "message",
);
const sinceId = msgEntries.length > 0 ? msgEntries[0].id : undefined;
```

This also avoids walking the full `entries` list to compute a max that's always at the front.

hamza-56 (approved):

PR #351, opened 2026-05-21
feat(slack): replace semantic RAG with per-file agentic retrieval

### LINKED ISSUE #345: [Slack bot] Replace Semantic RAG with Agentic RAG in Slack TA Bot
The Slack TA Bot currently uses semantic similarity search to retrieve relevant document chunks from the database. This approach has the same limitation as the previous system: queries fail when user intent is clear but no matching keywords are present (e.g., "What does the syllabus say about late submissions?" when "syllabus" and "late submissions" don't appear together in any chunk).

The bot should be updated to use the same agentic RAG approach being implemented for the main system, where an LLM agent uses a read-only MCP server to intelligently select relevant document content based on query intent rather than pure keyword/semantic matching.

**Proposed Solution**
Integrate the read-only MCP server (built on top of the synced documents database) into the Slack TA Bot's retrieval pipeline. The agent will:

- Analyze the student's question to understand intent
- Access document metadata through the MCP server
- Select relevant document sections based on semantic understanding
- Return targeted content for LLM synthesis

This replaces the current semantic similarity search while maintaining the existing LLM synthesis flow introduced in the ticket #305 .

**Acceptance Criteria**

- [x] The Slack TA Bot uses the read-only MCP server for document retrieval instead of direct semantic similarity search
- [x] Agent-based retrieval correctly handles intent-based queries that lack explicit keywords (e.g., "What's the policy on X?" maps to relevant policy documents)
- [x] When the agent retrieves content, it's passed to the LLM synthesis layer (if `llm_config_id` is configured) exactly as raw chunks were previously
- [x] Bots with `llm_config_id = null` receive agent-retrieved chunks in the formatted raw-chunks response (backward compatible)
- [x] Agent retrieval failures (timeout, MCP server errors, no relevant documents found) return a helpful "no relevant content found" message
- [x]  The agent respects document access permissions synced from the source (students only see content they're authorized to access)
- [x]  Response latency remains acceptable (agent + MCP + synthesis completes within reasonable Slack response time expectations)
- [x] All existing Slack bot tests continue to pass with updated retrieval logic

zaira-bibi:
## What
Replaces the Slack TA Bot's similarity-based retrieval with the per-file MCP-driven agentic flow the chat plugin already uses. The bot runs on the instructor's own configured AI key.

Closes #345 

## Changes

**Agentic retrieval**
- Rewrites `answer_question` to fan out one `get_context_via_agent` call per resolved file concurrently via `asyncio.gather`
- Adds `_resolve_files_for_sources` (maps `allowed_sources` → RAG-ready DriveFile IDs, capped at 5 when no sources are configured) and `_run_agent_fan_out` (concurrency wrapper)

**Per-failure response messages**
- Each failure mode now returns a distinct hard-coded message and `ResponseType` value instead of falling through to `config.fallback_message`:
  - `no_files_resolved` — no RAG-ready documents found
  - `rag_not_ready` — indexing still in progress
  - `drive_file_not_found` — document no longer accessible
  - `retrieval_error` — unexpected error during retrieval
  - `config_incomplete` — no AI key configured or key unresolvable
- `config.fallback_message` is now reserved exclusively for "agent ran successfully but found no relevant content"
- New messages for config failures live in `slack/constants.py` (`NO_AI_KEY_MESSAGE`, `AI_KEY_UNAVAILABLE_MESSAGE`)

**User LLM key 
- Removes `app/llm/system.py` (`get_system_llm_provider`)
- `_dispatch_event` now builds both the agent LLM and the synthesis provider from the instructor's `llm_config_id`; if unconfigured or unresolvable, replies with the appropriate hard-coded message
- `generate_conversation_title` now accepts `provider: BaseChatProvider` directly; the chat routes build a `title_provider` from the already-resolved `llm_config` and pass it in

**Migration**
- `2b8cd7530d90`: extends the `responsetype` PostgreSQL enum with the four new per-failure values (`ALTER TYPE … ADD VALUE IF NOT EXISTS`)

## How to Test
1. Run `make migrations` to apply the `responsetype` enum extension.
2. Run `make test`. Expect green.
3. `make dev.up`, connect a Slack workspace, set `llm_config_id` in the bot config with at least one `allowed_sources` entry pointing to a RAG-ready DriveFile.
4. Mention the bot with a question and confirm:
   - Reply uses agent-retrieved content (log: `Slack agentic RAG: user=…`).
   - With `llm_config_id` set, the reply is synthesized; without it, the bot replies with "I'm not configured with an AI key yet."
   - Remove the DriveFile from RAG sources → bot replies with the "no documents available" message, not the fallback.
5. Trigger conversation title generation in the chat plugin and confirm it uses the instructor's own LLM key.

## Notes
**Breaking env-var change:** remove `CHAT_TITLE_GENERATION_API_KEY` /`_MODEL` / `_PROVIDER` from any deployed `.env`.
Neither feature has a platform-paid fallback anymore — both depend on the instructor's configured AI key.

**Migration required:** `2b8cd7530d90` extends the `responsetype` PostgreSQL enum. Run `make migrations` before deploying. Non-destructive on Postgres; no-op on SQLite (CI).

### DIFF
--- a/app/core_plugins/chat/routes.py
+++ b/app/core_plugins/chat/routes.py
@@ -355,6 +355,8 @@ async def chat_completion(
                 provider_name=provider_name,
                 api_key=api_key,
                 model=model,
+                temperature=0.3,
+                max_tool_executions=0,
             )
             background_tasks.add_task(
                 generate_conversation_title,
--- a/app/core_plugins/chat/routes.py
+++ b/app/core_plugins/chat/routes.py
@@ -351,12 +351,18 @@ async def chat_completion(
 
         first_user_text = get_first_user_text(request.messages)
         if first_user_text:
+            title_provider = get_provider(
+                provider_name=provider_name,
+                api_key=api_key,
+                model=model,
+            )
             background_tasks.add_task(
                 generate_conversation_title,
                 conversation_id=conversation.id,  # type: ignore
                 user_id=current_user.id,  # type: ignore
                 first_user_message=first_user_text,
                 service=service,
+                provider=title_provider,
             )
 
     # Attach any drive files included with the request (covers new-conversation flow
--- a/app/core_plugins/slack/rag.py
+++ b/app/core_plugins/slack/rag.py
@@ -1,26 +1,102 @@
-"""RAG dispatch for the Slack TA Bot plugin."""
+"""RAG dispatch for the Slack TA Bot plugin — agentic retrieval."""
+
+import asyncio
+from typing import Any
 
 import httpx
 from langchain_core.exceptions import LangChainException
 from pydantic import ValidationError
+from sqlmodel import col, select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.core.logger import get_logger
 from app.core_plugins.slack.config import SlackConfig
+from app.core_plugins.slack.models import ResponseType
 from app.core_plugins.slack.synthesis import synthesize_answer
 from app.llm.providers import BaseChatProvider
+from app.models.drive import DriveFile
 from app.rag.context_service import RAGContextService, format_chunks_as_context
-from app.rag.embeddings import BaseEmbeddingProvider
-from app.rag.provider import get_provider
+from app.rag.exceptions import (
+    DriveFileNotFoundError,
+    RAGNotReadyError,
+    RAGRetrievalError,
+)
 from app.rag.store import SimilarityResult
+from app.rag.types import RAGContext, RagStatus
+from app.rag.utils import resolve_source_name
 
 logger = get_logger(__name__)
 
-# Module-level singleton — embedding model and vector store loaded once per process
+# Module-level singleton — vector store and embedding provider loaded once per process.
+# Reused only for its agentic entry point (get_context_via_agent); no semantic search runs.
 _rag_service: RAGContextService = RAGContextService()
 
+# Cap on files when allowed_sources is empty (broad fan-out across owner files).
+_MAX_AGENT_FILES = 5
 
-_SIMILARITY_THRESHOLD = 0.5
+_NO_FILES_RESOLVED_MESSAGE = (
+    "I don't have any documents available to answer that. "
+    "Please ask your instructor to add documents to my reference list."
+)
+_RAG_NOT_READY_MESSAGE = "I'm still indexing the course documents. Please try again in a few minutes."
+_DRIVE_FILE_NOT_FOUND_MESSAGE = (
+    "I couldn't access one of my reference documents. Please ask your instructor to check the bot's configured sources."
+)
+_RETRIEVAL_ERROR_MESSAGE = "Something went wrong while looking that up. Please try again in a moment."
+
+
+async def _resolve_files_for_sources(
+    session: AsyncSession,
+    user_id: int,
+    allowed_sources: list[str],
+) -> list[int]:
+    """Resolve a list of source names to RAG-ready DriveFile IDs for *user_id*.
+
+    Returns matching DriveFile IDs filtered to rag_status=READY and is_deleted=False.
+    When *allowed_sources* is empty, returns all ready owner files ordered by
+    DriveFile.id ASC, capped at _MAX_AGENT_FILES.
+    """
+    stmt = (
+        select(DriveFile)
+        .where(
+            col(DriveFile.user_id) == user_id,
+            col(DriveFile.rag_status) == RagStatus.READY,
+            col(DriveFile.is_deleted) == False,  # noqa: E712
+        )
+        .order_by(col(DriveFile.id).asc())
+    )
+    result = await session.execute(stmt)
+    files: list[DriveFile] = list(result.scalars().all())
+
+    if allowed_sources:
+        allowed_set = set(allowed_sources)
+        return [f.id for f in files if f.id is not None and resolve_source_name(f) in allowed_set]
+
+    return [f.id for f in files[:_MAX_AGENT_FILES] if f.id is not None]
+
+
+async def _run_agent_fan_out(
+    session: AsyncSession,
+    user_id: int,
+    file_ids: list[int],
+    question: str,
+    agent_llm: Any,
+) -> list[RAGContext]:
+    """Run agentic RAG per file concurrently. Any per-file failure propagates."""
+    if not file_ids:
+        return []
+
+    coros = [
+        _rag_service.get_context_via_agent(
+            session=session,
+            user_id=user_id,
+            file_db_id=file_id,
+            query=question,
+            llm=agent_llm,
+        )
+        for file_id in file_ids
+    ]
+    return await asyncio.gather(*coros)
 
 
 async def answer_question(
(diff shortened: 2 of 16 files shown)


app/core_plugins/slack/rag.py
```diff
@@ -1,114 +1,198 @@
...
+        _rag_service.get_context_via_agent(
+            session=session,
+            user_id=user_id,
+            file_db_id=file_id,
+            query=question,
+            llm=agent_llm,
+        )
+        for file_id in file_ids
+    ]
+    return await asyncio.gather(*coros)
```
abdulrafey1: 🚫 **BLOCKING: `asyncio.gather` without `return_exceptions=True` — a single file failure aborts all results**

`asyncio.gather(*coros)` raises on the first exception, cancelling all other coroutines mid-flight. So if file 2 fails with `RAGRetrievalError`, the successful results from files 1, 3, 4, 5 are silently discarded, and the student gets an error response even though most files retrieved fine.

The `except` blocks in `answer_question` catch the exception, but by then all partial results are gone.

**Fix:** use `return_exceptions=True` and handle exceptions per-result in `answer_question`:
```python
results = await asyncio.gather(*coros, return_exceptions=True)
contexts, errors = [], []
for r in results:
    if isinstance(r, BaseException):
        errors.append(r)
    else:
        contexts.append(r)
# surface the first error only if ALL files failed
if errors and not contexts:
    raise errors[0]
```

abdulrafey1 (commented):
## Code Review — feat(slack): per-file agentic retrieval

Solid direction overall — agentic retrieval is a real improvement over pure semantic search, and the error-type granularity (four distinct `ResponseType` values) is excellent. A few blocking issues that need fixing before merge, plus some important and minor ones listed below.

**Summary**
| Priority | Count |
|---|---|
| 🚫 Blocking | 2 |
| ⚠️ Important | 5 |
| 💡 Suggestion | 3 |

app/core_plugins/slack/rag.py
abdulrafey1: 🚫 **BLOCKING: `asyncio.gather` without `return_exceptions=True` — one file failure drops all results**

`asyncio.gather(*coros)` raises on the first exception and cancels all remaining coroutines. If file 2 raises `RAGRetrievalError`, the results from files 1, 3, 4, 5 are silently lost. The student gets an error response even though most files retrieved fine.

The `except` blocks in `answer_question` catch the exception, but by then every partial result is gone.

**Fix:** pass `return_exceptions=True` and handle per-result in `answer_question`:
```python
raw = await asyncio.gather(*coros, return_exceptions=True)
contexts = [r for r in raw if isinstance(r, RAGContext)]
errors   = [r for r in raw if isinstance(r, BaseException)]
if errors and not contexts:
    raise errors[0]   # surface only when all files failed
return contexts
```

abdulrafey1: 🚫 **BLOCKING: `asyncio.gather` without `return_exceptions=True` — one file failure drops all results**

`asyncio.gather(*coros)` raises on the first exception and cancels all remaining coroutines. If file 2 raises `RAGRetrievalError`, the results from files 1, 3, 4, 5 are silently lost. The student gets an error even though most files retrieved fine.

The `except` blocks in `answer_question` catch the exception, but by then every partial result is gone.

**Fix:** pass `return_exceptions=True` and handle per-result in `answer_question`:
```python
raw = await asyncio.gather(*coros, return_exceptions=True)
contexts = [r for r in raw if isinstance(r, RAGContext)]
errors   = [r for r in raw if isinstance(r, BaseException)]
if errors and not contexts:
    raise errors[0]   # only surface when ALL files failed
return contexts
```

app/core_plugins/slack/routes.py
```diff
@@ -36,7 +37,7 @@
     get_workspace_by_team,
     save_workspace,
 )
-from app.core_plugins.slack.rag import answer_question
+from app.core_plugins.slack.rag import _RETRIEVAL_ERROR_MESSAGE, answer_question
```
abdulrafey1: 🚫 **BLOCKING: importing a private symbol across modules**

```python
from app.core_plugins.slack.rag import _RETRIEVAL_ERROR_MESSAGE, answer_question
```

The underscore prefix is Python's "private to this module" signal. Importing `_RETRIEVAL_ERROR_MESSAGE` in `routes.py` (and `tests/slack/test_routes.py`) breaks that contract and creates a cross-module coupling to an implementation detail.

`NO_AI_KEY_MESSAGE` and `AI_KEY_UNAVAILABLE_MESSAGE` are correctly defined in `constants.py`. `_RETRIEVAL_ERROR_MESSAGE` (and the other three RAG error messages in `rag.py`) should live there too, without the underscore prefix.

app/core_plugins/slack/rag.py
```diff
@@ -1,114 +1,198 @@
...
 from app.rag.store import SimilarityResult
+from app.rag.types import RAGContext, RagStatus
+from app.rag.utils import resolve_source_name
 
 logger = get_logger(__name__)
 
-# Module-level singleton — embedding model and vector store loaded once per process
+# Module-level singleton — vector store and embedding provider loaded once per process.
+# Reused only for its agentic entry point (get_context_via_agent); no semantic search runs.
 _rag_service: RAGContextService = RAGContextService()
```
abdulrafey1: ⚠️ **IMPORTANT: module-level singleton loads HuggingFace model weights at import time for no reason**

```python
_rag_service: RAGContextService = RAGContextService()
```

`RAGContextService()` with no args initializes the default `HuggingFaceEmbeddingProvider`, which downloads and loads model weights when the module is first imported.

The previous version justified this singleton with "embedding model loaded once per process". But this PR removes all semantic search — `get_context_via_agent` goes through the LLM agent and never calls the embedding provider. The model load is now entirely wasted.

**Fix:** instantiate inside `_run_agent_fan_out`, or accept `rag_service` as a parameter (which also eliminates the need to patch `_rag_service` in every test).

app/core_plugins/slack/rag.py
```diff
@@ -1,114 +1,198 @@
...
+            col(DriveFile.is_deleted) == False,  # noqa: E712
+        )
+        .order_by(col(DriveFile.id).asc())
+    )
+    result = await session.execute(stmt)
+    files: list[DriveFile] = list(result.scalars().all())
+
+    if allowed_sources:
+        allowed_set = set(allowed_sources)
+        return [f.id for f in files if f.id is not None and resolve_source_name(f) in allowed_set]
```
abdulrafey1: ⚠️ **IMPORTANT: `_resolve_files_for_sources` fetches ALL ready user files and filters in Python**

```python
files: list[DriveFile] = list(result.scalars().all())   # loads ALL ready files
if allowed_sources:
    return [f.id for f in files if f.id is not None and resolve_source_name(f) in allowed_set]
```

When `allowed_sources` is non-empty the SQL query loads every `READY`, non-deleted file for the user before Python filters it. An instructor with 200 indexed documents but 3 configured as `allowed_sources` incurs an O(200) DB round-trip where only 3 rows are relevant.

The "empty sources" path is already DB-efficient (`.order_by().limit(_MAX_AGENT_FILES)`). The non-empty path should get the same treatment.

If `resolve_source_name` maps to a stored column (e.g., `DriveFile.name`), add a WHERE clause:
```python
if allowed_sources:
    stmt = stmt.where(col(DriveFile.name).in_(allowed_sources))
```
If the name transformation is more complex, at least add `.limit(len(allowed_sources) + some_buffer)` so the scan is bounded.

app/core_plugins/slack/rag.py
```diff
@@ -1,114 +1,198 @@
...
 
 logger = get_logger(__name__)
 
-# Module-level singleton — embedding model and vector store loaded once per process
+# Module-level singleton — vector store and embedding provider loaded once per process.
+# Reused only for its agentic entry point (get_context_via_agent); no semantic search runs.
 _rag_service: RAGContextService = RAGContextService()
 
+# Cap on files when allowed_sources is empty (broad fan-out across owner files).
+_MAX_AGENT_FILES = 5
```
abdulrafey1: ⚠️ **IMPORTANT: `_MAX_AGENT_FILES` violates the project constants convention**

```python
_MAX_AGENT_FILES = 5
```

Project convention (see `app/rag/constants.py` for examples): all constants belong in `constants.py`, named to reflect their module, and read from `os.getenv` with a default so they are env-configurable.

This value is hardcoded in `rag.py` and cannot be tuned without a code change.

**Fix:** move to `app/core_plugins/slack/constants.py`:
```python
SLACK_MAX_AGENT_FILES = int(os.getenv("SLACK_MAX_AGENT_FILES", "5"))
```

app/core_plugins/slack/rag.py
```diff
@@ -1,114 +1,198 @@
...
+    return [f.id for f in files[:_MAX_AGENT_FILES] if f.id is not None]
 
-_SIMILARITY_THRESHOLD = 0.5
+
+async def _run_agent_fan_out(
+    session: AsyncSession,
+    user_id: int,
+    file_ids: list[int],
+    question: str,
+    agent_llm: Any,
```
abdulrafey1: ⚠️ **IMPORTANT: `agent_llm: Any` loses type safety**

`Any` turns off mypy for everything that flows through this parameter. `RAGSearchAgent` and `get_context_via_agent` internally expect a LangChain LLM — use the same type they use:

```python
from langchain_core.language_models import BaseLanguageModel

async def _run_agent_fan_out(
    ...
    agent_llm: BaseLanguageModel,
) -> list[RAGContext]:
```

Same fix needed at the `answer_question` signature (line 107).

app/core_plugins/chat/routes.py
```diff
@@ -351,12 +351,18 @@ async def chat_completion(
 
         first_user_text = get_first_user_text(request.messages)
         if first_user_text:
+            title_provider = get_provider(
+                provider_name=provider_name,
+                api_key=api_key,
+                model=model,
+            )
```
abdulrafey1, line 360: ⚠️ **IMPORTANT: title generation now burns user quota (behavioral regression)**

```python
title_provider = get_provider(
    provider_name=provider_name,
    api_key=api_key,
    model=model,
)
```

This builds `title_provider` from the same credentials the user is actively chatting with. The previous implementation had platform-owned `CHAT_TITLE_GENERATION_*` env vars specifically to keep title generation off user quota — an instructor configured with GPT-4o will now silently use their GPT-4o key to generate a 3–6 word title on every first message.

If removing platform-owned creds is intentional, the PR description should document this trade-off. If it was an oversight, consider: skip title generation when no platform creds exist, or use a cheaper model tier.

app/core_plugins/slack/rag.py
```diff
@@ -1,114 +1,198 @@
...
+
+_NO_FILES_RESOLVED_MESSAGE = (
+    "I don't have any documents available to answer that. "
+    "Please ask your instructor to add documents to my reference list."
+)
+_RAG_NOT_READY_MESSAGE = "I'm still indexing the course documents. Please try again in a few minutes."
+_DRIVE_FILE_NOT_FOUND_MESSAGE = (
+    "I couldn't access one of my reference documents. Please ask your instructor to check the bot's configured sources."
+)
+_RETRIEVAL_ERROR_MESSAGE = "Something went wrong while looking that up. Please try again in a moment."
```
abdulrafey1: 💡 **SUGGESTION: move all four user-facing error messages to `constants.py`**

```python
_NO_FILES_RESOLVED_MESSAGE = ...
_RAG_NOT_READY_MESSAGE = ...
_DRIVE_FILE_NOT_FOUND_MESSAGE = ...
_RETRIEVAL_ERROR_MESSAGE = ...    # already leaked to routes.py
```

`NO_AI_KEY_MESSAGE` and `AI_KEY_UNAVAILABLE_MESSAGE` are correctly in `constants.py`. These four should live there too — no underscore prefix, no private convention.

This also fixes the blocking import issue in `routes.py`: once they're public constants in `constants.py`, the import is clean and consistent.

app/core_plugins/slack/rag.py
```diff
@@ -1,114 +1,198 @@
...
-_SIMILARITY_THRESHOLD = 0.5
+
+async def _run_agent_fan_out(
+    session: AsyncSession,
+    user_id: int,
+    file_ids: list[int],
+    question: str,
+    agent_llm: Any,
+) -> list[RAGContext]:
+    """Run agentic RAG per file concurrently. Any per-file failure propagates."""
```
abdulrafey1: 💡 **SUGGESTION: clarify the partial-failure design intent in the docstring**

```python
"""Run agentic RAG per file concurrently. Any per-file failure propagates."""
```

"Any per-file failure propagates" is accurate but does not explain WHY this is the chosen behaviour versus returning partial results. If the "all-or-nothing" approach is intentional, add a sentence:

```
# All-or-nothing: if any file fails, the entire fan-out raises so the caller
# can return a single clear error rather than silently incomplete context.
```

If partial results are preferred (arguably better UX), see the `return_exceptions=True` comment on the `asyncio.gather` line.

tests/slack/test_rag.py
```diff
@@ -6,257 +6,366 @@
...
-    chunk = MagicMock()
-    chunk.content = content
-    chunk.chapter = None
-    chunk.section = None
-    chunk.subsection = None
-    return chunk
+@pytest.mark.asyncio
+async def test_resolve_files_for_sources_filters_by_allowed_sources(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Returns only DriveFile IDs whose source_name is in allowed_sources."""
+    from unittest.mock import AsyncMock
```
abdulrafey1: 💡 **SUGGESTION: `AsyncMock` is already imported at module level — remove these inline re-imports**

```python
from unittest.mock import AsyncMock   # line 28 (also 51, 66, 89, 112, 131)
```

`AsyncMock` (and `MagicMock`, `patch`) is already imported on line 3:
```python
from unittest.mock import AsyncMock, MagicMock, patch
```

The six inline `from unittest.mock import AsyncMock` statements inside individual test function bodies are redundant. Delete them.

hamza-56 (approved):

abdulrafey1 (approved):
lets gooo!

app/migrations/versions/2b8cd7530d90_extend_responsetype_enum_with_per_.py
```diff
@@ -0,0 +1,32 @@
...
+
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "2b8cd7530d90"
+down_revision: Union[str, Sequence[str], None] = "6bdd5691de9d"
```
hamza-56: Heads up: this revises from `6bdd5691de9d`, but `164908f2fa97` (already on main) also revises from there. Once this lands, `alembic heads` returns two tips and the migrate-on-deploy step will refuse to run. Per [CLAUDE.md's split-heads rule](https://github.com/edly-io/sparkth/blob/main/CLAUDE.md#preventing-split-heads), please rebase onto main and regenerate the migration so `down_revision = "164908f2fa97"`.

app/core_plugins/slack/routes.py
```diff
@@ -440,32 +445,50 @@ async def _dispatch_event(
...
-            except (SQLAlchemyError, LangChainException, OSError) as exc:
-                logger.error("RAG dispatch failed for workspace %s: %s", workspace_id, exc)
-                answer = config.fallback_message
+                answer = AI_KEY_UNAVAILABLE_MESSAGE
                 rag_matched = False
-                response_type = ResponseType.fallback
-                await session.rollback()
+                response_type = ResponseType.config_incomplete
+            else:
+                agent_llm = llm_provider.create_llm()
```
hamza-56: If `create_llm()` throws (e.g. an unknown model name in `llm_model_override` raises a pydantic `ValidationError`), the exception escapes `_dispatch_event` entirely, so the Slack user gets no reply, no `BotResponseLog` row is written, and nothing is rolled back. Worth either moving this call inside the `try:` below (and adding `ValidationError` / `ValueError` to the except list), or wrapping it so it falls back to `AI_KEY_UNAVAILABLE_MESSAGE` like the `llm_provider is None` branch right above.

app/core_plugins/slack/rag.py
```diff
@@ -1,114 +1,235 @@
...
+            if s.lower().endswith(".pdf"):
+                candidate_names.add(s[:-4])
+        stmt = stmt.where(col(DriveFile.name).in_(candidate_names))
+
+    result = await session.execute(stmt)
+    files: list[DriveFile] = list(result.scalars().all())
+
+    if allowed_sources:
+        allowed_set = set(allowed_sources)
+        return [f.id for f in files if f.id is not None and resolve_source_name(f) in allowed_set]
```
hamza-56: `SLACK_MAX_AGENT_FILES` is only enforced in the `else` branch below. If an instructor configures, say, 30 allowed_sources, this returns all 30 and `_run_agent_fan_out` spins up 30 concurrent agentic LLM calls + 30 pooled DB sessions per question. Either slice this with the same cap, or constrain it at `SlackConfig.allowed_sources` so the operator gets a clear validation error instead of silent resource pressure.

app/core_plugins/slack/rag.py
```diff
@@ -1,114 +1,235 @@
...
+    if errors:
+        logger.warning(
+            "Slack agentic RAG: %d/%d files failed for user=%d: %s",
+            len(errors),
+            len(file_ids),
+            user_id,
+            errors[0],
+        )
+    if errors and not contexts:
+        raise errors[0]
```
hamza-56: Two things worth thinking about here:

1. The log a few lines up only prints `errors[0]`, so if 3 files fail with `DriveFileNotFoundError` and 2 with `RAGRetrievalError`, on-call only sees one type. Something like `Counter(type(e).__name__ for e in errors)` would be much more informative.
2. `raise errors[0]` is order-dependent because `asyncio.gather` preserves submission order, so the user-facing `ResponseType` is decided by whichever `file_id` happened to be first. If the first raised `RAGRetrievalError` but the rest all raised `DriveFileNotFoundError`, the user sees "Something went wrong" instead of the more accurate "I couldn't access one of my reference documents." A small priority order (DriveFileNotFound > RAGNotReady > RAGRetrievalError) would be friendlier when all failures agree.

app/core_plugins/chat/routes.py
```diff
@@ -351,12 +351,18 @@ async def chat_completion(
 
         first_user_text = get_first_user_text(request.messages)
         if first_user_text:
+            title_provider = get_provider(
```
hamza-56, line 354: The previous title-generation path explicitly used `temperature=0.3, max_tool_executions=0`, which is deterministic and tool-free, both of which matter for a 3-to-6-word title generator. This call accepts the provider defaults (`0.7` / `50`), so titles will become noticeably more variable for the same opening message, and the model can technically request tool calls (it won't, but the budget is there). Worth passing `temperature=0.3, max_tool_executions=0` here to preserve the prior behaviour.

PR #352, opened 2026-05-21
refactor(rag): class-based refactor, dead code removal, and structural cleanup

abdulrafey1:
# [[Sparkth] Refactor RAG codebase](https://github.com/edly-io/sparkth/issues/353)

## Summary

- **Class-based refactors** across `agent.py`, `chunking.py`, `context_service.py`, and `provider.py` — related logic grouped into classes (`RAGSearchAgent`, `DocumentChunker`, `RAGContextService`, `EmbeddingProviderRegistry`) with thin module-level backward-compat wrappers where needed
- **Dead code removal** — removed stale backward-compat aliases and wrappers from `agent.py` (`run_agentic_rag_search`, `_bind_user_context` alias), fixed stale docstrings in `provider.py`, moved `_cosine_similarity` into `RAGContextService` as a `@staticmethod`
- **Bug fix** — `slack/rag.py` was creating a second cold `HuggingFaceEmbeddingProvider` instance at module load, bypassing the warm singleton; now `RAGContextService` is constructed inside `answer_question` with `active_provider`
- **Structural splits** — `models.py` split into `db_models.py` + `enums.py` + `pydantic_models.py`; `extraction.py` promoted to `extraction/` package with per-format extractors and shared `render_markdown_table` utility; `memory_profiler.py` moved from `app/rag/` to `app/` (not RAG-specific)
- **Test co-location** — `tests/rag/` moved to `app/rag/tests/` to live alongside the code; `pyproject.toml` updated with `testpaths`; `Makefile` default path dropped in favour of `testpaths`; rag conftest made self-contained (env vars, circular-import guard, `engine`/`session` fixtures)

## Test plan

- [ ] `make test.backend` — 1096 tests pass
- [ ] `make mypy` — passes strict
- [ ] `make lint.backend` — no errors
- [ ] Slack RAG: `answer_question` uses the singleton provider for all embedding ops (no second model load)
- [ ] Running `pytest app/rag/tests/` in isolation passes without fixture or circular-import errors

## Migration notes

None — all internal refactors, no API or schema changes.


app/rag/extraction/pdf.py
```diff
@@ -0,0 +1,103 @@
...
+from app.rag.types import DocType, ExtractionResult
+
+logger = get_logger(__name__)
+
+
+class PDFExtractor(BaseExtractor):
+    """Extracts text from PDF files using PyMuPDF, with scanned-PDF detection."""
+
+    def extract(self, data: bytes, source_name: str, *, batch_size: Optional[int] = None) -> ExtractionResult:
+        settings = get_settings()
```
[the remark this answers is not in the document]
abdulrafey1: @claude The LSP concern is valid in principle, but the override is actually safe in practice: `batch_size` is keyword-only with a default, so every existing call site (`extractor.extract(data, filename)`) works without change, and mypy strict doesn't flag optional keyword-only additions as violations.

The `__init__` approach has a concrete downside here: `_REGISTRY` stores singletons instantiated at import time, but `batch_size` is intentionally read from `settings` at call time (lazy, respects runtime config). Moving it to `__init__` would bake the setting value at startup. `**kwargs` on the base is too permissive.

app/rag/chunking.py
```diff
@@ -11,122 +19,97 @@
...
-    chunks = _split(result.markdown, result.source_name)
-
-    logger.info(
-        "Chunking complete: '%s' → %d chunks",
-        result.source_name,
-        len(chunks),
-    )
-    return chunks
+    """Thin wrapper around DocumentChunker.chunk for backward compatibility."""
+    return DocumentChunker().chunk(result)
```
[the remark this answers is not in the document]
abdulrafey1: @claude `DocumentChunker` construction is cheap — no model loading, and `tiktoken` is already a module-level singleton. The wrapper is a temporary backward-compat shim and will be removed once the last caller migrates directly to `DocumentChunker`. Optimising it now would be premature.

zaira-bibi (approved):
LGTM!

app/rag/context_service.py
```diff
@@ -320,12 +309,39 @@ async def rank_sections(
...
-            sim = _cosine_similarity(query_embedding, title_emb)
+            sim = self._cosine_similarity(query_embedding, title_emb)
             scored.append((sim, all_sections[i]))
 
         scored.sort(key=lambda x: x[0], reverse=True)
         return [s[1] for s in scored[:top_n]]
 
+    @staticmethod
+    def _cosine_similarity(a: list[float], b: list[float]) -> float:
+        va, vb = np.array(a), np.array(b)
```
hamza-56: This used to be pure Python (`sum(x*y for x, y in zip(a, b))` etc.) and the prior docstring even called out `(pure Python, no numpy)`. The rewrite allocates two `np.array` objects on every call, and `rank_sections` invokes this once per section title (can be in the hundreds for a textbook-sized doc). Cleanest fix is to vectorise the whole loop in `rank_sections` into a single matmul (`np.array(title_embeddings) @ np.array(query_embedding)`), so the numpy cost is paid once per query instead of once per section.

app/models/__init__.py
```diff
@@ -1,4 +1,4 @@
-from app.rag.models import DocumentChunk  # noqa: E402
+from app.rag.db_models import DocumentChunk  # noqa: E402
```
hamza-56: Two small things in one place. First, the `# noqa: E402` is leftover from before the rename. E402 only fires on imports that come after non-import code, and this is line 1, so the suppression has no effect (was already misleading on main; the rename is a natural moment to drop it). Second, asymmetric re-export: `DocumentChunk` is re-exported here for Alembic metadata ordering, but its sibling table `DriveFileChunkLink` (same file in `app/rag/db_models.py`) is not. Worth picking one convention, either re-export both or have callers always go straight to `app.rag.db_models`.

app/rag/extraction/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+from app.rag.extraction.pdf import PDFExtractor
+from app.rag.extraction.txt import TXTExtractor
+from app.rag.types import ExtractionResult
+
+logger = get_logger(__name__)
+
+_REGISTRY: dict[str, BaseExtractor] = {
+    "pdf": PDFExtractor(),
+    "docx": DocxExtractor(),
+    "html": HTMLExtractor(),
```
hamza-56: Extractors are stateless, so this allocates 2x HTMLExtractor and 2x TXTExtractor for no real reason. The old monolithic `extraction.py` mapped both keys to the same function reference. Cleaner to bind once and reuse: `_html = HTMLExtractor(); _txt = TXTExtractor()`, then `"html": _html, "htm": _html, "txt": _txt, "md": _txt`.

hamza-56:
A few more things worth fixing that didn't fit as inline comments because the files or specific lines aren't part of this PR's diff:

**`.github/workflows/ci.yml` line 63**, `run: uv run pytest tests/`
This branch moves rag tests to `app/rag/tests/` and adds them to `testpaths` in `pyproject.toml`, but passing `tests/` as an explicit arg here overrides `testpaths`. The net effect is that all 20 test files (~308 tests) under `app/rag/tests/` are silently skipped in CI. Either drop the `tests/` arg so pytest reads `testpaths`, or list both paths: `uv run pytest tests/ app/rag/tests/`.

**`CLAUDE.md` line 37 and line 111**
The directory tree on line 37 still says `tests/  # ... (api/, chat/, mcp/, rag/)`, and the TDD section on line 111 still tells contributors to put tests under `tests/services/test_foo.py` mirroring `app/services/foo.py`. Both contradict the new co-located convention this PR introduces. Worth updating both spots so future contributors know rag-package tests live next to the source.

**`app/rag_mcp/schemas.py` line 5**
The NOTE there reads `rag.types -> rag.models -> models.__init__ -> rag.models`, but `app/rag/models.py` no longer exists (this PR split it into `db_models`, `enums`, and `pydantic_models`). `RagStatus` also moved from `rag.types` to `app.rag.enums`. Either update the chain to reference the new module paths or delete the note if the cycle no longer applies.

**`app/rag/context_service.py` line 36** (in-diff file, but this specific line is unchanged)
`self._embedding_provider = embedding_provider or HuggingFaceEmbeddingProvider()`
The Slack double-model-load fix in commit 79f2923 patched the call site, but the trap is still in this constructor. Calling `RAGContextService()` with no args silently instantiates a fresh `HuggingFaceEmbeddingProvider()`, which loads a cold model and bypasses the warmed singleton in `app.rag.provider`. Any future caller who forgets to inject the singleton hits the same regression Slack just fixed. Worth making `embedding_provider` a required arg and updating the few `RAGContextService()` calls in `app/rag/tests/test_context_service_agent.py` to pass `MagicMock()` explicitly.

PR #355, opened 2026-05-22
chore(ci): include migrations in ruff lint checks

abdulrafey1:
## Summary

- Removes the `exclude` block from `[tool.ruff]` in `pyproject.toml` that was silently skipping all migration files from linting
- Migrations contain real Python code and should be subject to the same ruff checks as the rest of the codebase
- Verified all existing migration files pass with zero violations before raising this PR

## Test plan

- [ ] `make lint.backend` passes cleanly


zaira-bibi (approved):
looks good!

PR #356, opened 2026-05-24
feat(frontend): add playwright e2e test suite

### LINKED ISSUE #357: Setup Playwright e2e tests

hamza-56:
## What
Adds a Playwright E2E test suite to the frontend (login, sign-up incl. email verification via Mailpit, authenticated dashboard) so user-facing regressions are caught before merge.

## Changes
- test(frontend): add Playwright suite under `frontend/tests/` with `auth.setup.ts` (API-login seeds storage state) and chromium project that depends on it
- test(frontend): add `login.spec.ts`, `sign-up.spec.ts`, `dashboard.spec.ts` — 12 specs covering success/error paths and the end-to-end email verification flow via Mailpit
- feat(core): add `--email-verified` flag to `create-user` CLI so the E2E seed can produce a login-able admin in one command (login enforces `email_verified`)
- test(core): add `tests/cli/test_users.py` covering the new flag (default off, explicit on sets `email_verified_at`)
- ci(frontend): add `.github/workflows/playwright.yml` with Postgres/Redis/Mailpit service containers, seed step, and blob-report artifacts
- chore(frontend): wire `make test.e2e`, `make test.e2e.ui`, `make test.e2e.install` into the Makefile; document local + CI usage in `frontend/tests/README.md`
- chore(frontend): add `@playwright/test` + `dotenv` devDependencies, `Dockerfile.playwright`, and Playwright entries in `.gitignore`

## How to Test
1. `make test.e2e.install` (one-time browser install).
2. `make up` to bring up Postgres, Redis, Mailpit, the API, and the frontend.
3. `make create-user -- --username admin --email admin@sparkth.local --password 'Sparkth-admin-1!' --superuser --email-verified` to seed the user that `auth.setup.ts` logs in as.
4. `make test.e2e` — expect 12 specs green; `make test.e2e tests/login.spec.ts` to run a single file.
5. `make test.backend tests/cli/` — expect 2 passed (covers the new `--email-verified` flag).
6. Open the PR and confirm the `Playwright Tests` workflow runs and passes.

## Notes
- New CLI flag `--email-verified` on `create-user` — existing invocations are unaffected (defaults to `false`).
- New devDependencies: `@playwright/test`, `dotenv`. Playwright browsers install on demand via `make test.e2e.install`.
- New CI workflow only runs on PR when files under `app/`, `frontend/`, `tests/`, `pyproject.toml`, `uv.lock`, `alembic.ini`, or the workflow itself change.
- No DB migrations. No breaking changes.

Closes #357


frontend/tests/utils/user.ts
```diff
@@ -0,0 +1,69 @@
...
+  password: string;
+}
+
+/**
+ * Generates a unique user payload. Username/email are randomized so tests can
+ * run in parallel without colliding on the database's UNIQUE constraints.
+ */
+export function randomUser(prefix = "e2e"): RandomUser {
+  const suffix = `${Date.now().toString(36)}${Math.random().toString(36).slice(2, 8)}`;
+  const username = `${prefix}-${suffix}`.slice(0, 20);
```
abdulrafey1: **⚠️ Important — `slice(0, 20)` collapses the unique suffix for long prefixes**

The suffix is `Date.now().toString(36)` (~8 chars) + 6 random chars = 14 chars total. With `prefix="signup-verify"` (13 chars), `"signup-verify-XXXXXXXXXXXXXX".slice(0, 20)` keeps only the first 6 chars of the suffix — and those 6 chars are the *start* of the Date.now() timestamp, so two tests running in the same millisecond collide entirely.

Worse with `prefix="login-unverified"` (16 chars in `login.spec.ts`) → only 3 chars of suffix retained.

CI is safe because `workers: 1`, but `make test.e2e` locally uses the default worker count (= CPU cores). The next person to enable parallel CI workers will see flaky DB-unique-constraint failures.

Suggested:
```typescript
export function randomUser(prefix = "e2e"): RandomUser {
  // crypto.randomUUID gives us 32 hex chars; slice 12 for length, no truncation collision risk.
  const suffix = crypto.randomUUID().replace(/-/g, "").slice(0, 12);
  const username = `${prefix.slice(0, 8)}-${suffix}`;
  ...
}
```
Or drop the `slice(0, 20)` entirely if the User model's username column allows it.

frontend/tests/utils/mailpit.ts
```diff
@@ -0,0 +1,82 @@
...
+  throw new Error(`Timed out waiting for email to ${recipient}: ${lastError}`);
+}
+
+/**
+ * Extract the first URL from a message body (text or HTML). Verification and
+ * password-reset emails embed a single link, so the first hit is the one we want.
+ */
+export function extractFirstLink(message: MailpitMessageDetail): string {
+  const haystack = message.Text || message.HTML;
+  const match = haystack.match(/https?:\/\/[^\s"<>]*[^\s"<>.,:;!?)]/);
```
abdulrafey1: **⚠️ Important — `extractFirstLink` picks the FIRST URL, not the verification link**

The regex `/https?:\/\/[^\s"<>]*[^\s"<>.,:;!?)]/` matches any URL. Today's verification template is text-only and contains one link, so this works. The moment the email template adds:
- a logo (``)
- a header brand link
- a social-media footer link

…this picks the wrong URL and the test fails confusingly (404 on `/`, or stale page).

The `message.Text || message.HTML` fallback is a partial mitigation, but if the text body is empty (some SMTP setups strip it), HTML wins and you lose.

Suggested: anchor on the actual verification path:
```typescript
const match = haystack.match(/https?:\/\/[^\s"<>]*\/auth\/verify-email[^\s"<>]*/);
```
Or search for the token query param. Either is robust to template evolution.

frontend/playwright.config.ts
```diff
@@ -0,0 +1,74 @@
...
+    //   dependencies: ["setup"],
+    // },
+  ],
+
+  // Start `next dev` locally; in CI the app is started by docker compose
+  // (see .github/workflows/playwright.yml), and PLAYWRIGHT_NO_WEBSERVER=1 is set.
+  webServer: process.env.PLAYWRIGHT_NO_WEBSERVER
+    ? undefined
+    : {
+        command: "bun run dev",
```
abdulrafey1, line 69: **⚠️ Important — local dev (`:3000` via `next dev`) vs CI (`:7727` via static export) is a meaningful divergence**

The CI workflow builds the frontend (`bun run build` → `output: "export"` static HTML) and serves it through FastAPI on `:7727`. Local runs use `next dev` on `:3000` with HMR and the dev router.

These have different routing/CORS topologies:
- **Local**: page on `:3000`, API on `:7727` → CORS preflights, different cookie scope, no proxy in `next.config.ts` (just `output: "export"`)
- **CI**: single origin `:7727`, no CORS at all

Classes of bugs that only appear in one mode:
- CORS issues only fire locally (and the test would fail to log in)
- `trailingSlash: true` static-export routing quirks only fire in CI
- Env-var inlining differences between dev/prod builds

Suggested: either standardize on the static-export topology locally too (add a `make test.e2e.serve-built` target that runs the build + serves through the API), or document the divergence prominently in `tests/README.md` so debuggers know to check both modes.

frontend/tests/sign-up.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    const user = randomUser("signup-verify");
+    await signUpViaUi(page, user);
+
+    const message = await findLatestMessageTo(request, user.email);
+    expect(message.Subject.toLowerCase()).toContain("verify");
+
+    const verificationLink = extractFirstLink(message);
+    // The link points at FRONTEND_BASE_URL — rewrite the origin to whatever
+    // the test runner is hitting so we land on the local app.
+    const path = new URL(verificationLink).pathname + new URL(verificationLink).search;
```
abdulrafey1: **💡 Suggestion — host rewrite drops the hash fragment**

```typescript
const path = new URL(verificationLink).pathname + new URL(verificationLink).search;
```
If the verification URL ever gains a `#fragment` (e.g., SPA routing, anchor link, OAuth-style flow), this drops it silently.

Suggested:
```typescript
const u = new URL(verificationLink);
await page.goto(u.pathname + u.search + u.hash);
```
Also saves the second `new URL()` parse.

frontend/Dockerfile.playwright
```diff
@@ -0,0 +1,23 @@
+# Dockerfile.playwright — Playwright E2E test runner.
```
abdulrafey1: **💡 Suggestion — unreferenced Dockerfile**

`Dockerfile.playwright` isn't referenced by any compose file, workflow step, or Makefile target — only by its own header comment, which describes a use case ("environments where installing Playwright system deps directly on the host is not desirable") that no current CI/dev workflow exercises.

Unreferenced Dockerfiles rot fast: when the Playwright minor version bumps (`v1.60.0-jammy` → `v1.65.0-noble`), nothing in CI catches the drift.

Suggested: either wire it up (`make test.e2e.docker` target invoking `docker build -f Dockerfile.playwright`) or remove it. If kept, add a comment pointing to where its assumed entry point is.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,171 @@
...
+        working-directory: frontend
+
+      - name: Build frontend (static export)
+        run: bun run build
+        working-directory: frontend
+
+      - name: Run database migrations
+        run: uv run alembic upgrade head
+
+      - name: Seed E2E superuser
```
abdulrafey1, line 128: **💡 Suggestion — `Seed E2E superuser` step isn't idempotent**

`users create-user` exits with code 1 if the user already exists:
```python
if existing:
    typer.secho("User with username '...' or email '...' already exists!", ...)
    raise typer.Exit(code=1)
```

Fine for ephemeral GitHub Actions runners. Breaks under:
- `workflow_dispatch` re-runs against a persistent test environment
- Local `act` runs against a host Docker daemon that retains the Postgres volume between attempts

Suggested options (any one):
- Wrap the seed step in `set +e` and `|| true` after explicitly inspecting the user table
- Add a `--if-not-exists` flag to the CLI (probably the right durable fix)
- Drop the database before seeding (heavy-handed)

frontend/playwright.config.ts
```diff
@@ -0,0 +1,74 @@
...
+const BASE_URL = process.env.PLAYWRIGHT_BASE_URL ?? "http://localhost:3000";
+
+export default defineConfig({
+  testDir: "./tests",
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
```
abdulrafey1: **💡 Suggestion — `retries: 2` in CI masks flaky tests**

Auto-retrying tests on CI silently rescues flaky failures. The blob reporter records retry counts so they're visible *if* someone looks, but they're easy to miss in the green PR check.

For a brand-new suite, you want flakiness to be loud — fix the test, don't paper over it.

Suggested:
- Start with `retries: 0` on CI; bump only if/when an unfixable flake forces it.
- Or keep `retries: 1` and fail the build if any test required a retry (the blob report exposes this — could gate via a small post-step).

abdulrafey1 (commented):
## Code Review — PR #356

The Playwright suite is a meaningful addition — `auth.setup.ts` with API-login + persisted storage state is the right pattern, the Mailpit integration to assert the end-to-end verification round-trip is great coverage, and the new `--email-verified` CLI flag is a clean way to keep CI reproducible. The CI workflow has good hygiene (path filters, blob reports, log artifacts on failure).

No blocking issues. A few **important** items worth tightening before merge, and some **suggestions**.

### Summary of findings

**⚠️ Important**
1. `randomUser()` username truncation collapses the unique suffix for long prefixes — `signup-verify` (13 chars) keeps only 6 chars of suffix, `login-unverified` (16 chars) keeps only 3. With `Math.random()` over those few chars, parallel runs can collide on the DB `UNIQUE` constraint. Safe in CI (`workers: 1`) but fragile locally.
2. `extractFirstLink` picks the *first* URL in the email body via a generic `https?://` regex. Works for the current text-only verification template, but the moment HTML branding is added (logo ``, header link), it picks the wrong URL silently.
3. Local dev runs Playwright against `next dev` on `:3000` while CI runs against the production static export served by FastAPI on `:7727`. That's two different routing/CORS topologies — bugs that only manifest in one mode will sneak through.

**💡 Suggestion**
4. Verification URL host rewrite drops the hash fragment: `new URL(...).pathname + new URL(...).search` — if the link ever gains a `#fragment`, the test will navigate to the wrong place.
5. `frontend/Dockerfile.playwright` is added but never referenced by any compose file, workflow, or doc command. Dead-code risk — it'll rot.
6. `Seed E2E superuser` step in the workflow isn't idempotent — the CLI exits with code 1 when the user already exists. Fine for ephemeral CI VMs, breaks `workflow_dispatch` re-runs against persistent state.
7. Specs lean on user-facing copy (`/manage your enabled plugins/i`, `/incorrect username or password/i`, `/check your email/i`) — these will break on marketing-copy changes. `data-testid` or stable headings would be more durable.
8. `retries: 2` on CI hides flaky tests instead of surfacing them. The blob reporter records retry counts but it's easy to ignore.
9. PR description has a typo at the bottom: `Closes #357 357` (extra `357`).

Inline detail on each below 👇

.github/workflows/playwright.yml
hamza-56: Acknowledged, but I'd like to push back on this for now: the workflow's Postgres is a GitHub Actions service container, which is recreated fresh per job run — there's no state to retain between attempts. Step-level re-runs in GHA recreate services with the job. The two cases you flagged (`workflow_dispatch` against persistent state, local `act` runs) aren't how this repo is operated today.

Happy to revisit with a real `--if-not-exists` flag on the CLI if/when we point CI at a persistent test environment, since that's the durable fix and avoids the `|| true` swallow-failure shape.

hamza-56:
Thanks for the review @abdulrafey1:

**Addressed**
- ⚠️ `randomUser()` collision: switched to `crypto.randomUUID()`-derived suffix with prefix capped to 8 chars — random bits now survive the 20-char truncation.
- ⚠️ `extractFirstLink`: renamed to `extractVerificationLink` and anchored on `/verify-email` path so future template logos/footer links can't be picked up.
- 💡 Hash fragment: single URL parse with `pathname + search + hash`.
- 💡 Unreferenced `Dockerfile.playwright`: removed.
- 💡 `retries: 2` → `retries: 0` (CI and local).
- Local-vs-CI topology divergence: added a section to `frontend/tests/README.md` so debuggers know to check both.
- PR description typo (`Closes #357 357`): fixed.

**Pushing back**
- 💡 `Seed E2E superuser` idempotency — replied inline. Postgres is a GHA service container, recreated per job run; the cases you flagged aren't how the repo runs CI today.
- Copy-based selectors (`/incorrect username or password/i`, `/check your email/i`, etc.) — leaving as-is. The copy isn't marketing — it's product UX strings the team owns. Adding `data-testid` everywhere is a wider conventions change worth a separate discussion, and Playwright's accessible-name selectors are robust to most copy tweaks. Happy to revisit if we see real breakage.

frontend/tests/utils/user.ts
hamza-56: Fixed in `62da0aa`: switched to `crypto.randomUUID()` (11 hex chars retained, prefix capped to 8). Random bits now survive the 20-char username truncation.

frontend/tests/utils/mailpit.ts
hamza-56: Fixed in `62da0aa`: renamed to `extractVerificationLink` and anchored the regex on `/verify-email` so future template logos or footer links can't be picked up.

frontend/playwright.config.ts
hamza-56: Documented in `62da0aa` (new 'Local vs CI topology' section in `frontend/tests/README.md`). Standardizing locally on the static-export topology is worth a separate scope, so leaving that for a follow-up.

frontend/tests/sign-up.spec.ts
hamza-56: Fixed in `62da0aa`: single URL parse, `pathname + search + hash`.

frontend/Dockerfile.playwright
hamza-56: Removed in `62da0aa`. Agreed it's dead weight; we can re-add a real Dockerfile if/when there's a wired-up use case.

frontend/playwright.config.ts
hamza-56: Fixed in `62da0aa`: dropped to `retries: 0` across the board so flakes stay loud.

abdulrafey1 (approved):
All comments resolved, looks good

hamza-56:
@claude-review

PR #359, opened 2026-05-25
refactor(chat): cleanup and organize chat plugin codebase with env vars documentation

### LINKED ISSUE #358: Refactor chat plugin
### Background

The `chat` core plugin has grown organically and now has several maintainability issues that need to be addressed before the codebase becomes harder to work with.

### Problems

1. **Magic numbers** scattered throughout the code with no env-var override mechanism.
2. **Monolithic `routes.py`** (~1376 lines) is too large to navigate or maintain effectively.
3. **Dead code**: `middleware.py` contains `RateLimitMiddleware` that was never imported; `ChatPlugin.get_middleware()` returns `[]`.
4. **Duplicate route registration**: `chat_router` is registered both in `main.py` and via the plugin lifespan, causing every chat endpoint to appear twice in the OpenAPI schema at `/docs`.
5. **Unused schema classes** in `schemas.py` (`ToolResult`, `ErrorResponse`, `ConversationCreate`, and duplicate `ProviderInfo`/`ProviderCatalogResponse`) whose canonical versions already exist in `app/llm/schemas.py`.
6. **Undocumented env vars**: No `CHAT_*` variables are listed in the root `.env.example` or `CLAUDE.md`.
7. **Inconsistent logging**: `%s`-style lazy logging used inconsistently throughout.
8. **No test isolation for the chat router**: Unlike `slack` and `googledrive`, there is no `conftest.py` to register routes at test load time.

### Acceptance Criteria

- [ ] Extract magic numbers into `constants.py` with env-var overrides:
  - `CHAT_TITLE_PROMPT_MAX_CHARS`
  - `CHAT_TITLE_LLM_MAX_TOKENS`
  - `CHAT_TITLE_DB_MAX_LENGTH`
  - `CHAT_TITLE_LLM_TEMPERATURE`
  - `CHAT_DEFAULT_SIMILARITY_THRESHOLD`
- [ ] Split `routes.py` into a `routes/` package — `completions.py`, `conversations.py`, `attachments.py`, `tool_endpoints.py`, `dependencies.py`, `__init__.py` — with backward-compat re-exports via `__all__`.
- [ ] Introduce `ConversationTitleGenerator` and `LMSCredentialsBuilder` class-based wrappers; add `ToolRegistry.reset()` for test isolation.
- [ ] Remove duplicate `chat_router` registration from `main.py`; plugin lifespan should be the single registration path, consistent with `googledrive` and `slack`.
- [ ] Delete dead `middleware.py` and remove orphaned `rate_limit_*` fields from `ChatSystemConfig`.
- [ ] Remove unused schema classes from `schemas.py`.
- [ ] Delete `app/core_plugins/chat/.env.example` (never wired up); document all `CHAT_*` env vars in root `.env.example` and `CLAUDE.md`.
- [ ] Add `tests/chat/conftest.py` with `_ensure_chat_routes()` for test isolation (same pattern as `slack` and `googledrive`).
- [ ] All existing tests pass (`make test.backend`) and linting is clean (`make lint.backend`).
- [ ] No duplicate routes visible in the OpenAPI schema at `/docs`.

_Issue description written with the help of an LLM (Claude)._

zaira-bibi:
## What
Refactor the `chat` core plugin to improve code organization, reduce magic numbers, remove dead code, and document all environment-configurable variables.

Closes #358.

## Changes
- `refactor(chat)`: extract magic numbers into `constants.py` with env-var overrides (`CHAT_TITLE_PROMPT_MAX_CHARS`, `CHAT_TITLE_LLM_MAX_TOKENS`, `CHAT_TITLE_DB_MAX_LENGTH`, `CHAT_TITLE_LLM_TEMPERATURE`, `CHAT_DEFAULT_SIMILARITY_THRESHOLD`)
- `refactor(chat)`: introduce `ConversationTitleGenerator` and `LMSCredentialsBuilder` class-based wrappers; add `ToolRegistry.reset()` for test isolation; fix `%s`-style lazy logging throughout
- `refactor(chat)`: split monolithic `routes.py` (~1376 lines) into a `routes/` package — `completions.py`, `conversations.py`, `attachments.py`, `tool_endpoints.py`, `dependencies.py`, `__init__.py` — with backward-compat re-exports via `__all__`
- `fix(chat)`: remove duplicate `chat_router` registration from `main.py`; plugin lifespan is now the single registration path, consistent with `googledrive` and `slack`
- `chore(chat)`: delete dead `middleware.py` (`RateLimitMiddleware` was never imported; `ChatPlugin.get_middleware()` returns `[]`)
- `chore(chat)`: remove unused schema classes from `schemas.py` — `ToolResult`, `ErrorResponse`, `ConversationCreate`, and the duplicate `ProviderInfo`/`ProviderCatalogResponse` (canonical versions live in `app/llm/schemas.py`)
- `chore(chat)`: delete `app/core_plugins/chat/.env.example` (variables were never wired up; nothing loaded the file); remove orphaned `rate_limit_*` fields from `ChatSystemConfig` (left over from deleted middleware)
- `chore(chat)`: document all `CHAT_*` env vars in root `.env.example` and `CLAUDE.md`
- `test(chat)`: add `tests/chat/conftest.py` with `_ensure_chat_routes()` so tests register the router at module load time without needing the ASGI plugin lifespan (same pattern as `tests/slack/conftest.py` and `tests/googledrive/conftest.py`)

## How to Test
1. `make test.backend` — all tests should pass
2. `make lint.backend` — ruff and mypy clean
3. Start the server (`make api`) and confirm `/api/v1/chat/completions`, `/api/v1/chat/conversations`, and `/api/v1/tools` all respond correctly
4. Confirm no duplicate routes in OpenAPI schema at `/docs` (previously each chat endpoint appeared twice)

## Notes
All backward-compat module-level aliases are preserved (`generate_conversation_title`, `build_lms_credentials_message`, etc.) — no call-site changes required outside the chat module.


app/core_plugins/chat/conversation_title.py
```diff
@@ -2,68 +2,81 @@
...
-            text = " ".join(
-                block.get("text", "")
-                for block in msg.content
-                if isinstance(block, dict) and block.get("type") == "text"
-            ).strip()
-        if text:
-            return text
-    return None
-
+class ConversationTitleGenerator:
```
hamza-56: Same anti-pattern as the slack refactor PR. Every method here is `@staticmethod`, no state, no constructor, no DI. It's three module-level functions wearing a class hat, and then re-exported as lowercase aliases at the bottom for the tests. Suggest reverting to plain module functions, which is what the rest of the codebase does for utilities like this.

app/core_plugins/chat/conversation_title.py
```diff
@@ -2,68 +2,81 @@
...
-                    session=session,
-                    conversation_id=conversation_id,
-                    user_id=user_id,
-                    title=title[:255],
-                )
-            logger.info(f"Generated title for conversation {conversation_id}: {title!r}")
-    except (KeyError, ValueError, RuntimeError, OSError) as e:
-        logger.warning(f"Title generation failed for conversation {conversation_id}: {e}")
+# Backward-compatible module-level aliases
+get_first_user_text = ConversationTitleGenerator.get_first_user_text
```
hamza-56: These aliases only exist because `tests/chat/test_conversation_title.py` still imports the lowercase names. Production code (completions.py) already moved to the class API, so we have two parallel APIs in one file. Either migrate the tests to the class names and delete these, or drop the class and keep module functions. Don't keep both.

app/core_plugins/chat/lms_credentials.py
```diff
@@ -18,78 +18,85 @@
...
-
-    if not credential_sections:
-        return _LMS_RULES
-
-    return (
-        _LMS_RULES
-        + "\n5. Use the credentials below automatically when calling LMS tools without asking the user to provide them again:\n\n"
-        + "\n\n".join(credential_sections)
-    )
+class LMSCredentialsBuilder:
```
hamza-56: Same shape as `ConversationTitleGenerator`: stateless class, all `@staticmethod`. The lazy `from app.plugins import PLUGIN_CONFIG_CLASSES` is the only real complexity, and a plain `async def` handles that just as well. Suggest reverting to module-level functions.

app/core_plugins/chat/lms_credentials.py
```diff
@@ -18,78 +18,85 @@
...
+
+        return (
+            _LMS_RULES
+            + "\n5. Use the credentials below automatically when calling LMS tools without asking the user to provide them again:\n\n"
+            + "\n\n".join(credential_sections)
+        )
+
+
+# Backward-compatible module-level aliases
+_lms_tool_prefixes = LMSCredentialsBuilder.lms_tool_prefixes
```
hamza-56: Pylance flags `_lms_tool_prefixes` and `_has_lms_tools` here as unused. That's a static analyzer confirming the aliases are dead weight. Production uses `LMSCredentialsBuilder.build_message(...)` directly. If only tests need the lowercase names, that's a sign the class isn't earning its keep.

app/core_plugins/chat/routes/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+    router as _conversations_router,
+)
+from app.core_plugins.chat.routes.dependencies import (
+    get_chat_service,
+    get_chat_system_config,
+    get_rag_context_service,
+)
+from app.core_plugins.chat.routes.tool_endpoints import router as _tool_endpoints_router
+
+__all__ = [
```
hamza-56: Underscore-prefixed names in `__all__` is contradictory: the underscore says 'private', `__all__` says 'public'. They're here because some tests import them at the package level. The codebase is also inconsistent: `tests/chat/test_intent_router_integration.py:106` patches the submodule path (`routes.completions._resolve_drive_file_blocks`), while other tests use the package-level import. Pick one direction. Either drop the underscore on names that need to be in `__all__`, or remove them from `__all__` and have all tests import from the submodule.

app/core_plugins/chat/routes/attachments.py
```diff
@@ -0,0 +1,135 @@
...
+from app.core_plugins.chat.schemas import (
+    AttachedDriveFileResponse,
+    ConversationAttachmentCreate,
+    ConversationAttachmentResponse,
+)
+from app.core_plugins.chat.service import ChatService
+from app.models.drive import DriveFile as DriveFileModel
+from app.models.user import User
+
+router = APIRouter()
```
hamza-56, line 18: All three handlers below repeat the same pattern: get the conversation by UUID, raise 404 if missing, then (for POST/DELETE) hand-roll a DriveFile ownership query. Both are good candidates for FastAPI sub-dependencies: a `get_owned_conversation` that returns the conversation or 404s, and an `assert_owns_drive_file` lifted into the service layer. Each handler then becomes a few lines of actual logic. The route file also stops needing to know about `DriveFileModel` internals, which belong behind the service boundary.

app/core_plugins/chat/routes/dependencies.py
```diff
@@ -0,0 +1,22 @@
+from functools import lru_cache
+
+from app.core_plugins.chat.config import ChatSystemConfig
+from app.core_plugins.chat.service import ChatService
+from app.rag.context_service import RAGContextService
+from app.rag.provider import get_provider as get_rag_provider
+
+
+@lru_cache
```
hamza-56: `@lru_cache` on a zero-arg function turns it into a singleton, which is what you want here, but it also hides config caching from test authors. If a test sets env vars after the first call they have to remember `get_chat_system_config.cache_clear()`. Same pattern as `app.core.config:get_settings` so it's at least consistent, but a one-line comment pointing at that precedent would save someone debugging an env-override test later.

app/core_plugins/chat/routes/tool_endpoints.py
```diff
@@ -0,0 +1,42 @@
...
+
+from app.api.v1.auth import get_current_user
+from app.core_plugins.chat.schemas import ToolListResponse, ToolSchema
+from app.core_plugins.chat.tools import get_tool_registry
+from app.models.user import User
+
+router = APIRouter()
+
+
+def get_parameters_schema(args_schema: type[BaseModel] | dict[str, Any] | None) -> dict[str, Any]:
```
hamza-56: This helper is a generic 'shape an args_schema into a dict' utility. It doesn't know anything about HTTP. It belongs next to `ToolRegistry` in `chat/tools.py`, or as a small method on a tool-schema type. Any future endpoint that wants tool schemas will be tempted to either re-import this from the routes layer (wrong direction) or duplicate it.

app/core_plugins/chat/routes/completions.py
```diff
@@ -1,62 +1,40 @@
...
-    ConversationDetailResponse,
-    ConversationListResponse,
-    ConversationResponse,
-    MessageResponse,
-    ToolListResponse,
-    ToolSchema,
-)
+from app.core_plugins.chat.lms_credentials import LMSCredentialsBuilder
+from app.core_plugins.chat.models import Conversation
+from app.core_plugins.chat.routes.dependencies import get_chat_service, get_chat_system_config, get_rag_context_service
```
hamza-56: Heads up on the bigger picture for this file: `chat_completion` (the endpoint below) is around 410 lines and `stream_chat_response` is around 350 lines with 14 parameters. The commit message says 'split routes.py into routes/ package', but the split only extracted the easy endpoints (conversations, attachments, tools) and left the two heavy hitters in this file as-is. The real win is extracting helpers/services so `chat_completion` becomes a short orchestrator (LLM-config resolution, scope preflight, conversation get-or-create, RAG intent decision, response branch). Worth tackling in this PR while the structural work is fresh.

app/core_plugins/chat/routes/completions.py
```diff
@@ -1,62 +1,40 @@
...
 from app.api.v1.auth import get_current_user
 from app.core.db import async_engine, get_async_session
 from app.core.logger import get_logger
 from app.core_plugins.chat.config import ChatSystemConfig
-from app.core_plugins.chat.conversation_title import (
-    extract_title_from_messages,
-    generate_conversation_title,
-    get_first_user_text,
-)
+from app.core_plugins.chat.conversation_title import ConversationTitleGenerator
```
hamza-56: On the DRY side: `_resolve_drive_file_blocks` (further down) and the in-stream RAG resolution loop inside `stream_chat_response` do the same work. Iterate `drive_file` blocks, call `rag_service.get_context_via_agent`, classify the three error types (`DriveFileNotFoundError`, `RAGNotReadyError`, `RAGRetrievalError`). They diverge only in how they signal errors (HTTPException vs SSE event). Worth extracting the shared core so a bug fix in one place doesn't silently miss the other.

app/core_plugins/chat/conversation_title.py
abdulrafey1: **⚠️ Important — class-of-staticmethods is an anti-pattern**

Every method on `ConversationTitleGenerator` is `@staticmethod`. No state, no inheritance, no override points. In Python, a module is already a namespace — replacing `module_function()` with `ClassName.static_method()` adds visual noise (longer call sites, indirection at every call) without adding any structure.

The smell is most visible at the bottom of the file:
```python
get_first_user_text = ConversationTitleGenerator.get_first_user_text
extract_title_from_messages = ConversationTitleGenerator.extract_title_from_messages
generate_conversation_title = ConversationTitleGenerator.generate
```
The "backward-compat aliases" admit the class is artificial — we're re-exporting the methods under their pre-refactor module-level names.

Suggested: drop the class wrapper, keep these as module-level functions (they already were). If config-injection is the goal, pass `ChatSystemConfig` as a parameter — same effect, simpler call sites.

Same critique applies to `LMSCredentialsBuilder` in `lms_credentials.py`.

app/core_plugins/chat/lms_credentials.py
abdulrafey1: **⚠️ Important — same class-of-staticmethods concern as in `conversation_title.py`**

All three methods are `@staticmethod`; the class adds no state and no override points. The aliases at the bottom (`_lms_tool_prefixes = LMSCredentialsBuilder.lms_tool_prefixes`, etc.) admit the wrapper is artificial.

The self-call on line 37 (`LMSCredentialsBuilder.lms_tool_prefixes()`) is the giveaway — when static methods on the same class have to call each other by class name, it's because Python doesn't have a `cls.` shortcut for static methods. With module-level functions, this just becomes `lms_tool_prefixes()`.

Suggested: revert to module-level functions. Existing modules were already proper namespaces.

app/core_plugins/chat/routes/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+    "_extract_query_text",
+    "_resolve_drive_file_blocks",
+    "_stream_out_of_scope_refusal",
+    "_streaming_error_message",
+    "stream_chat_response",
+    "_parse_metadata_list",
+    "get_chat_service",
+    "get_chat_system_config",
+    "get_rag_context_service",
+]
```
abdulrafey1: **⚠️ Important — `__all__` re-exports private (`_`-prefixed) symbols**

`_extract_query_text`, `_resolve_drive_file_blocks`, `_stream_out_of_scope_refusal`, `_streaming_error_message`, `_parse_metadata_list` are all private to their submodules per the underscore convention. Re-exporting them as public package symbols breaks that contract.

These re-exports exist to preserve untouched test imports:
```python
# tests/chat/test_rag_status_events.py
from app.core_plugins.chat.routes import _parse_metadata_list
# tests/llm/test_provider_model.py
from app.core_plugins.chat.routes import _streaming_error_message
```

Proper fix: update those test imports to their real modules (`from app.core_plugins.chat.routes.completions import _parse_metadata_list`, etc.) and drop the underscore entries from `__all__`. Those tests aren't in this PR's diff, but adding 5-6 import path updates is a small cost for a clean public API.

app/core_plugins/chat/conversation_title.py
```diff
@@ -2,68 +2,81 @@
...
+                ).strip()
+            if text:
+                return text
         return None
-    if len(text) <= max_length:
-        return text
-    return text[:max_length].rsplit(" ", 1)[0] + "..."
+
+    @staticmethod
+    def extract_title_from_messages(messages: list[ChatMessage], max_length: int = 60) -> str | None:
```
abdulrafey1: **⚠️ Important — `max_length: int = 60` default drifts from `ChatSystemConfig.title_max_length`**

This PR introduces `ChatSystemConfig.title_max_length: int = 60` (env-configurable via `CHAT_TITLE_MAX_LENGTH`). But this method still has its own hardcoded `max_length: int = 60` default.

If an operator sets `CHAT_TITLE_MAX_LENGTH=80`, any caller that omits the arg still gets 60. Safe today — the only caller (`completions.py:298`) passes `max_length=config.title_max_length` explicitly — but it's a latent footgun if a future caller is added.

Suggested: drop the default. Force explicit passing of the config-derived value.

tests/chat/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+    global _chat_routes_registered
+    if _chat_routes_registered:
+        return
+    existing = {getattr(r, "path", None) for r in app.routes}
+    if f"{_CHAT_PREFIX}/chat/completions" not in existing:
+        app.include_router(chat_router, prefix=_CHAT_PREFIX)
+    _chat_routes_registered = True
+
+
+_ensure_chat_routes()
```
abdulrafey1: **⚠️ Important — module-level side effect + redundant boolean**

Two concerns:

1. `_ensure_chat_routes()` is called at module import time and mutates `app.routes` permanently. This works because pytest loads conftest files in dir scope, but the pattern is fragile and undocumented — a future contributor sees this code with no explanation.

2. The `_chat_routes_registered` boolean is redundant. The path-membership check on line 13 (`if f"{_CHAT_PREFIX}/chat/completions" not in existing`) already prevents double-registration. Two guards for the same invariant is one too many.

Suggested:
```python
"""Register chat routes for tests.

The plugin lifespan (which normally registers chat routes via
`plugin_manager.load_all_enabled`) does not run in tests, so we mount
the router directly. Same pattern as `tests/slack/conftest.py` and
`tests/googledrive/conftest.py`.
"""
from app.core_plugins.chat.routes import chat_router
from app.main import app

_CHAT_PREFIX = "/api/v1"

if f"{_CHAT_PREFIX}/chat/completions" not in {getattr(r, "path", None) for r in app.routes}:
    app.include_router(chat_router, prefix=_CHAT_PREFIX)
```

app/core_plugins/chat/lms_credentials.py
```diff
@@ -18,78 +18,85 @@
...
+        Lazy-imported to avoid circular dependencies at module load time.
+        """
+        from app.plugins import PLUGIN_CONFIG_CLASSES  # lazy import — avoids circular dep
+
+        return tuple(prefix for cls in PLUGIN_CONFIG_CLASSES.values() if (prefix := cls.lms_tool_prefix()) is not None)
+
+    @staticmethod
+    def has_lms_tools(tools: list[Any]) -> bool:
+        """Return True if any tool name starts with an LMS-specific prefix."""
+        prefixes = LMSCredentialsBuilder.lms_tool_prefixes()
```
abdulrafey1: **💡 Suggestion — self-call via class name**

`prefixes = LMSCredentialsBuilder.lms_tool_prefixes()` — calling a sibling static method by typing out the class name is the canonical smell that a class-of-staticmethods should be a module. With module-level functions, this would just be `prefixes = lms_tool_prefixes()`.

See the main thread on the class definition (line 21) for the broader fix.

app/core_plugins/chat/conversation_title.py
```diff
@@ -2,68 +2,81 @@
...
+        """Derive a provisional conversation title from the first user message."""
+        text = ConversationTitleGenerator.get_first_user_text(messages)
```
abdulrafey1: **💡 Suggestion — self-call via class name**

`text = ConversationTitleGenerator.get_first_user_text(messages)` — same anti-pattern as in `lms_credentials.py:37`. A static method calling its sibling static method by class name. With module-level functions this is just `text = get_first_user_text(messages)`.

app/core_plugins/chat/conversation_title.py
```diff
@@ -2,68 +2,81 @@
...
+    async def generate(
+        conversation_id: int,
+        user_id: int,
+        first_user_message: str,
+        service: ChatService,
+        provider: BaseChatProvider,
+        config: ChatSystemConfig | None = None,
+    ) -> None:
+        """Background task: ask the LLM for a short title and persist it."""
+        cfg = config or ChatSystemConfig()
```
abdulrafey1: **💡 Suggestion — `config: ChatSystemConfig | None = None` fallback is dead code**

```python
config: ChatSystemConfig | None = None,
) -> None:
    cfg = config or ChatSystemConfig()
```

The sole caller (`completions.py:319`) always passes `config=config` explicitly. The `None` branch creates a *new* `ChatSystemConfig()` instance, bypassing the `@lru_cache` on `get_chat_system_config()` — so if you ever hit this branch you also silently lose env-var overrides set after the first cache miss elsewhere.

Suggested: make `config` required (drop the default), or document why a fallback exists if there's a non-obvious reason.

abdulrafey1 (commented):
## Code Review — PR #359

Nice refactor overall. The env-var extraction is well-documented, the dead-code removal (rate-limit middleware, unused schemas, stale `.env.example`) is welcome, and the route split into a `routes/` package is a meaningful improvement over the previous 1376-line monolith. Logging fixes to `%s`-style lazy formatting in `tools.py` are good.

No blocking issues. Findings categorized as **important** (worth addressing before merge) and **suggestion** (nice-to-have). Inline comments follow.

### Summary of findings

**⚠️ Important**
1. `ConversationTitleGenerator` and `LMSCredentialsBuilder` are classes-of-staticmethods — Python modules are already namespaces; the class wrappers add visual noise without adding state, inheritance, or override points. The trailing module-level aliases (`get_first_user_text = ConversationTitleGenerator.get_first_user_text`) admit the wrapper is artificial.
2. `routes/__init__.py` re-exports private (`_`-prefixed) symbols via `__all__`. Tests outside this PR rely on these imports, but the underscore prefix means "private" — the proper fix is to update those test imports to point at the actual submodule.
3. `extract_title_from_messages` has a hardcoded `max_length: int = 60` default that drifts from the new env-configurable `ChatSystemConfig.title_max_length`. Safe today (sole caller passes explicitly) but a latent footgun.
4. `tests/chat/conftest.py` runs `_ensure_chat_routes()` at module load time and mutates the global `app.routes`. The pattern works but is undocumented; the `_chat_routes_registered` boolean is redundant with the existing path-membership check.

**💡 Suggestion**
5. `LMSCredentialsBuilder.has_lms_tools` calls `LMSCredentialsBuilder.lms_tool_prefixes()` — calling sibling static methods via the class name is the smell that gives the class-as-namespace away.
6. Same in `ConversationTitleGenerator.extract_title_from_messages` → calls `ConversationTitleGenerator.get_first_user_text(...)`.
7. `ConversationTitleGenerator.generate` accepts `config: ChatSystemConfig | None = None` with a `cfg = config or ChatSystemConfig()` fallback — the PR's sole caller always passes `config` explicitly, so the fallback is dead code that also bypasses the `@lru_cache` on `get_chat_system_config`.
8. `routes/__init__.py` uses two separate `from app.core_plugins.chat.routes.completions import (...)` blocks — combine them.
9. Removing the explicit `app.include_router(chat_router)` from `app/main.py` means chat now depends entirely on `plugin_lifespan`, which broadly catches `(ImportError, RuntimeError, OSError)`. If plugin loading silently fails in production, chat will 404 with no explicit error path. Consistent with `googledrive`/`slack` so it's a pre-existing pattern, but worth flagging.
10. `tests/chat/conftest.py` would benefit from a short docstring explaining *why* it exists (plugin lifespan doesn't run in tests).

Detailed comments inline 👇

app/core_plugins/chat/routes/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+from app.core_plugins.chat.routes.completions import (
+    _extract_query_text,
+    _resolve_drive_file_blocks,
+    _stream_out_of_scope_refusal,
+    _streaming_error_message,
+    stream_chat_response,
+)
+from app.core_plugins.chat.routes.completions import (
+    router as _completions_router,
+)
```
abdulrafey1, line 6: **💡 Suggestion — combine the two `from app.core_plugins.chat.routes.completions import ...` blocks**

Lines 4-10 and 11-13 both import from `completions`. Combine:
```python
from app.core_plugins.chat.routes.completions import (
    _extract_query_text,
    _resolve_drive_file_blocks,
    _stream_out_of_scope_refusal,
    _streaming_error_message,
    router as _completions_router,
    stream_chat_response,
)
```
Same for `conversations` (lines 14-19).

app/main.py
```diff
@@ -136,4 +135,3 @@ async def lifespan(application: FastAPI) -> AsyncIterator[None]:
 )
 
 app.include_router(api_router, prefix="/api/v1")
```
abdulrafey1, line 116: **💡 Suggestion — explicit `chat_router` registration removed**

Removing `app.include_router(chat_router, prefix="/api/v1")` means chat is now entirely dependent on `plugin_lifespan`, which broadly catches `(ImportError, RuntimeError, OSError)` and only logs failures. If a plugin-load error happens in production (e.g., DB hiccup at startup), `/api/v1/chat/*` will silently 404 with no obvious cause.

This is consistent with `googledrive`/`slack` (also lifespan-only), so it's not a regression *introduced by this PR* — but if you want defense-in-depth, consider either:
- Re-adding the explicit `app.include_router` as a belt-and-suspenders backup, or
- Re-raising in `plugin_lifespan` for `(ImportError, RuntimeError, OSError)` instead of just logging.

Low priority — flag rather than block.

zaira-bibi: Deliberately removed as the chat endpoints were being registered twice.

app/core_plugins/chat/routes/dependencies.py
```diff
@@ -0,0 +1,46 @@
+from functools import lru_cache
```
regisb: "dependencies.py" is a weird name for a module. Do we mean that this is the module that will serve as its own internal API? If yes, then "api.py" is a better name.

app/core_plugins/chat/routes/dependencies.py
```diff
@@ -0,0 +1,46 @@
...
+from app.api.v1.auth import get_current_user
+from app.core.db import get_async_session
+from app.core_plugins.chat.config import ChatSystemConfig
+from app.core_plugins.chat.models import Conversation
+from app.core_plugins.chat.service import ChatService
+from app.models.user import User
+from app.rag.context_service import RAGContextService
+
+
+@lru_cache  # singleton by design — mirrors app.core.config.get_settings(); call .cache_clear() in tests that mutate env vars
```
regisb: Let's get rid of this comment which refers to unit tests and will become obsolete as soon as we change app.core.config.get_settings().

app/core_plugins/chat/routes/dependencies.py
```diff
@@ -0,0 +1,46 @@
...
+    """Dependency to get chat system configuration from environment variables."""
+    return ChatSystemConfig()
+
+
+def get_chat_service() -> ChatService:
+    """Dependency to get chat service."""
+    return ChatService()
+
+
+def get_rag_context_service() -> RAGContextService:
```
regisb: There's already a `_get_rag_service()` function in the slack plugin and a `get_rag_context_service()` in the chat plugin `routes.py`. This tells us that this function should really be part of a shared library (see #379).

app/core_plugins/chat/conversation_title.py
```diff
@@ -43,17 +44,18 @@ async def generate_conversation_title(
     first_user_message: str,
     service: ChatService,
     provider: BaseChatProvider,
+    config: ChatSystemConfig,
```
regisb: What's the point of a `get_chat_system_config` function if we need to pass the config object around in all child function calls?

app/core_plugins/chat/routes/attachments.py
```diff
@@ -0,0 +1,95 @@
...
+async def list_conversation_attachments(
+    conversation: Conversation = Depends(get_owned_conversation),
+    session: AsyncSession = Depends(get_async_session),
+    service: ChatService = Depends(get_chat_service),
+) -> list[AttachedDriveFileResponse]:
+    drive_files = await service.list_conversation_attachments(
+        session,
+        conversation_id=conversation.id,  # type: ignore
+    )
+    return [AttachedDriveFileResponse(id=f.id, name=f.name, size=f.size) for f in drive_files]  # type: ignore
```
regisb: Why the `# type: ignore`? There are about ~15 in this PR. Were they backported from a previous change? Can we get rid of them?

app/core_plugins/chat/tests/conftest.py
```diff
@@ -0,0 +1,135 @@
+import os
+from collections.abc import AsyncGenerator, Generator
+from typing import Any, cast
+from unittest.mock import MagicMock
+
+# Set test environment variables BEFORE importing app modules.
+os.environ.setdefault("DATABASE_URL", "sqlite+aiosqlite:///:memory:")
+os.environ.setdefault("SECRET_KEY", "test-secret-key")
+os.environ.setdefault("RAG_MCP_URL", "http://localhost:8000")
+os.environ.setdefault("LLM_ENCRYPTION_KEY", "QL9oJuLxl0gKCbJpQgkzrdlsZUmvIVR3Cp0gSPcVLvQ=")
```
regisb: All the settings above are core settings, and should thus be defined in a `tests/lib/env.py` (or similarly named) module or file. Otherwise we have to duplicate the definition of core settings with core tests.

app/core_plugins/chat/tests/conftest.py
```diff
@@ -0,0 +1,135 @@
...
+os.environ.setdefault("SLACK_CLIENT_ID", "test-slack-client-id")
+os.environ.setdefault("SLACK_CLIENT_SECRET", "test-slack-client-secret")
+os.environ.setdefault("SLACK_SIGNING_SECRET", "test-slack-signing-secret")
+os.environ.setdefault("SLACK_REDIRECT_URI", "http://localhost:7727/api/v1/slack/callback")
```
regisb: Why does the chat plugin need to define Slack settings? In my understanding the Slack plugin depends on the chat plugin, but not vice-versa.

app/core_plugins/chat/tests/conftest.py
```diff
@@ -0,0 +1,135 @@
...
+from sqlmodel.ext.asyncio.session import AsyncSession
+
+from app.api.v1.auth import get_current_user
+from app.core.db import get_async_session
+from app.core_plugins.chat.routes import chat_router
+from app.main import app  # must be imported before chat_router to avoid circular imports
+from app.models.user import User
+
+_CHAT_PREFIX = "/api/v1"
+_DATABASE_URL = "sqlite+aiosqlite:///:memory:"
```
regisb: We really should not have to duplicate this database url definition across every plugin. Instead, we should have a base test environment that we can share across tests.

app/core_plugins/chat/tests/conftest.py
```diff
@@ -0,0 +1,135 @@
...
+
+# Register chat routes for tests.
+# The plugin lifespan (which normally registers chat routes via
+# `plugin_manager.load_all_enabled`) does not run in tests, so we mount
+# the router directly. Same pattern as `tests/slack/conftest.py` and
+# `tests/googledrive/conftest.py`.
```
regisb: This is not a great pattern. Can we do better? Likewise, many of the functions below should be re-usable across tests.

app/core_plugins/chat/tests/test_conversation_title.py
```diff
@@ -156,6 +157,7 @@ async def _call(
                 first_user_message=first_user_message,
                 service=mock_service,
                 provider=mock_provider,
+                config=ChatSystemConfig(),
```
regisb: What is the point of the `get_chat_system_config` function if we don't use it?

app/core_plugins/chat/tests/test_rag_agent_integration.py
```diff
@@ -56,7 +56,7 @@ async def test_calls_get_context_via_agent(self) -> None:
     @pytest.mark.asyncio
     async def test_drive_file_not_found_returns_422(self) -> None:
         """Test that DriveFileNotFoundError returns 422."""
-        from app.core_plugins.chat.routes import _resolve_drive_file_blocks
+        from app.core_plugins.chat.routes.completions import _resolve_drive_file_blocks
```
regisb: Can we use this opportunity to move all imports to the top of the file?

app/core_plugins/chat/routes/completions.py
```diff
@@ -65,8 +49,7 @@
 if not isinstance(_RAG_CONTEXT_PROMPT, str):
     raise TypeError("rag_context_replacement_prompt asset must be a string")
 
-chat_router = APIRouter(prefix="/chat", tags=["Chat"])
-
+router = APIRouter()
 
 _PROVIDER_API_ERRORS = (
```
abdulrafey1: could we move this to `constants.py` and give it the name `LLM_PROVIDER_API_ERRORS`

app/core_plugins/chat/routes/completions.py
```diff
@@ -261,7 +216,156 @@ async def _persist_pre_stream_error(
         logger.exception("Failed to persist pre-stream error message for conversation %s", request.conversation_id)
 
 
-@chat_router.post("/completions", response_model=ChatCompletionResponse)
+async def _get_or_create_conversation(
```
abdulrafey1: lets move all chat routes related utility functions to a separate `app/core_plugins/chat/routes/utils.py` file.
Lets only keep route definitions in the relevant routes file. It would make it more readable and maintainable

Same for the following `_attach_request_drive_files`, `_persist_incoming_messages`, `_classify_in_scope`, `_resolve_tools` functions

app/core_plugins/chat/routes/completions.py
```diff
@@ -261,7 +216,156 @@ async def _persist_pre_stream_error(
...
+        title_provider = get_provider(
+            provider_name=provider_name,
+            api_key=api_key,
+            model=model,
+            temperature=config.title_llm_temperature,
+            max_tool_executions=0,
+        )
+        background_tasks.add_task(
+            generate_conversation_title,
+            conversation_id=conversation.id,  # type: ignore[arg-type]
```
abdulrafey1: we can get rid of this type-ignore by doing:
```
from typing import cast
...
conversation_id=cast(int, conversation.id),
```

The `typing.cast` function does nothing in itself, it just returns the same value, it is just a helper for mypy to know the type of argument being passed, in our case we know that `conversation.id` here is an `int`, since it is being retreived from the `Conversation` DB model and the model has an `int` column for `id`

app/core_plugins/chat/routes/completions.py
```diff
@@ -478,51 +503,40 @@ async def chat_completion(
         should_run_rag = False
```
abdulrafey1: lets create a utility function for rag_intent_routing as well, since most of the logic above has been abstracted into utility functions, we should do the same with this.

app/core_plugins/chat/routes/completions.py
```diff
@@ -261,7 +216,156 @@ async def _persist_pre_stream_error(
...
+    if request.tools and isinstance(request.tools, list):
+        tools = await tool_registry.get_tools_by_names(request.tools)
+        if not tools:
+            logger.warning("No tools found for: %s", request.tools)
+        return tools
+    return None
+
+
+@router.post("/completions", response_model=ChatCompletionResponse)
 async def chat_completion(
```
abdulrafey1, line 54: Even after being refactored this route handler spans almost 300 lines. Thats a clear indication that it should be broken down into more pieces.

One of the suggestions is to take the rag-intent-router logic out in a utility function, second is that the exception handling alone is almost one-sixth of this function, from line 385-to-399 and from line 627-to-668. Thats 55 lines just for exception handling

app/core_plugins/chat/routes/conversations.py
```diff
@@ -0,0 +1,185 @@
...
+)
+from app.core_plugins.chat.service import ChatService
+from app.models.user import User
+
+logger = get_logger(__name__)
+
+router = APIRouter()
+
+
+def _parse_metadata_list(model_metadata: str | None, key: str) -> list[dict[str, Any]] | None:
```
abdulrafey1: Same with this function, I think its better to move this to a `utils.py` file and only keep route handlers here

app/core_plugins/chat/routes/dependencies.py
zaira-bibi: It's a FastAPI dependency injection module. I think `api.py` would mean endpoint definitions, which these are not.

app/core_plugins/chat/routes/conversations.py
```diff
@@ -0,0 +1,185 @@
...
+        session=session,
+        conversation_id=conversation.id,  # type: ignore
+        limit=limit,
+        offset=offset,
+        exclude_errors=False,
+    )
+
+    message_count = len(messages)
+
+    return ConversationDetailResponse(
```
abdulrafey1, line 120: above in the `list_conversations` function, we are populating the `conversation_responses` list on line 65 and then making a clean and compact return on line 80.

But here we are populating the message list and returning all in one return statement. Lets break this down as well to keep this consistent. Lets keep the return statements short and concise

app/core_plugins/chat/routes/tool_endpoints.py
```diff
@@ -0,0 +1,31 @@
+from fastapi import APIRouter, Depends
```
abdulrafey1: for every other routes file, we did not use the `_endpoints` suffix, lets do the same here, a simple `tools.py` should be a good enough name.

PS: It might be a good thing to reconsider having this endpoint, do we really need it? let me know if I am missing something

app/core_plugins/chat/tests/test_conversation_title.py
```diff
@@ -1,5 +1,6 @@
 from unittest.mock import AsyncMock, MagicMock, patch
 
+from app.core_plugins.chat.config import ChatSystemConfig
```
abdulrafey1: can we use the `get_chat_system_config` function here?

app/core_plugins/chat/tests/test_rag_integration.py
```diff
@@ -7,7 +7,7 @@
 import pytest
 from fastapi import HTTPException
 
-from app.core_plugins.chat.routes import _extract_query_text, _resolve_drive_file_blocks
+from app.core_plugins.chat.routes.completions import _extract_query_text, _resolve_drive_file_blocks
```
abdulrafey1: these should've been imported from `utils.py` instead of `completions.py`

app/core_plugins/chat/tests/test_rag_status_events.py
```diff
@@ -586,31 +586,31 @@ async def _failing_stream(*_args: object, **_kwargs: object) -> AsyncGenerator[s
 
 class TestParseMetadataList:
     def test_returns_none_for_no_metadata(self) -> None:
-        from app.core_plugins.chat.routes import _parse_metadata_list
+        from app.core_plugins.chat.routes.conversations import _parse_metadata_list
```
abdulrafey1: lets move all the lazy imports form this file to the top of the file

app/core_plugins/chat/tests/test_scope_validation.py
```diff
@@ -78,7 +78,7 @@ class TestStreamOutOfScopeRefusal:
     @pytest.mark.asyncio
     async def test_emits_single_done_event_with_refusal_content(self) -> None:
         """_stream_out_of_scope_refusal yields exactly one SSE done event."""
-        from app.core_plugins.chat.routes import _stream_out_of_scope_refusal
+        from app.core_plugins.chat.routes.completions import _stream_out_of_scope_refusal
```
abdulrafey1: lets also move these lazy imports to the top of the file

app/core_plugins/chat/conversation_title.py
```diff
@@ -2,6 +2,7 @@
 
 from app.core.db import async_engine
 from app.core.logger import get_logger
+from app.core_plugins.chat.config import ChatSystemConfig
```
abdulrafey1: lets use the `get_chat_system_config` function here

app/core_plugins/chat/lms_credentials.py
abdulrafey1: This is a system prompt and should be moved to `app/core_plugins/chat/assets/lms_rules_system_prompt.txt`

app/core_plugins/chat/lms_credentials.py
```diff
@@ -19,19 +19,12 @@
 
 
 def _lms_tool_prefixes() -> tuple[str, ...]:
```
abdulrafey1, line 8: Could all these functions be bundled up in a class?
Lets prioritize class based solutions to plain functions

app/core_plugins/chat/tools.py
```diff
@@ -345,6 +353,15 @@ def _get_python_type(self, json_schema: dict[str, Any]) -> type:
         return type_map.get(json_type, str)
 
 
+def get_parameters_schema(args_schema: type[BaseModel] | dict[str, Any] | None) -> dict[str, Any]:
```
abdulrafey1, line 355: lets move this utility function to a utility file for the chat plugin and only keep tool specific code in this file

We can move it to `app/core_plugins/chat/utils.py`

app/core_plugins/chat/routes/dependencies.py
regisb: Oh OK I wasn't aware of that.

app/core_plugins/chat/lms_credentials.py
regisb: I disagree. There's no reason to bundle functions in a class if we don't persist state within class attributes. Modules work just fine.

app/core_plugins/chat/tools.py
regisb: Let's be careful with "utils" modules. Whenever possible, let's name the modules more accurately.

app/core_plugins/chat/routes/attachments.py
regisb: This is a great catch. Did you detect this with the help of an LLM?

tests/lib/env.py
```diff
@@ -0,0 +1,15 @@
+"""Test environment setup — import this before any app module that reads settings.
```
regisb: I think this file is unnecessary now that we have a base conftest.py and app/testing.py.

app/core_plugins/chat/routes/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+from app.core_plugins.chat.routes.conversations import (
+    router as _conversations_router,
+)
+from app.core_plugins.chat.routes.dependencies import (
+    get_chat_service,
+    get_chat_system_config,
+)
+
+__all__ = [
```
abdulrafey1: lets get rid of all `__all__` attributes throughout the plugin as suggested by regis in today's standup.

app/core_plugins/chat/constants.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
+
+import anthropic
+import httpx
+import openai
+from google.api_core import exceptions as google_exceptions
+
+from app.rag.utils import get_asset
+
+RAG_CONTEXT_PROMPT = get_asset("rag_context_replacement_prompt", "txt")
```
abdulrafey1: oh this is going to be a handful to manage depending on which PR gets merged first, #405 or this one 

Nothing to change here, just a note:
its just that I have relocated the `rag_context_replacement_prompt.txt` asset into the chat plugin since only this plugin uses it, but this PR still gets it from inside the rag module 😢

app/core_plugins/chat/routes/dependencies.py
```diff
@@ -0,0 +1,41 @@
...
+from app.api.v1.auth import get_current_user
+from app.core_plugins.chat.config import ChatSystemConfig
+from app.core_plugins.chat.models import Conversation
+from app.core_plugins.chat.service import ChatService
+from app.lib.db import get_async_session
+from app.models.user import User
+
+
+@lru_cache
+def get_chat_system_config() -> ChatSystemConfig:
```
abdulrafey1: lets move this to `config.py` and rename the function to `get_chat_settings` to stay consistent with the core and rest of the codebase

tests/lib/fixtures.py
```diff
@@ -0,0 +1,108 @@
+"""Shared pytest fixtures re-exported from ``app.testing``.
```
abdulrafey1: Not sure why they need to be moved from `app.testing` to here, can you please explain why?

abdulrafey1 (approved):
Lets get this merged!!!

hamza-56 (approved):

tests/lib/fixtures.py
zaira-bibi: i think a rebase placed them here. i'll revert the changes.

PR #381, opened 2026-06-01
refactor(docker): rename api to app, add healthchecks, parameterize credentials

### LINKED ISSUE #380: Refactor docker-compose for reliability and configurability
## Problem
The docker-compose setup had hardcoded Postgres credentials, no healthchecks on `db` or `redis`, an ambiguously named `api` service, and stale Makefile targets. This made local dev setup brittle (services could start before the DB was ready) and harder to configure for teams that want non-default credentials.

## Acceptance Criteria
- [x] `api` service renamed to `app` throughout compose files and Makefile
- [x] All services have explicit `container_name` values
- [x] Postgres credentials use env vars with sane defaults (`${POSTGRES_USER:-sparkth}`)
- [x] `db` and `redis` services have healthchecks
- [x] `depends_on` uses `condition: service_healthy` / `condition: service_completed_successfully`
- [x] `x-ml-env` YAML anchor DRYs up `TOKENIZERS_PARALLELISM` and `HF_HUB_DISABLE_XET`
- [x] `make logs` and `make shell` accept an optional service argument
- [x] `.env.example` documents the new Postgres env vars
- [x] `CLAUDE.md` updated to reflect new `make shell` / `make logs` signatures

## Changes
- refactor(docker): rename api→app, add healthchecks, parameterize credentials

## Notes
- `.env.example` now lists `POSTGRES_USER`, `POSTGRES_PASSWORD`, `POSTGRES_DB` — teams may want to populate these in their `.env` before re-running `make up`
- `SMTP_FROM_EMAIL` and `SMTP_FROM_NAME` were removed from the hard-coded `app` env block; they must now come from `.env`
- `rag-cleanup` no longer mounts `./app` — it uses the image-baked source, consistent with how other one-shot services behave

## Pull Request
https://github.com/edly-io/sparkth/pull/381

abdulrafey1:
## What
Hardens the docker-compose setup by renaming the `api` service to `app`, adding healthchecks to `db` and `redis`, replacing hardcoded Postgres credentials with parameterized env vars, and making `make shell` / `make logs` accept an optional service argument.

## Changes
- `refactor(docker)`: rename `api` service to `app` in `docker-compose.yml` and `docker-compose.dev.yml`
- `refactor(docker)`: add `container_name` to all services
- `refactor(docker)`: replace hardcoded Postgres credentials with `${POSTGRES_USER:-sparkth}` env vars
- `refactor(docker)`: add healthchecks to `db` (pg_isready) and `redis` (redis-cli ping)
- `refactor(docker)`: update `depends_on` to use `condition: service_healthy` / `condition: service_completed_successfully`
- `refactor(docker)`: extract `TOKENIZERS_PARALLELISM` and `HF_HUB_DISABLE_XET` into `x-ml-env` YAML anchor
- `refactor(docker)`: add `pull_policy: never` to `app` service
- `refactor(docker)`: simplify migrations command to `alembic upgrade head` (remove misleading echo)
- `refactor(docker)`: remove `SMTP_FROM_EMAIL`/`SMTP_FROM_NAME` from hard-coded app env block
- `refactor(docker)`: remove `./app` volume mount from `rag-cleanup` (one-shot, uses image-baked source)
- `chore(docker)`: update Makefile — `api` → `app` in all targets; `make logs`/`make shell` accept `$(ARGS)`
- `chore(docs)`: update `.env.example` with `POSTGRES_USER`, `POSTGRES_PASSWORD`, `POSTGRES_DB`
- `docs`: update `CLAUDE.md` to reflect new `make shell [service]` and `make logs [service]` signatures

## How to Test
1. Copy `.env.example` → `.env` and fill required values (`SECRET_KEY`, `LLM_ENCRYPTION_KEY`, etc.)
2. Run `make base` if the `sparkth-base:local` image is not present
3. Run `make up` — verify all services start cleanly and `sparkth-app` appears in `docker ps`
4. Run `make shell` — should open a shell in the `app` container
5. Run `make shell service=db` → should open a shell in the `db` container (`make shell ARGS=db`)
6. Run `make logs ARGS=app` — should tail only the app container logs
7. Stop and wipe with `make clean`, then `make up` again — migrations should run before `app` starts
8. Run `make rag-cleanup` and confirm it completes without error

## Notes
- **Env var change:** `POSTGRES_USER`, `POSTGRES_PASSWORD`, `POSTGRES_DB` are now read from the environment (with safe defaults). Existing `.env` files that omit these will still work via the `:-` defaults.
- **SMTP:** `SMTP_FROM_EMAIL` and `SMTP_FROM_NAME` were removed from the hard-coded compose env block. They must now come from `.env`. Existing `.env` files that already set these are unaffected.
- No database migrations required.
- No breaking changes to the REST API or MCP server.


docker-compose.yml
```diff
@@ -1,48 +1,46 @@
 services:
-  # ----------------------------------------
-  # Sparkth API Service
-  # ----------------------------------------
-  api:
+  # Sparkth App
+  app:
+    container_name: sparkth-app
```
regisb: Having container names that are different from service names is confusing. With docker compose, all container names are prefixed with the parent directory name ("sparkth"). So why define a "sparkth-" prefix? I suggest we get rid of all `container_name`.

.env.example
```diff
@@ -1,4 +1,15 @@
-DATABASE_URL=postgresql://USERNAME:PASSWORD@HOST:PORT/DATABASE
+# Postgres credentials (used by the docker-compose db service)
```
regisb: I'm confused by this .env.example file which looks more and more like an .env.dev file with dev settings -- but never quite becomes 100% compatible with a dev environment.

I think we should embrace the fact that .env is really .env.dev, our docker-compose environment is for development only, and we should provide specific instructions for production deployment.

regisb (changes requested):

docker-compose.yml
abdulrafey1: without the `container_name` field, containers show up as `db-1`, `migrations-1`. We can get rid of the `sparkth-` prefix but I suggest we keep the `container_name` field to avoid the number suffix added by docker

regisb: Yes, this is a convention by docker compose to allow running the same set of services in different directories. In theory, we could then run multiple sparkth applications, which would be named `app-1`, `app-2`, etc. It's not a bug, it's a feature!

.env.example
abdulrafey1: what do you suggest we do? Rename this file to `.env.dev` or create a new file with populated env which will be committed to the repo? 🤔 

Even if we commit a `.env.dev` to the repo (for easier development setup), we still cant commit our google oauth secrets in it.

docker-compose.yml
abdulrafey1: yes i know its a feature, but the numbers in names sometime annoy me 😞 
But ok, we can skip the `container_name` field

.env.example
regisb: I suggest we keep `.env`, make it fully compatible with our dev environment, and add comments to explain how to make changes to it in production. Same with docker-compose.yml, which should be merged with docker-compose.dev.yml. But I'll be happy with any other solution that 1. makes the dev onboarding easier and 2. makes it trivial to keep the dev environment up-to-date.

CLAUDE.md
```diff
@@ -125,6 +135,18 @@ For every new feature, endpoint, service method, utility, or plugin tool:
...
 For bug fixes: write a test that reproduces the bug first, verify it fails, then fix.
 
+## Documentation Hygiene
+
+**Always update documentation alongside every code change — no exceptions.**
+
+Documentation includes:
+
+- **Docstrings** — module, class, and function docstrings must reflect current behaviour. If a function no longer does what its docstring says, update the docstring in the same commit.
+- **Inline comments** — remove or update comments that describe logic that has changed. Never leave comments that contradict the code.
```
regisb, line 145: Glad to see this. I noticed that Claude was very sparse on comments.

docker-compose.yml
```diff
@@ -1,145 +1,120 @@
...
     build:
       context: .
       dockerfile: Dockerfile.local
       # Requires sparkth-base:local — run `make base` first (or `make up` handles it automatically)
     image: sparkth:local
+    pull_policy: never
     restart: unless-stopped
     ports:
       - "127.0.0.1:7727:7727"
+    command: ${APP_COMMAND:-fastapi run --host 0.0.0.0 app/main.py}
```
regisb, line 25: Add a comment that explains how to change this in production

docker-compose.yml
```diff
@@ -1,145 +1,120 @@
...
-  # ----------------------------------------
-  api:
+  # Sparkth App
+  app:
     build:
       context: .
       dockerfile: Dockerfile.local
       # Requires sparkth-base:local — run `make base` first (or `make up` handles it automatically)
     image: sparkth:local
+    pull_policy: never
```
regisb, line 13: Add comments here: the image name should be different in production, and we should definitely pull the remote image.

The goal here is to document how to transition from dev to prod, though we don't need (yet) to provide a fully functional prod environment.

regisb (approved):
This will be good to merge after adding just a few comments.

Issue #379, opened 2026-06-01
[EPIC] Plugins are missing a well-defined core API


regisb:
At the moment, plugins import from `app.*` (including `app.core.*`) because they don't have any API to rely on. For instance, the Slack plugin has the following imports in `routes.py`:

```
from app.api.v1.auth import get_current_user
from app.core.cache import get_cache_service
from app.core.config import get_settings
from app.core.db import async_engine, get_async_session, get_session
from app.core.encryption import get_encryption_service
from app.core.logger import get_logger
[app.core_plugins.slack imports are fine]
from app.llm.providers import BaseChatProvider, get_provider
from app.llm.service import LLMConfigService
from app.models.user import User
from app.rag.store import VectorStoreService
from app.services.plugin import PluginService
```

This is a serious issue, because it implicitly means that every function in Sparkth is part of the API. And that means that we can't change anything in Sparkth without breaking downstream plugins. This makes it effectively impossible to develop 3rd-party plugins.

(I realised this while investigating how to simplify the `app.rag.store.VectorStoreService`. I realised that this class is really just a module with no instance attributes. Its only use is really to be easily mocked in unit tests. But then I noticed that it's being imported in by `app.core_plugins.slack.routes`. )

Ideally, Sparkth should expose different libraries for different parts of the platform. For instance:

```
from sparkth.lib.rag import ...
from sparkth.lib.logging import get_logger
from sparkth.lib.types import ...
etc.
```

But to implement this, we need to decide which functions are part of the Sparkth library. Once we do, we should be able to do the following:

- publish a reference documentation of the Sparkth library
- forbid core imports from plugins
- avoid changes to the Sparkth API

---

## Core API perimeter

The following catalogues the de-facto public API today: every `app.*` symbol imported by the built-in plugins (`canvas`, `openedx`, `chat`, `googledrive`, `slack`), excluding intra-plugin (`app.core_plugins.*`) imports. This is the surface that must be deliberately defined — or removed from the plugin surface — before the library boundary can be enforced.

Two usage signals are reported:

- **Plugins** — how many of the 5 plugins import the symbol (breadth; the strongest signal that a symbol is genuinely shared).
- **Sites** — number of import sites across all plugin files (depth).

The **Done** column tracks remediation: add a ✅ and the PR number once the symbol has been given a stable boundary or removed from the plugin surface.

### Symbols imported by 2 or more plugins

| # | Symbol | Module | Plugins | Sites | Used by | Done |
[...]

PR #387, opened 2026-06-02
Refactor and simplify PluginManager (renamed to PluginLoader)

regisb:
## What

This is ongoing work to simplify the Plugin API (see #379). We start by drastically simplifying the PluginManager, now renamed to PluginLoader. Also, we resolve a circular dependency by moving plugin object creation to the PluginService.

## Notes

This should not be a breaking change.


app/plugins/loader.py
```diff
@@ -0,0 +1,158 @@
...
+            Continues loading other plugins if one fails.
+            Check logs for any failures.
+        """
+        for plugin_name, plugin_class in self.iter_plugin_classes():
+            try:
+                plugin_instance = plugin_class(plugin_name)
+            except (PluginLoadError, PluginNotFoundError, PluginAlreadyLoadedError) as e:
+                logger.error(f"Failed to load plugin '{plugin_name}'")
+                logger.exception(e)
+                continue
```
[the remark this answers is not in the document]
regisb: Actually, I want plugin loading to never crash the system, so we'll just catch all exceptions here.

app/services/plugin.py
```diff
@@ -128,6 +130,28 @@ async def get_by_name(self, session: AsyncSession, name: str) -> Plugin | None:
...
+        Load and instantiate all plugins. This must be called at startup time to make
+        sure that all plugins exist in the database.
+
+        Raises:
+            PluginLoadError: If plugin fails to load
+        """
+        plugin_loader = get_plugin_loader()
+        for plugin_name, plugin_instance in plugin_loader.get_loaded_plugins():
+            try:
+                async with AsyncSession(async_engine, expire_on_commit=False) as session:
```
abdulrafey1: can we use the `app.core.db.get_async_session` helper here

app/plugins/base.py
```diff
@@ -557,24 +519,25 @@ def get_mcp_tools(self) -> list[dict[str, Any]]:
...
-                        self.add_middleware(
-                            Middleware(
-                                CORSMiddleware,
-                                allow_origins=["*"],
-                                allow_methods=["*"]
-                            )
-                        )
+        def __init__(self, name: str):
+            super().__init__(name)
+            from starlette.middleware.cors import CORSMiddleware
```
abdulrafey1, line 532: do we need to have this lazy import in this code example? because claude might start making this lazy middleware import looking at this.

app/plugins/loader.py
```diff
@@ -0,0 +1,158 @@
...
+Manages plugin discovery and instantiation.
+"""
+
+import importlib
+import inspect
+import re
+from typing import Iterator, Type
+
+from app.core.config import get_plugin_settings
+from app.core.logger import get_logger
```
abdulrafey1: If #385 gets merged before this, we should update this import to:
`from app.lib.log import get_logger`

app/plugins/loader.py
```diff
@@ -0,0 +1,158 @@
...
+        Discover all plugins defined in get_plugin_settings().
+
+        Yields:
+            (name, class) tuples
+
+        Note:
+            All plugins are enabled by default.
+            Format: "module.path:ClassName"
+        """
+        for module_string in get_plugin_settings():
```
abdulrafey1, line 77: the `app.core.config.get_plugin_settings` returns a list of hard-coded plugin paths. Any time a new plugin is created, this `list` inside the core of sparkth will have to be updated. I believe we are trying to fix this in our refactoring effort .

My question is that should we be using this function here? or should we come up with a better way of discovering plugin module paths, since we already have a PR for this, might as well create a new function for discovering plugins.

app/services/plugin.py
abdulrafey1: can we use the `app.core.db.get_async_session` helper here?

abdulrafey1 (commented):
A few suggestions to note

app/plugins/base.py
regisb: Yes, good point.

app/plugins/loader.py
regisb: Yes, this is on my TODO-list, but I didn't want to add more scope to this PR. I wanted to get this merged before I moved on to the next step of refactoring.

app/plugins/base.py
regisb: Though on a second note it looks like this `add_middleware` method is not used anywhere 😅 I'll create a PR to address this.

abdulrafey1 (approved):
looks good

PR #388, opened 2026-06-03
refactor(tests): reorganize tests

regisb:
## What

- Plugin tests are moved to app/core_plugins/<name>/tests/
- a shared testing.py library is created
- duplicated conftest code is refactored to be shared across tests
- Duplicated openedx tests are deleted (the last openedx/tests.py file was never run)
- "testpaths" is removed from pyproject.yml to avoid missing some tests
- RAG tests are moved to tests/rag


app/core/config.py
```diff
@@ -23,10 +23,10 @@ class Settings(BaseSettings):
...
 
     # Slack TA Bot OAuth
-    SLACK_CLIENT_ID: str
-    SLACK_CLIENT_SECRET: str
-    SLACK_SIGNING_SECRET: str
-    SLACK_REDIRECT_URI: str
+    SLACK_CLIENT_ID: str = ""
+    SLACK_CLIENT_SECRET: str = ""
+    SLACK_SIGNING_SECRET: str = ""
+    SLACK_REDIRECT_URI: str = ""
```
abdulrafey1, line 29: I believe that we had these as required so that incase any of these is missed in a deployment (dev or prod), the deployment would crash instead of initially going up and then failing in a couple of usecases.

Let me know if you have a better approach

app/core_plugins/slack/tests/__init__.py
abdulrafey1, line 1: I think these are already being moved in #361, but we can rebase depending on which PR get merged first

app/core_plugins/googledrive/tests/__init__.py
abdulrafey1, line 1: The googledrive tests are being moved to their plugin in #377 
Nothing to change here, just thought it would be good to point it out

abdulrafey1 (approved):
LGTM, just some notes below, nothing to change.

One things I'd like to add is to update the `CLAUDE.md` file so it understands how we want to organize our tests and how it should use `app.testing` as a shared lib for writing tests.

app/core/config.py
regisb: My problem with the current approach is that the Slack config leaks into the core. In the future I hope to have the Slack plugin inject these settings at runtime and perform the checks itself.

regisb:
Requesting another review because I moved RAG tests to tests/rag @abdulrafey1.

abdulrafey1 (approved):
LGTM

PR #390, opened 2026-06-03
refactor(core): move db session API into app.lib.db

regisb:
Plugins imported get_session, get_async_session, and async_engine straight from app.core.db, making the raw SQLAlchemy engine part of the implicit plugin surface (issue #379). Expose a curated public API in app.lib.db instead: session_scope (an async context manager that owns the AsyncSession(async_engine, expire_on_commit=False) idiom that was copy-pasted at 8 sites) plus the two FastAPI session dependencies, moved out of app.core.db. The engines themselves stay in app.core.db.

session_scope takes an optional expire_on_commit (default False, async-safe); the sync get_session keeps the True default explicitly to document the deliberate asymmetry. The knob lives on session_scope, not on the FastAPI deps, since FastAPI would otherwise expose it as a query parameter.

Lib tests move to tests/lib so they inherit tests/conftest.py (test-env setup + the shared in-memory engine fixture) now that app.lib.db imports app.core.db at module load.

This is to address part of issue #379.

## What

Create a shared library for sync/async database session access.

### DIFF
--- a/app/lib/db.py
+++ b/app/lib/db.py
@@ -1,0 +1,95 @@
+"""Database session access for Sparkth — the curated public session API.
+
+This is the single public entry point for obtaining a database session. All code,
+including plugins, should acquire sessions from here rather than reaching for the
+raw SQLAlchemy engines in :mod:`app.core.db`.
+
+The engines themselves stay in :mod:`app.core.db` (they read settings); this
+module is only the public face that hands out sessions over them.
+"""
+
+from collections.abc import AsyncGenerator, Generator
+from contextlib import asynccontextmanager
+
+from sqlmodel import Session
+from sqlmodel.ext.asyncio.session import AsyncSession
+
+from app.core.db import async_engine, engine
+
+
+@asynccontextmanager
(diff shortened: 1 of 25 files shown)


app/lib/db.py
```diff
@@ -0,0 +1,95 @@
...
+    exits (whether normally or via an exception). Always use it as a context
+    manager so the connection is never leaked::
+
+        from app.lib.db import session_scope
+
+        async with session_scope() as session:
+            session.add(obj)
+            await session.commit()
+
+    Use this for code that runs **outside** an HTTP request — background tasks,
```
abdulrafey1, line 36: This part of the doc-string feels like a module level doc-string. WDYS?

abdulrafey1 (approved):
just one small note, rest looks good

hamza-56 (approved):

app/lib/db.py
regisb: hmmm no, I think this comment is only valid for this specific function, so it makes sense to keep it here.

Issue #369, opened 2026-05-27
CI is not running many tests


regisb:
Currently, CI is facing the following issues:

## Not running frontend tests

Linting and formatting tests are run, but not vitest tests. In other words, we should be running `make tests.frontend`.

## Broken frontend tests

If CI did run frontend tests, we would have realised that they fail:

```
$ make test.frontend
...
 Test Files  2 failed | 5 passed (7)
      Tests  13 failed | 51 passed (64)
   Start at  13:22:39
   Duration  13.18s (transform 503ms, setup 436ms, import 2.21s, tests 14.12s, environment 4.78s)

error: "vitest" exited with code 1
```

## Backend tests pass in CI but fail locally: (EDIT: ✅ fixed in https://github.com/edly-io/sparkth/pull/377)

```
$ uv run pytest
...
FAILED app/rag/tests/test_provider_singleton.py::test_init_warms_model - RuntimeError: Error: cannot set number of interop threads after parallel work has started or set_num_interop_threads called
FAILED app/rag/tests/test_provider_singleton.py::test_get_returns_same_instance - RuntimeError: Error: cannot set number of interop threads after parallel work has started or set_num_interop_threads called
FAILED app/rag/tests/test_provider_singleton.py::test_init_reads_settings - RuntimeError: Error: cannot set number of interop threads after parallel work has started or set_num_interop_threads called
FAILED app/rag/tests/test_provider_singleton.py::test_reset_clears_singleton - RuntimeError: Error: cannot set number of interop threads after parallel work has started or set_num_interop_threads called
```

## Tests are hardcoded in .github/workflows/ci.yml

Instead of duplicating test commands in ci.yml, we should run simple `make` commands there, such that failures are easy to reproduce locally. I recommend that we refactor ci.yml as well as `make test.*` target to make them more explicit.

regisb:
EDIT: ✅  this was fixed in https://github.com/edly-io/sparkth/pull/377

## Root cause of the backend `test_provider_singleton.py` failures

These four failures are a torch-path test-hygiene bug, masked in CI because CI never installs torch.

[...]

PR #394, opened 2026-06-03
refactor(ci): run high-level make test targets

hamza-56:
## Related: [CI is not running many tests](https://github.com/edly-io/sparkth/issues/369)

Addresses the "Tests are hardcoded in .github/workflows/ci.yml" part of #369.

## What
Collapse the per-sub-target CI steps into single `make test.frontend` / `make test.backend` invocations so a CI failure reproduces locally with exactly one command. As part of this, fold the standalone react-doctor workflow into the same high-level frontend target instead of running it as a parallel, duplicate gate.

## Changes
- refactor(ci): run `make test.frontend` (lint + react-doctor + vitest + format) in the frontend job
- refactor(ci): run `make test.backend` (lint + mypy + pytest + format) in the backend job
- ci(frontend): delete the standalone `react-doctor.yml` workflow now that `make test.frontend` runs react-doctor (it duplicated the check on every frontend PR)
- ci(frontend): give the frontend job `fetch-depth: 0` and pass `REACT_DOCTOR_BASE=origin/<base>` + `RD_ARGS=--annotations` so the diff resolves against the PR base and findings still surface as inline annotations

## How to Test
1. `make test.frontend` passes locally (lint, react-doctor, vitest, format check).
2. `make test.backend` passes locally (lint, mypy, pytest, format check).
3. Confirm the CI run for this PR is green with one frontend and one backend test step, and that no separate "React Doctor" check runs.
4. On a PR that touches `frontend/**` with a react-doctor finding (e.g. an array index used as a React `key`), the frontend job fails and the finding shows as an inline annotation.

## Notes
- No production code changed. The bundled targets already existed in the Makefile (`test.frontend`, `test.backend`); this calls them from CI and removes the now-redundant react-doctor workflow added in #396.
- react-doctor stays diff-scoped with `--fail-on warning`: pre-existing debt in untouched files does not block; new/changed code must be clean.


abdulrafey1 (approved):
there is a merge conflict, rest looks good

.github/workflows/ci.yml
```diff
@@ -24,17 +31,15 @@ jobs:
...
-      - name: Run unit tests
-        run: make test.frontend.vitest
-
-      - name: Check formatting (oxfmt)
-        run: make test.frontend.format
+      # Single high-level target so a CI failure reproduces locally with
+      # exactly `make test.frontend` (lint + react-doctor + vitest + format check).
+      # Override the react-doctor base to the PR's actual base and emit inline annotations.
+      - name: Run frontend tests
+        run: make test.frontend REACT_DOCTOR_BASE=origin/${{ github.event.pull_request.base.ref }} RD_ARGS=--annotations
```
regisb, line 38: I like how this change simplifies the CI script and makes it consistent with our Makefile. But I wonder if these values shouldn't be the defaults for all users? I don't know, it's your call.

regisb (approved):

PR #396, opened 2026-06-03
ci(frontend): gate PRs on react-doctor instead of commenting

### LINKED ISSUE #375: CI fails to run react-doctor
The [react-doctor](https://github.com/edly-io/sparkth/blob/regisb/project-management-skill/.github/workflows/react-doctor.yml) workflow fails to run in pull request reviews.

See for instance this PR: https://github.com/edly-io/sparkth/pull/368
Which triggered this workflow: https://github.com/edly-io/sparkth/actions/runs/26739656086/workflow
Which failed with:

```
(Line: 27, Col: 15): Expected format {org}/{repo}[/path]@ref. Actual 'millionco/react-doctor@react-doctor@0.0.38'
```

The fact that react-doctor is failing is one issue, but another one is that CI still greenlights PRs despite the failing job runs.  This is a little strange...

hamza-56:
## What
Make react-doctor a real CI gate (and a locally runnable check) instead of a PR-only commenter that never failed the build.

## Changes
- ci(frontend): run the react-doctor CLI via `make lint.frontend.react-doctor` in the workflow, with `--annotations` for inline findings; drop the `pull-requests: write` comment mode
- chore(frontend): add `make lint.frontend.react-doctor` target (scans files changed vs `REACT_DOCTOR_BASE`, default `main`; `--fail-on warning`)
- docs(core): document the new target in CLAUDE.md

## How to Test
1. Run `make lint.frontend.react-doctor` on a clean branch off `main`: scans nothing, exits 0.
2. Introduce a change with a react-doctor finding (e.g. an array index used as a React `key`) and re-run: the target reports the finding and exits non-zero.
3. On a PR touching `frontend/**`, the React Doctor job runs the same command against `origin/main` and gates the PR, surfacing findings as inline annotations.

## Notes
- Strictest gate per maintainer direction: diff-scoped, `--fail-on warning`. Pre-existing findings in untouched files do not block; new/changed code must be clean.
- A full scan currently reports 5 errors + 253 warnings of pre-existing debt (e.g. `frontend/lib/useRagStatusPolling.ts:33`). Those are not gated unless a PR touches the relevant files. Cleaning them up is follow-up work.
- react-doctor CLI is pinned to `0.2.16`. No new committed dependency; it is fetched via `bunx` in CI and locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,6 +191,12 @@ lint.frontend: ## Check frontend lint errors (oxlint)
 lint.backend: ## Check backend lint errors (ruff)
 	uv run ruff check
 
+# Base branch react-doctor diffs against; override in CI (e.g. origin/main).
+REACT_DOCTOR_BASE ?= main
+.PHONY: lint.frontend.react-doctor
+lint.frontend.react-doctor: ## Run react-doctor on files changed vs REACT_DOCTOR_BASE (default main)
+	cd frontend && bunx react-doctor@0.2.16 . --diff $(REACT_DOCTOR_BASE) --fail-on warning $(RD_ARGS)
+
 .PHONY: lint.fix.frontend
 lint.fix.frontend: ## Auto-fix frontend lint errors (oxlint)
 	cd frontend && bun run lint:fix
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ test.backend.format: ## Run backend formatting tests
 	$(MAKE) lint.format.backend check=1
 
 .PHONY: test.frontend
-test.frontend: lint.frontend test.frontend.vitest test.frontend.format ## Run frontend linting, unit and formatting tests
+test.frontend: lint.frontend lint.frontend.react-doctor test.frontend.vitest test.frontend.format ## Run frontend linting, react-doctor, unit and formatting tests
 
 .PHONY: test.frontend.vitest
 test.frontend.vitest: ## Run frontend unit tests (make test.frontend.vitest [path] [with-coverage=1])
--- a/.github/workflows/react-doctor.yml
+++ b/.github/workflows/react-doctor.yml
@@ -8,29 +8,33 @@ on:
     paths:
       - "frontend/**"
 
+concurrency:
+  group: react-doctor-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   react-doctor:
+    name: React Doctor
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     timeout-minutes: 5
     permissions:
       contents: read
-      pull-requests: write
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
         with:
-          ref: ${{ github.event.pull_request.head.ref }}
           fetch-depth: 0 # required for --diff
 
-      - name: Run React Doctor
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -191,6 +191,12 @@ lint.frontend: ## Check frontend lint errors (oxlint)
 lint.backend: ## Check backend lint errors (ruff)
 	uv run ruff check
 
+# Base branch react-doctor diffs against; override in CI (e.g. origin/main).
+REACT_DOCTOR_BASE ?= main
+.PHONY: lint.frontend.react-doctor
+lint.frontend.react-doctor: ## Run react-doctor on files changed vs REACT_DOCTOR_BASE (default main)
```
regisb, line 197: 1. Should we run this target as part of `make test.frontend`? (I think we should)
2. will it succeed on the main branch?

hamza-56: Good points, thanks.

1. Done in c4ca5a1 — `lint.frontend.react-doctor` now runs as part of `make test.frontend`, so local runs gate on it alongside lint/vitest/format. (`ci.yml` still doesn't invoke `test.frontend`, and PR gating in CI stays on the dedicated `react-doctor.yml` workflow which sets `fetch-depth: 0` for the diff.)

2. Yes. On `main` the diff base resolves to `main` itself, so the diff is empty: react-doctor reports "No changed source files, skipping" and exits 0. Verified locally.

regisb (approved):

abdulrafey1 (approved):

regisb:
I like how we now have a convenient `make lint.frontend.react-doctor` command, which is (conveniently) part of `make test` and `make test.frontend`. But I don't like the fact that this command is prompting me to fix the issues and not showing the actual issues:

image

I think that I'd like instead to run the command with `--annotations --yes`, which adds the actual errors to stdout:

```
::warning file=lib/auth-context.tsx,line=8,title=react-doctor/only-export-components::Fast Refresh stops working when a file exports non-components.
::warning file=lib/auth-context.tsx,line=13,title=react-doctor/only-export-components::Fast Refresh stops working when a file exports non-components.
::warning file=lib/auth-context.tsx,line=3,title=react-doctor/no-react19-deprecated-apis::useContext is replaced by `use()` in React 19, which reads context inside ifs & loops too, so switch to `import { use } from 'react'`.
::warning file=lib/auth-context.tsx,line=111,title=react-doctor/no-effect-chain::Your screen redraws several times from a single action because one useEffect changes "token", which sets off this one.
::warning file=lib/auth-context.tsx,line=105,title=react-doctor/no-initialize-state::Your users see an extra render with empty "token" because a useEffect sets its starting value.
::warning file=lib/auth-context.tsx,line=107,title=react-doctor/no-initialize-state::Your users see an extra render with empty "loading" because a useEffect sets its starting value.
::warning file=lib/auth-context.tsx,line=113,title=react-doctor/no-derived-state::Storing "user" in state when you can derive it from other values costs an extra render.
::warning file=lib/auth-context.tsx,line=124,title=react-doctor/jsx-no-constructed-context-values::Every reader of this context redraws on each render because you build its `value` inline.
```

What do you think @hamza-56?

hamza-56:
Good catch! Switched to `--annotations --yes` and also dropped `--fail-on warning` so warnings don't block PRs. Fix is in #412.

PR #397, opened 2026-06-03
Introduce hook-based plugin API

### LINKED ISSUE #372: [EPIC] Plugin API is inconsistent, closed to extension, and disconnected from the frontend
## Summary

The plugin API exposed by `SparkthPlugin` (`app/plugins/base.py`) does not scale as a foundation for a plugin ecosystem. Its extension surface is a set of unrelated, hard-coded accessors; it cannot be extended without editing the base class; plugins cannot offer extension points to one another; and the backend plugin has no link to its frontend counterpart. These limitations make every new capability a special case and prevent plugins from composing.

## Where it happens

- `app/plugins/base.py` — `SparkthPlugin`, `PluginMeta`, the `@tool` decorator
- `app/plugins/manager.py` — `PluginManager`
- `app/plugins/adapters.py` — `PLUGIN_ADAPTERS` (config adapters living outside the plugin)
- `app/plugins/__init__.py` — `PLUGIN_CONFIG_CLASSES` (config classes living outside the plugin)
- `frontend/lib/plugins/registry.ts`, `frontend/lib/plugins/types.ts` — the independent frontend plugin registry

### 1. No unifying extension primitive

`SparkthPlugin` models each kind of contribution as its own separate list with its own add/get pair, all unrelated to each other:

| Concern | Mechanism |
|---|---|
| API routes | `add_route()` / `get_routes()` (+ `_tag_router_with_plugin`) |
| DB models | `add_model()` / `get_models()` |
| Migrations | `get_migrations_path()` |
| MCP tools | `@tool` → `PluginMeta._tool_registry` → `add_mcp_tool()` / `get_mcp_tools()` |
| Middleware | `add_middleware()` / `get_middleware()` |
| Config | `config_schema` / `get_config_schema()` + adapters in a *separate* `PLUGIN_ADAPTERS` dict |
| Lifecycle | `initialize()` / `enable()` / `disable()` |

There is no single concept that expresses "this is how a plugin contributes something to the system." Each row is a different mechanism, so the base class accumulates unrelated registries.

### 2. Closed for extension — every new contribution type means editing the base class

Adding a new *kind* of contribution (e.g. scheduled jobs, frontend nav entries, context providers) requires editing `SparkthPlugin` itself: a new `_x` list, a new `add_x()`, a new `get_x()`, plus new consumer code wherever those are loaded. The set of extension points is hard-coded into the framework.

This already leaks: config adapters didn't fit the existing accessors, so they were bolted on *outside* the plugin as a parallel `PLUGIN_ADAPTERS` dict (`app/plugins/adapters.py`) and a separate `PLUGIN_CONFIG_CLASSES` dict (`app/plugins/__init__.py`). When the n-th concern doesn't fit the existing slots, it ends up in a side-channel.

### 3. Extension points are framework-owned — plugins cannot extend each other

Only the core defines extension points. A plugin can *consume* the fixed slots, but it cannot *offer* a new slot for other plugins to fill. Canvas cannot declare "other plugins may contribute course importers to me"; the chat plugin cannot declare "other plugins may register context providers." Extension is centralized in the core, so plugins compose only with the framework, never with each other.

### 4. Backend and frontend plugin APIs ignore each other

The backend `SparkthPlugin` carries nothing frontend-facing — no display name, no component, no nav entry. The frontend maintains a **completely independent** registry (`frontend/lib/plugins/registry.ts`) where a `PluginDefinition` (`name`, `displayName`, `loadComponent`, routes, events) is registered by hand. The only thing tying the two together is a matching kebab-case `name` string — an implicit, uncheckable convention.

Consequences:
- A plugin's identity is split across two codebases with no single source of truth.
- The frontend can register a plugin the backend has never heard of (and vice versa).

A backend plugin should be able to declare its frontend counterpart so one manifest drives both tiers.

## Why it matters

Problems 1–3 are the same root issue at increasing magnification: there is no single, plugin-definable extension primitive, so concerns are ad-hoc (1), the surface is closed (2), and extension points are centralized (3). Problem 4 compounds it by splitting a plugin's declaration across two disconnected registries. Together they mean every new capability is a special case, and third parties cannot build plugins that extend the platform — or each other — without modifying core.

## Note

Problems 1–3 point toward a single uniform extension primitive that plugins can both consume and define, rather than a fixed set of framework-owned accessors. The Sparkth frontend already has a comparable event/handler system (`PluginEvent` / `PluginEventHandler` / `PluginEventType` in `frontend/lib/plugins/types.ts`); the backend, which needs it more, has no equivalent. The concrete design is open for discussion.

regisb:
## What

Create a more consistent plugin API.

## Changes

All features from SparkthPlugin are migrated to a hook-based API.


app/lib/mcp/hooks.py
```diff
@@ -0,0 +1,144 @@
...
+    except (TypeError, NameError, AttributeError) as e:
+        logger.warning(f"Failed to generate input schema for {func.__name__}: {e}. Using empty schema.")
+        return {"type": "object", "properties": {}}
+
+
+def type_to_json_schema(py_type: Type[Any]) -> dict[str, Any]:
+    """Convert a Python type to a JSON Schema type definition."""
+    # Check if it's a Pydantic BaseModel
+    try:
+        from pydantic import BaseModel
```
abdulrafey1, line 75: Lets move this lazy import to the top, Claude might start getting the wrong ideas because of this

abdulrafey1 (commented):
Everything looks good, just one question related to organizing the hooks, instead of having a `hooks.py` for everyhting can we have one hooks module and then add files in it for things like `routes`, `models` etc?

So everything in `app/lib/hooks.py` would move to `app/lib/hooks/__init__.py` and will be available for import with the same path `app.lib.hook`
and everything inside `app/lib/routes/hooks.py` would move to `app/lib/hooks/routes.py` and so on

Let me know if you think there is a downside in organizing hooks like this

regisb:
Thanks for the review Rafey! Locating imports is really tricky, and I do want to avoid any lazy or circular imports. I expect that hooks will import a bunch of modules. For instance, the router hook will import the APIRouter from FastAPI. For this reason, I think that we should not have all hooks in a single place, which would then quickly become a kitchen sink of imports and would almost certainly create circular dependencies.

app/main.py
```diff
@@ -42,19 +42,24 @@ async def plugin_lifespan(application: FastAPI) -> AsyncIterator[None]:
...
-                        tags = plugin.get_route_tags()
-                        tags_param: Union[list[Union[str, Enum]], None] = (
-                            cast(Union[list[Union[str, Enum]], None], tags) if tags else None
-                        )
-                        application.include_router(router, prefix=prefix if prefix else "", tags=tags_param)
-            except (AttributeError, TypeError, ValueError) as e:
-                logger.error(f"Failed to register routes for plugin '{plugin_name}': {e}")
+        from app.lib.routes.hooks import ROUTES
+
+        for plugin, (route_prefix, route_tags, router) in ROUTES.iter_items():
```
regisb, line 47: **Route registration lost its per-router error isolation.** The old code wrapped each plugin's `include_router` in `try/except (AttributeError, TypeError, ValueError)` and logged + continued. This loop has no per-item guard, and the only enclosing handler is `except (ImportError, RuntimeError, OSError)` on line 64. If any plugin's router raises `ValueError`/`TypeError`/`AttributeError` during `include_router` (e.g. a duplicate path operation or a malformed route), every subsequent plugin's routes silently fail to register and the uncaught exception propagates out of `plugin_lifespan` -> the app fails to start. Suggest restoring per-router isolation (try/except inside the loop).

app/lib/mcp/hooks.py
```diff
@@ -0,0 +1,144 @@
...
+    input schema from the handler signature).
+    """
+    tool_registry = getattr(type(plugin), "_tool_registry", {})
+
+    for method_name, tool_info in tool_registry.items():
+        try:
+            handler = getattr(plugin, method_name)
+            tool_def: dict[str, Any] = {
+                "name": tool_info["name"],
+                "handler": handler,
```
regisb, line 29: **These `WeakKeyDictionary` entries can never be GC-evicted.** `MCP_TOOLS.add_item(plugin, tool_def)` keys weakly on the plugin, but `tool_def["handler"]` is a bound method that strong-refs the plugin via `__self__`. The dict holds the value, the value keeps the weak key alive, so the entry lives until the dict is cleared. After `plugin_loader.unload_all()` the plugins and their tools are *not* freed. `tests/lib/test_hooks.py::test_hook_drops_items_when_plugin_is_garbage_collected` passes only because it stores plain ints — so it gives false confidence that the eviction the design relies on actually works for the real payload. Worth deciding whether the weakref is buying anything here vs. a plain dict the loader clears explicitly.

app/lib/hooks.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Generic, Iterator, TypeVar
+
+from app.plugins.base import SparkthPlugin
+
+T = TypeVar("T")
+
+
+class BasePluginHook(Generic[T]):
+    def __init__(self) -> None:
+        self._items: weakref.WeakKeyDictionary[SparkthPlugin, T] = weakref.WeakKeyDictionary()
```
regisb, line 11: **These hooks are process-global mutable state with no reset.** If a plugin class is instantiated more than once in a process (a test constructing a plugin directly, a future hot-reload, or a loader rebuilt while old instances are still alive — which the bound-method retention above guarantees they will be), `iter_items()` yields items keyed under *both* instances: `ROUTES` would include the same router twice (duplicate FastAPI routes), `MCP_TOOLS` yields duplicate tool defs. There is no teardown that resets the hooks, so the state only ever grows. Consider having the loader own/clear these registries rather than module-level globals.

regisb (commented):
Automated review of the hook-based plugin API refactor. The mechanics work for the singleton-loader happy path, but the move drops a route-registration guard and the global/weakref design has a few sharp edges worth a look. Findings posted inline, most important first.

app/migrations/env.py
```diff
@@ -35,12 +36,14 @@ def import_plugin_models():
...
-    plugin_loader = get_plugin_loader()
-    for plugin_name, plugin in plugin_loader.get_loaded_plugins():
-        models = plugin.get_models()
-        if models:
-            logger.info(f"Loaded {len(models)} model(s) from plugin '{plugin_name}'")
+    # Instantiate plugins so their models register into SQLModel.metadata and
+    # are contributed to the MODELS hook.
+    get_plugin_loader()
+    model_counts: dict[str, int] = {}
+    for plugin, _model in MODELS.iter_items():
```
regisb, line 43: **The `MODELS` hook is effectively dead infrastructure.** Its only consumer is this loop, and it's used solely to emit a per-plugin log-count line. Alembic autogenerate relies on `SQLModel.metadata`, which is populated by import side-effects when plugins are instantiated — independent of this hook. The comment on line 40 even says models "are contributed to the MODELS hook", which reads as load-bearing but isn't. The hook plus every `MODELS.add_item(...)` call in the chat/slack plugins could be dropped (or the count derived without it) with no functional change.

app/lib/config/hooks.py
```diff
@@ -0,0 +1,25 @@
...
+    """Yield (plugin_name, config_class) for every plugin that contributed a config."""
+    # Ensure plugins are instantiated so CONFIG_SCHEMAS is populated.
+    from app.plugins import get_plugin_loader
+
+    get_plugin_loader()
+    for plugin, config_schema in CONFIG_SCHEMAS.iter_items():
+        yield plugin.name, config_schema
+
+
+def get_plugin_config_schema(plugin_name: str) -> type[PluginConfig] | None:
```
regisb, line 20: **Behavior change: `google-drive` is now configurable where it was previously excluded.** The old `PLUGIN_CONFIG_CLASSES` was `{canvas, open-edx, chat, slack}` — no `google-drive` — so `PluginService.configure()` raised `InternalServerError` for it and `update_config` skipped field-filtering. `GoogleDrivePlugin` now calls `CONFIG_SCHEMAS.add_item(self, GoogleDriveConfig)`, so `get_plugin_config_schema("google-drive")` returns a class and google-drive becomes configurable via the user-plugins API. If intended, great (it removes a prior inconsistency); if not, it exposes a config surface that was previously rejected. (LMS-credential paths are unaffected — `GoogleDriveConfig` doesn't override `lms_tool_prefix`.)

app/lib/config/hooks.py
```diff
@@ -0,0 +1,25 @@
...
+# The Pydantic config class contributed by each plugin (one per plugin).
+CONFIG_SCHEMAS: PluginHook[type[PluginConfig]] = PluginHook()
+
+
+def iter_plugin_config_schemas() -> Iterator[tuple[str, type[PluginConfig]]]:
+    """Yield (plugin_name, config_class) for every plugin that contributed a config."""
+    # Ensure plugins are instantiated so CONFIG_SCHEMAS is populated.
+    from app.plugins import get_plugin_loader
+
+    get_plugin_loader()
```
regisb, line 15: **Minor efficiency:** `get_plugin_config_schema(name)` runs `iter_plugin_config_schemas()` -> `get_plugin_loader()` + a sort (`BasePluginHook._iter_plugin_items` sorts on every call) + a linear scan. `PluginService.sync_plugins` calls it once per plugin in a loop, making startup O(n^2) where the old `PLUGIN_CONFIG_CLASSES.get(name)` was O(1). `n` is small today, but this replaced a dict lookup with a repeated full scan + sort for no functional gain.

app/lib/routes/hooks.py
```diff
@@ -0,0 +1,6 @@
+from fastapi import APIRouter
+
+from app.lib.hooks import PluginCollectionHook
+
+# Each item is a (prefix, tags, router) triple contributed by a plugin.
+ROUTES: PluginCollectionHook[tuple[str, list[str], APIRouter]] = PluginCollectionHook()
```
regisb, line 6: **Maintainability:** the `(prefix, tags, router)` 3-tuple is unpacked positionally in `app/main.py:47`. `ROUTES.add_item(self, (tags, prefix, router))` by mistake would pass type-checking (both are str-ish) and silently mount routes at the wrong prefix. A `NamedTuple`/dataclass would make the field order self-documenting and catchable. Separately, the per-route plugin-tagging + endpoint stamping that used to live in `SparkthPlugin._tag_router_with_plugin` now sits inline in `app/main.py` — plugin internals leaked into the app entrypoint.

app/lib/mcp/hooks.py
regisb: This is a good point but I'll have a closer look at this when I actually refactor MCP tools.

app/migrations/env.py
regisb: OK let's get rid of the `.get_models()` logic then. See https://github.com/edly-io/sparkth/pull/413

regisb:
Closed in favour of #417 which is more feature-complete and up-to-date with the main branch. This branch served as the proof-of-concept to discuss with the team.

PR #405, opened 2026-06-04
refactor(rag): create public API for RAG

### LINKED ISSUE #398: RAG module lacks a stable public API, forcing plugins to import implementation internals
## Summary

`app/rag/` is a multi-layered pipeline (extraction → chunking → storage → retrieval) but it currently has no stable public API boundary. Plugins reach directly into its implementation internals — `models`, `store`, `chunking`, `extraction`, `utils` — instead of calling a well-defined interface. This makes every internal detail an implicit contract that blocks refactoring and makes third-party plugin development unreliable.

This is a sub-task of #379 (plugins lack a well-defined core API).

---

## What has been done (PR #405)

A stable public API has been established at `app/lib/rag.py`. All plugins now import from there only.

### Public API (`app.lib.rag`)

```python
from app.lib.rag import (
    ingest_document,               # extract → chunk → store+dedup+link (self-contained session)
    agentic_retrieve_context,      # validate-all + concurrent agent retrieval → list[RetrievedChunk]
    IngestionResult,               # dataclass(new_chunks, reused_chunks)
    RetrievedChunk,                # dataclass(source_name, chapter, section, subsection, content)
    RagStatus,                     # enum: QUEUED / PROCESSING / READY / FAILED
    DriveFileNotFoundError,
    RAGNotReadyError,
    RAGRetrievalError,
    ScannedPDFError,
    UnsupportedFileTypeError,
)
```

### What plugins now look like

**Chat** (`app/core_plugins/chat/routes.py`):
```python
from app.lib.rag import (
    DriveFileNotFoundError, RAGNotReadyError, RAGRetrievalError,
    RetrievedChunk, agentic_retrieve_context,
)
# builds prompt blocks from list[RetrievedChunk] directly
```

**Google Drive** (`app/core_plugins/googledrive/utils.py`):
```python
from app.lib.rag import (
    RagStatus, ScannedPDFError, UnsupportedFileTypeError, ingest_document,
)
result = await ingest_document(user_id, file_id, file_bytes, filename)
```

**Slack** (`app/core_plugins/slack/rag.py`):
```python
from app.lib.rag import (
    DriveFileNotFoundError, RAGNotReadyError, RAGRetrievalError,
    RagStatus, RetrievedChunk, agentic_retrieve_context,
)
chunks = await agentic_retrieve_context(user_id, file_ids, question, agent_llm)
```

### Internal restructuring

The `app/rag/` package was reorganised so its internal structure mirrors the public operations:

```
app/rag/
  ingestion/            # extraction → chunking (used by ingest_document)
    __init__.py
    chunking.py         # DocumentChunker
    extraction/         # PDFExtractor, DocxExtractor, HTMLExtractor, TXTExtractor
  retrieval/            # agentic retrieval (used by agentic_retrieve_context)
    __init__.py         # RETRIEVAL_REGISTRY + retrieve_context_from_file dispatcher
    agent.py            # run_agentic_rag_retrieval, get_context_via_agent
  cleanup.py            # cleanup_deleted_files background task
  config.py             # RAGSettings (get_rag_settings)
  constants.py          # structural constants only (no env vars)
  context_service.py    # shared utilities: _lookup_drive_file, format_chunks_as_context, _validate_files_ready
  enums.py              # DocType, RagStatus
  exceptions.py         # ingestion + retrieval exception hierarchy
  models.py             # DocumentChunk, DriveFileChunkLink SQLModel models
  schemas.py            # RAGSearchAgentResponse, SectionRef (Pydantic)
  store.py              # ChunkStoreService + store_and_link_chunks (shared by ingestion + retrieval)
  types.py              # IngestionResult, RetrievedChunk, RAGContext, SimilarityResult, Chunk, …
  utils.py              # resolve_source_name, get_asset (internal helpers)
  assets/               # rag_search_agent_system_prompt.txt
```

Other cleanup done in the same PR:
- `RAGContextService` class dissolved → module-level functions
- `RAGSearchAgent` class dissolved → `run_agentic_rag_retrieval`, `_bind_user_context`
- `chunk_document` wrapper removed → callers use `DocumentChunker().chunk` directly
- `VectorStoreService` renamed → `ChunkStoreService`
- `db_models.py` renamed → `models.py`, `pydantic_models.py` renamed → `schemas.py`
- `RAG_CONCURRENCY` / `RAG_MAX_FILE_SIZE_MB` moved out of `RAGSettings` → `GoogleDriveSettings` (plugin-owned)
- `RAG_ALLOWED_EXTENSIONS` / `parse_rag_allowed_extensions` removed (single source of truth: `SUPPORTED_EXTENSIONS_FOR_EXTRACTION`)
- `RagStatus` migrated to `app.lib.rag` across all client code; import cycle with `app.models.drive` resolved via lazy import in `store.py`
- `rag_context_replacement_prompt.txt` moved from `app/rag/assets/` → `app/core_plugins/chat/assets/`

---

## Remaining (deferred)

- **`app/rag_mcp/`**: still imports `RagStatus`, `DocumentChunk`, and `resolve_source_name` from `app.rag.*` directly — deferred to a separate PR.
- **googledrive duplicate-file path**: `DriveFileChunkLink` still accessed directly in `googledrive/utils.py` for the content-hash deduplication shortcut. A future `link_to_existing_document` public function would remove this residual.

---

## Current plugin imports from `app.rag.*` (residuals only)

```python
# googledrive/utils.py
from app.rag.models import DriveFileChunkLink  # duplicate-file path (see above)
from app.rag.store import ChunkStoreService    # residual; see #398

# slack/routes.py
from app.rag.store import ChunkStoreService    # residual; see #398

# rag_mcp/tools.py (deferred)
from app.rag.enums import RagStatus
from app.rag.models import DocumentChunk
from app.rag.utils import resolve_source_name
```

_This description was updated with the assistance of an LLM (Claude)._

abdulrafey1:
## What
Establishes a clean public boundary for the RAG module: plugins and external code now import only from `app.lib.rag`; the implementation details in `app/rag/` are private. Introduces two public RAG operations — `ingest_document` (extract → chunk → store) and `retrieve_context` (validate + concurrent agent-based retrieval → flat `list[RetrievedChunk]`).

### Public API (`app/lib/rag.py`)
- `refactor(rag)`: move `app/rag/` boundary: rename internal files (`db_models.py` → `models.py`, `pydantic_models.py` → `schemas.py`), extract `RAGSettings` + `get_rag_settings()` into `app/rag/config.py`, clean up re-exports from `types.py`
- `feat(rag)`: add `ingest_document(user_id, owner_file_id, file_bytes, filename) -> IngestionResult` — eligibility check → extract → chunk → store+dedup+link, self-contained session
- `feat(rag)`: add `retrieve_context(user_id, file_ids, query, llm, retrieval_method="agentic") -> list[RetrievedChunk]` — all-or-nothing readiness validation, concurrent per-file agentic retrieval, flat result
- `feat(rag)`: add `RETRIEVAL_REGISTRY` in `app/rag/retrieval/` — new retrieval methods only require a new file + one registry entry

### Internal restructuring (`app/rag/`)
- `refactor(rag)`: dissolve `RAGContextService` class → module-level functions (`get_context_via_agent`, `format_chunks_as_context`, `_lookup_drive_file`)
- `refactor(rag)`: dissolve `RAGSearchAgent` class → `run_agentic_rag_retrieval` + `_bind_user_context`
- `refactor(rag)`: create `app/rag/retrieval/` module; move `agent.py` + `get_context_via_agent` there; `retrieve_context_from_file` dispatches via the registry
- `refactor(rag)`: move env-configurable values out of `constants.py` into `RAGSettings`; `RAG_CONCURRENCY` / `RAG_MAX_FILE_SIZE_MB` moved to `GoogleDriveSettings` (plugin-owned, not RAG)
- `refactor(chat)`: move `rag_context_replacement_prompt.txt` to chat plugin assets; read via `Path` (removes last `app.rag.utils` import from chat)

### Plugin migration
- `refactor(chat)`: `chat/routes.py` calls `retrieve_context`; builds prompt blocks from `RetrievedChunk`
- `refactor(slack)`: `slack/rag.py` calls `retrieve_context`; drops hand-rolled `_run_agent_fan_out`
- `refactor(plugins)`: `googledrive/utils.py` calls `ingest_document`; imports from `app.lib.rag` only (two tracked residuals: `DriveFileChunkLink` duplicate-path + `get_rag_settings` size guard)
- `refactor(plugins)`: `RagStatus` migrated to `app.lib.rag` across all client code; circular import with `app.models.drive` resolved via lazy import in `store.py`

### Tests
- `test(rag)`: move `tests/rag/` → `tests/lib/rag/` → back to `tests/rag/`; strip duplicate fixtures; `tests/lib/test_rag.py` exercises the public boundary (imports only from `app.lib.rag`)

## How to Test
1. `make test.backend.pytest` — 1046 tests pass, 0 failures
2. `make mypy` — no issues in 234 source files
3. `make lint.backend` — all checks pass
4. Confirm `from app.lib.rag import ingest_document, retrieve_context, RagStatus, RetrievedChunk` works
5. Confirm no plugin file imports from `app.rag.*` retrieval internals (only tracked residuals remain)

## Notes
- **No migrations required.** No schema changes.
- **Behaviour changes (intentional):** slack loses partial per-file tolerance → strict (single `retrieve_context` call); chat's per-file `searching_document` stream event → single `searching_documents` event.
- **Deferred residuals (tracked in #398):** googledrive `DriveFileChunkLink` (duplicate-file path) and `rag_mcp` internal imports — separate follow-up PR.


app/rag/config.py
```diff
@@ -0,0 +1,39 @@
...
+    RAG_SCANNED_PDF_MIN_CHARS_PER_PAGE: int = 100
+    RAG_PDF_EXTRACTION_BATCH_SIZE: int = 10
+
+    # RAG_MCP_URL belongs to the rag_mcp server, not the RAG pipeline itself.
+    # It will be moved into rag_mcp's own config in a future refactor (see #398).
+
+
+@lru_cache
+def get_rag_settings() -> RAGSettings:
+    return RAGSettings()
```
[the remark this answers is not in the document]
abdulrafey1: not relevant as of now

regisb:
@abdulrafey1 The goal of the app/lib modules are to be used by other modules and plugins. If you move the entire app/rag/ directory, then it means that every function in there is an API for other modules and plugins --  which defeats the purpose of a library. This is why I mentioned yesterday that we would have to think long and hard about the library APIs that we want to expose for each feature. 

Does that make sense?

abdulrafey1:
> @abdulrafey1 The goal of the app/lib modules are to be used by other modules and plugins. If you move the entire app/rag/ directory, then it means that every function in there is an API for other modules and plugins -- which defeats the purpose of a library. This is why I mentioned yesterday that we would have to think long and hard about the library APIs that we want to expose for each feature.
> 
> Does that make sense?

@regisb Hmm, I'm still a little confused.

1 - Are you suggesting that we keep part of the rag codebase in `app/rag` and a part of it in `app/lib/rag` based on consumption across plugins/modules?

OR

2 - Are you suggesting that we keep all the rag codebase in `app/rag` and write some new public API functions in `app/lib/rag` that the plugins/modules will use?
And for the things such as `RagStatus`, we let it stay in `app/rag` but we re-export it in `app/lib/rag.py` since alot of other plugins use it.

If you suggesting 1, then i don't understand the logic behind keeping parts of the same codebase in different places. Let me know how you want to do this and I'll organize the codebase that way

regisb:
Yes, I do suggest scenario 2, but I do expect that some APIs will be migrated to app/lib as-is (thus scenario 1). In both cases, the common thing is that we have to think about what we want our APIs to be. We need to think hard about it because good APIs don't change too much over time.

app/core_plugins/slack/rag.py
```diff
@@ -217,11 +165,11 @@ async def answer_question(
...
 
     try:
-        contexts = await _run_agent_fan_out(
+        chunks = await retrieve_context(
             user_id=user_id,
             file_ids=file_ids,
-            question=question,
-            agent_llm=agent_llm,
+            query=question,
+            llm=agent_llm,
```
regisb: Don't use keyword arguments when not needed. Write instead:

```
retrieve_context(
            user_id,
            file_ids,
            question,
            agent_llm
)
```

app/core_plugins/googledrive/utils.py
```diff
@@ -322,7 +230,8 @@ async def _process_single_file(
         await session.refresh(drive_file)
         await _set_rag_status(session, drive_file, RagStatus.FAILED, error="Database integrity error")
     except Exception as e:
-        # We catch (then re-raise) all exceptions to make sure the file status is not stale
+        # Catch (then re-raise) all remaining exceptions so the file's rag_status
+        # is never left stale; the re-raise surfaces the unexpected failure upstream.
         logger.error("Unknown error during RAG processing for '%s': %s / %s", log_name, e.__class__, e)
         await _set_rag_status(session, drive_file, RagStatus.FAILED, error="Unknown error")
         raise
```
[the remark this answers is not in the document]
abdulrafey1: this is intentional

app/rag/retrieval/agent.py
```diff
@@ -0,0 +1,212 @@
...
+        raise RAGRetrievalError(f"Agent failed to process query: {exc}") from exc
+
+
+async def get_context_via_agent(
+    session: AsyncSession,
+    user_id: int,
+    file_db_id: int,
+    query: str,
+    llm: Any,
+    limit: int = get_rag_settings().RAG_DEFAULT_CHUNKS,
```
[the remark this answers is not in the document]
abdulrafey1: we will handle this if the said problem ever occurs

app/rag/retrieval/agent.py
```diff
@@ -0,0 +1,212 @@
...
+
+    try:
+        results = await store.fetch_chunks_by_sections(
+            session,
+            user_id,
+            source_name,
+            [s.model_dump() for s in decision.selected_sections],
+            limit,
+        )
+    except Exception as exc:
```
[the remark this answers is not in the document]
abdulrafey1: this is intentional

app/lib/rag.py
```diff
@@ -0,0 +1,134 @@
...
+    "ScannedPDFError",
+    "UnsupportedFileTypeError",
+    "agentic_retrieve_context",
+    "ingest_document",
+]
+
+
+async def ingest_document(
+    user_id: int,
+    owner_file_id: int,
```
regisb: And so this depends on the file being created (with the upcoming Document API) prior to chunking, right?

Do we expect the owner_file_id to become a owner_document_id in the future? If not, we should probably name this function `ingest_file`. If yes, then the name is fine.

app/lib/rag.py
```diff
@@ -0,0 +1,134 @@
...
+    RAGRetrievalError,
+    ScannedPDFError,
+    UnsupportedFileTypeError,
+)
+from app.rag.ingestion.chunking import DocumentChunker
+from app.rag.ingestion.extraction import check_extraction_eligibility, extract_to_markdown
+from app.rag.retrieval import get_context_via_agent_with_isolated_session
+from app.rag.retrieval.utils import validate_files_ready
+from app.rag.store import ChunkStoreService, store_and_link_chunks
+from app.rag.types import IngestionResult, RetrievedChunk  # noqa: F401 — re-exported in __all__
```
regisb, line 17: All these items which are imported are then going to be importable:

Image

1. This is making the API confusing for developers
2. We won't be able to catch unauthorised imports, such as `from app.lib.rag import validate_files_ready`, just by looking at the import statement
3. We are exposing internal functions to end users.

I wonder if we should move the exported modules outside of lib/ and then re-import them here. I'm not sure what the good practices are to make a subset of objects importable. Not sure if `__all__` resolves this.

app/core_plugins/slack/rag.py
```diff
@@ -20,48 +17,43 @@
...
+    RagStatus,
+    RetrievedChunk,
+    agentic_retrieve_context,
 )
-from app.rag.store import SimilarityResult
-from app.rag.types import RAGContext, RagStatus
-from app.rag.utils import resolve_source_name
+from app.llm.providers import BaseChatProvider
+from app.models.drive import DriveFile
+from app.rag.utils import resolve_source_name  # internal helper for source-name resolution (residual; see #398)
```
regisb: Let's not mention PR numbers in the code base...

app/lib/rag.py
abdulrafey1: yes, this will be updated with the coming Document API

app/core_plugins/slack/rag.py
```diff
@@ -176,16 +121,19 @@ async def answer_question(
...
+      2. Retrieve relevant chunks via agentic_retrieve_context (validates all files READY,
+         fans out internally across all file_ids).
+      3. If no chunks are returned, reply with the configured fallback_message.
       4. Optionally synthesize via LLM; otherwise return formatted raw chunks.
 
     Each operational failure mode maps to its own distinct user-facing message and
     ResponseType. config.fallback_message is reserved ONLY for the case where the
     agent ran successfully but found no relevant chunks.
 
+    Note: unlike the previous fan-out implementation, a single per-file agent failure
```
regisb: Let's not refer to past implementations.

app/core_plugins/slack/routes.py
```diff
@@ -54,7 +54,7 @@
 from app.llm.providers import BaseChatProvider, get_provider
 from app.llm.service import LLMConfigService
 from app.models.user import User
-from app.rag.store import ChunkStoreService
+from app.rag.store import ChunkStoreService  # residual; see #398
```
regisb: Same comment as above: let's not refer to other PRs.

app/lib/rag.py
abdulrafey1: hmm, let me go research on how to handle this, `__all__` might a good shout for this.

app/lib/rag.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return IngestionResult(new_chunks=new_count, reused_chunks=reused_count)
+
+
+async def agentic_retrieve_context(
+    user_id: int,
+    file_ids: list[int],
+    query: str,
+    llm: BaseChatModel,
+) -> list[RetrievedChunk]:
```
regisb: In both exposed functions I'm not a fan of the argument order. It's a personal thing, I don't have a fixed rule to order arguments. But I tend to list arguments from most to least relevant, and I try to keep the order consistent between different functions. In this case, I'd write:

```
async def agentic_retrieve_context(
    query: str,
    file_ids: list[int],
    user_id: int,
    llm: BaseChatModel,
) -> list[RetrievedChunk]:
```

This is a personal thing though, so I'm fine with a different approach.

app/lib/rag.py
```diff
@@ -0,0 +1,134 @@
...
+        async with session_scope() as session:
+            async with profile_memory("store_and_link", file=filename, chunks=len(chunks)):
+                new_count, reused_count = await store_and_link_chunks(session, user_id, owner_file_id, chunks, store)
+            await session.commit()
+
+    return IngestionResult(new_chunks=new_count, reused_chunks=reused_count)
+
+
+async def agentic_retrieve_context(
+    user_id: int,
```
regisb: What is this `user_id` argument used for? If I'm not mistaken, it's mostly used to check whether the files are actually owned by this user. But then, why do we need to do that? Can't we run a query across files that are owned by multiple different users? If not, why?

regisb (commented):
Made a few comments, let's discuss the implementation. Let's have a live call if you think this is necessary.

app/lib/rag.py
abdulrafey1: @regisb the imports that are only made for the public API functions (`ingest_document` and `retrieve_context`) could be moved inside the functions, this would completely solve problem.

I've gone through how `__all__` works, it will not solve the problem at hand. 

let me know about your thoughts on the lazy import fix.

app/lib/rag.py
abdulrafey1: the removal of `user_id` check in the rag module will be removed in a future PR. This change is out of scopre for this PR.

app/core_plugins/slack/rag.py
```diff
@@ -19,36 +16,43 @@
...
+    for source in order:
+        lines = [f"[DOCUMENT CONTEXT: {source}]", "The following excerpts were retrieved:", ""]
+        for i, c in enumerate(by_source[source], 1):
+            parts = [p for p in [c.chapter, c.section, c.subsection] if p]
+            label = " / ".join(parts) if parts else "General"
+            lines.append(f"--- Excerpt {i} (Section: {label}) ---")
+            lines.append(c.content.strip())
+            lines.append("")
+        blocks.append("\n".join(lines))
+    return "\n\n".join(blocks)
```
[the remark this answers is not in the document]
abdulrafey1: For now we will add let plugins have their own implementation for this, we can move this to a shared location in another PR

app/rag/retrieval/agent.py
```diff
@@ -0,0 +1,211 @@
...
```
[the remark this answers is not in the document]
abdulrafey1: do not flag this again, this is intentional

PR #407, opened 2026-06-05
chore(dev): simplify runtime and local development setup

regisb:
## What
Simplify the runtime and local-development setup by running the RAG metadata tools and the backend in-process/natively, collapsing the Docker setup to a single production image, and giving the frontend real hot reload in dev.

## Changes
- refactor(rag): run metadata tools in-process instead of mcp http
- chore(dev): run backend natively against dockerized services
- chore(docker): consolidate to a single production Dockerfile
- feat(frontend): hot reload in dev via backend proxy

## How to Test
1. `make backend.install.dev && make frontend.install`
2. `make services` — starts Postgres, Redis, Mailpit (published on 127.0.0.1).
3. `make migrations` — applies migrations natively against the dockerized DB.
4. `make api` — backend boots natively; `curl http://localhost:7727/docs` returns 200, and RAG search still works (metadata tools run in-process, no `rag-mcp` container).
5. `make frontend` — open http://localhost:3000, log in (the `/api` proxy round-trips to the backend on :7727), and edit a component to confirm hot reload.
6. `make test.backend` (pytest + mypy strict + ruff) and `make test.frontend` are green.
7. (optional) `docker build -t sparkth:test .` builds the single production image.

## Notes
- **Breaking dev-workflow change.** `make up` / `make up.dev` / `make base` and the dockerized `app`/`migrations`/`rag-cleanup` services are gone. Dev now runs the backend and frontend natively; docker-compose provides backing services only. `make migrations`/`create-user`/`rag-cleanup` run natively.
- **Ports:** `db` and `redis` now publish to `127.0.0.1:5432` / `127.0.0.1:6379` — conflicts with a pre-existing local Postgres/Redis on those ports.
- **Env:** `.env` is now localhost-first; `RAG_MCP_URL` is removed; `Settings` reads `.env` then `.env.local` (the latter wins; real env vars still win over both).
- **Dependencies:** removed `langchain-mcp-adapters`.
- **Files removed:** `Dockerfile.base`, `Dockerfile.local`, `app/rag_mcp/server.py`, `app/rag_mcp/main.py`.
- **Deployment:** out-of-repo k8s/Helm manifests that ran `rag-mcp` as a separate container or set `RAG_MCP_URL` must be updated; the published production image (built from the unchanged `Dockerfile`) is otherwise unaffected.


regisb:
Let me know if you want me to split this PR in two. Right now it addresses two different things: running RAG in the main runtime and running the frontend in development mode.

hamza-56 (approved):

app/rag/agent_tools.py
```diff
@@ -0,0 +1,78 @@
+"""Build in-process LangChain tools for the RAG search agent.
```
abdulrafey1, line 1: Would it be better if we move the entire `app/rag_mcp` directory to `app/rag/mcp` and have everything live there.

the `main.py` and `server.py` files are already being removed in this PR which only leaves the `schemas.py` and `tools.py`. These are only being used in the `app/rag` module, so maybe its better if we move these files there

docker-compose.yml
```diff
@@ -1,47 +1,11 @@
...
+#
+# In development the backend (make backend.up.dev) and frontend (make frontend.up.dev)
+# run natively on the host; this file only provides the Postgres, Redis, and Mailpit
+# services they talk to. Ports are published on 127.0.0.1 so the native processes can
+# reach them. The application image itself is built only for production (see Dockerfile,
+# published to GHCR by CI and orchestrated by Kubernetes).
 services:
-  # Sparkth App
-  # The main sparkth app container.
-  app:
```
abdulrafey1, line 4: why are we removing this entirely?

Shouldn't we keep this as an option that if someone wants to mimic production k8s, they can run the application in a docker container.

abdulrafey1 (commented):
a few comments

app/rag/agent_tools.py
regisb: I can do that. Would you rather have this change made in this PR or another one?

docker-compose.yml
regisb: In this PR, I propose that we run the app natively on the host in development. So in development it does not make sense to keep this app around.

The thing is that a development and a production docker-compose.yml files are very different. The existing app was not ready to run in production: we would have had to change the image label, pull policy, etc.

I think that at some point we'll have to introduce a production-ready docker-compose.yml, but since we do not have one at the moment, I didn't want to add one in this PR.

app/rag/agent_tools.py
abdulrafey1: Let's do that in a separate PR

PR #409, opened 2026-06-05
refactor(frontend): fix react-doctor size-axes and button-has-type

hamza-56:
## What

Resolve the top two React Doctor warnings in the frontend: redundant `w-N h-N` Tailwind pairs and `<button>` elements missing an explicit `type`.

## Changes

- refactor(frontend): collapse matching `w-N h-N` / `h-N w-N` to `size-N` across 38 files (Tailwind v4 supports `size-*`); pairs with non-matching axes left as-is
- fix(frontend): add `type="button"` to 7 plain `<button>` elements (Alert, FolderPicker x2, DriveFilePicker x2, ChatInput, Pill) so they never default to submit

## How to Test

1. `make lint.frontend` should report 0 warnings.
2. `make lint.frontend.react-doctor` should no longer flag `design-no-redundant-size-axes` or `button-has-type`.
3. `make test.frontend.vitest` should pass (64/64).
4. Visually spot-check icons in the sidebar, chat input, drive picker, and dialogs to confirm sizes are unchanged.

## Notes

None.

### DIFF
--- a/frontend/app/dashboard/[pluginName]/page-client.tsx
+++ b/frontend/app/dashboard/[pluginName]/page-client.tsx
@@ -38,7 +38,7 @@ export default function PluginPageClient() {
         <div className="text-center max-w-md">
           <div className="text-error-600 dark:text-error-400 mb-4">
             <svg
-              className="w-16 h-16 mx-auto"
+              className="size-16 mx-auto"
               fill="none"
               stroke="currentColor"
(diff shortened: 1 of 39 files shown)


abdulrafey1 (approved):
The issue that is fixed, seems to be present in a few more places (as claude has already mentioned). Please fix those and the rest looks good

PR #410, opened 2026-06-05
 refactor(plugins): restructure canvas and openedx plugin internals 

### LINKED ISSUE #383: Refactor canvas and open edX plugins

zaira-bibi:
## What
Restructured canvas and openedx plugin internals to improve separation of concerns and decouple both plugins from `app/mcp`.

## Changes
- `refactor(plugins)`: split `types.py` into `enums.py` and `models.py` in both canvas and openedx plugins
- `refactor(plugins)`: moved exception types (`BaseError`, `AuthenticationError`, `JsonParseError`, `LMSError`) from `app/mcp/types` to `app/lib/exceptions`
- `refactor(plugins)`: moved HTTP utilities (`request`, `handle_error_response`, `Auth`) from `app/mcp/request` to `app/lib/http`
- `refactor(plugins)`: introduced `Method` enum in `app/lib/http`, replacing bare HTTP method strings throughout
- `refactor(plugins)`: improved `CanvasClient.authenticate` — response status accessed inside `async with` block, error extraction made robust against unexpected Canvas response shapes
- `refactor(plugins)`: eliminated double response body reads in `request()` and `handle_error_response()` by using `json.loads(text)` after a single `response.text()` call
- `test(plugins)`: updated canvas and openedx tests to reflect new import paths and mock `response.text` instead of `response.json` where applicable

## How to Test
1. Run `make test.backend.pytest` — all 27 canvas and openedx tests should pass
2. Run `make mypy` — no errors across 233 source files

## Notes
`app/mcp/request.py` is deleted — nothing outside `app/lib` depended on it. Both `app/lib/exceptions.py` and `app/lib/http.py` are new additions to the public library surface; future plugins should import shared HTTP and exception types from `app/lib`.


app/core_plugins/canvas/models.py
```diff
@@ -1,8 +1,17 @@
 from datetime import datetime
-from enum import Enum
 
 from pydantic import BaseModel, Field
 
+from .enums import (
```
abdulrafey1: lets not do a `.enum` import, lets be consistent across the repo and use `from app.core_plugins.canvas.enums import ...`

app/core_plugins/canvas/plugin.py
```diff
@@ -13,11 +13,11 @@
 from typing import Any
 
 from app.core_plugins.canvas.config import CanvasConfig
-from app.mcp.types import AuthenticationError
+from app.lib.exceptions import AuthenticationError
 from app.plugins.base import SparkthPlugin, tool
 
 from .client import CanvasClient
-from .types import (
+from .models import (
```
abdulrafey1: same as above, lets not make relative imports.
Lets fix them across both the plugins

app/core_plugins/openedx/models.py
```diff
@@ -1,8 +1,9 @@
-from enum import Enum
 from typing import Any
 
 from pydantic import BaseModel
 
+from .enums import Component
```
abdulrafey1: same issue

app/lib/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+
+class AuthenticationError(BaseError):
+    pass
+
+
+class JsonParseError(BaseError):
+    pass
+
+
+class LMSError(BaseError):
```
abdulrafey1: Can you please help me in understanding the meaning of this error class.
Where is this needed and why?

This looks like a copy of the `BaseError` class which in itself is redundant. I dont think we need this, or if we do, we need to be specific about what it is, if its encapsulating an umbrella of exceptions, it might be better to just use the `Exception` class instead

app/lib/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def _format_message(self) -> str:
+        return f"{self.message} (status_code={self.status_code})"
+
+
+class AuthenticationError(BaseError):
+    pass
+
+
+class JsonParseError(BaseError):
```
abdulrafey1: this exception class inherits from `BaseError` and base error has a `status_code`. Im not sure if a json parse error will always have a error status code.

I think the `json.JSONDecodeError` class should be used for any json parsing related exception handling

app/lib/exceptions.py
```diff
@@ -0,0 +1,26 @@
+class BaseError(Exception):
```
abdulrafey1: The `BaseError` class feels like a copy of `fastapi.HTTPException` class, we should use exception classes already available in the package.

I dont think we need this, let me know if you think otherwise.

app/lib/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+class AuthenticationError(BaseError):
+    pass
+
+
+class JsonParseError(BaseError):
+    pass
+
+
+class LMSError(BaseError):
+    def __init__(self, method: str, url: str, status_code: int, message: str):
```
hamza-56: Every caller now passes `app.lib.http.Method`, but `method` is still typed as `str` here. Worth tightening to `Method` so the domain type flows end-to-end.

app/lib/http.py
```diff
@@ -39,25 +44,27 @@ async def request(
...
             raise LMSError(method, url, response.status, str(e)) from e
 
-        return json_value
+        if not isinstance(parsed, dict):
+            raise LMSError(method, url, response.status, f"Expected JSON object, got {type(parsed).__name__}")
+        return parsed
 
 
-async def handle_error_response(method: str, url: str, response: ClientResponse) -> LMSError:
+async def handle_error_response(method: Method, url: str, response: ClientResponse) -> LMSError:
```
hamza-56: This pulls the message from a top-level `message` key, but Canvas's `authenticate` reaches in for `errors[0].message`. Could we let callers pass an extractor (e.g. `error_extractor: Callable[[dict], str | None]`) so Canvas can drop its hand-rolled parsing?

app/core_plugins/canvas/client.py
```diff
@@ -19,49 +21,53 @@ async def __aenter__(self) -> CanvasClient:
...
-        _exc_type: Type[BaseException] | None,
+        _exc_type: type[BaseException] | None,
         _exc_val: BaseException | None,
         _exc_tb: TracebackType | None,
     ) -> None:
         await self.close()
 
     @staticmethod
-    async def authenticate(new_api_url: str, new_api_token: str) -> int:
+    async def authenticate(api_url: str, api_token: str) -> int:
```
hamza-56: This is basically a hand-rolled version of `handle_error_response` in `app/lib/http.py`. If we generalize that helper to take an error extractor, this method can just call it and only own raising `AuthenticationError`.

app/core_plugins/canvas/client.py
```diff
@@ -19,49 +21,53 @@ async def __aenter__(self) -> CanvasClient:
...
     async def put(self, endpoint: str, payload: dict[str, Any]) -> dict[str, Any]:
-        return await self.request_bearer("put", endpoint, payload)
+        return await self.request_bearer(Method.PUT, endpoint, payload)
 
     async def delete(self, endpoint: str) -> dict[str, Any]:
-        return await self.request_bearer("delete", endpoint)
+        return await self.request_bearer(Method.DELETE, endpoint)
 
-    async def request_bearer(self, method: str, endpoint: str, payload: dict[str, Any] | None = None) -> dict[str, Any]:
+    async def request_bearer(
```
hamza-56: `request_bearer` here and `OpenEdxClient.request_jwt` are near-identical wrappers over `app.lib.http.request` (just auth scheme + base-URL differ). Could be lifted into a small `BaseHttpClient` in `app/lib/http.py` so the token check and URL-join policy live in one place.

app/lib/http.py
```diff
@@ -39,25 +44,27 @@ async def request(
             return {}
 
         try:
-            json_value: dict[str, Any] = await response.json()
-        except (ContentTypeError, json.JSONDecodeError) as e:
+            parsed = json.loads(text)
+        except json.JSONDecodeError as e:
```
abdulrafey1: the fact that we do not use the `JsonParseError` class defined in `app.lib.exceptions` here is enough of a proof that we do not need this here

app/lib/http.py
abdulrafey1: this check enforces that this generic request method will never return a simple test response, it will only return dicts.

This does not seem like a good implementation for a generic public method.

abdulrafey1 (commented):
Following are some things that I feel are not being done right.
The additions to the public lib module don't seem right. The `http.py` file's request function might be acceptable if its implemented right, right now its not

app/lib/exceptions.py
zaira-bibi: `BaseError` only gives the status code and message, while the `LMSError` adds the `method` and `url` fields and formats the message. Callers use `except LMSError` to catch only LMS HTTP failures (e.g. API call returns a non-2xx status, fails JSON parsing, or returns a non-object etc.), not unrelated exceptions.

I can name it to something more specific so the intent becomes clear.

app/lib/exceptions.py
zaira-bibi: I didn't use those exception classes as they're mostly intended for the api layer and fastapi's handler returns them as HTTP responses to the client. 

whereas `BaseError` is meant to define internal domain errors. it lives in the plugins' client layers and is always caught internally so it never reaches fastapi's exception handler. that's why I would want to keep it. let me know if you disagree.

zaira-bibi:
@claude-review

app/core_plugins/canvas/client.py
```diff
@@ -1,68 +1,47 @@
...
+            if response.status < 200 or response.status >= 300:
+                err = await self._handle_error_response(Method.GET, url, response, error_extractor=_extract)
+                raise AuthenticationError(response.status, err.message)
+            return response.status
 
     async def get(self, endpoint: str) -> dict[str, Any]:
-        return await self.request_bearer("get", endpoint)
+        return await self._request_dict(Method.GET, endpoint)
+
+    async def get_list(self, endpoint: str) -> list[Any]:
```
abdulrafey1: lets name it just `list` instead of `get_list`

app/core_plugins/canvas/schemas.py
```diff
@@ -1,8 +1,17 @@
 from datetime import datetime
```
abdulrafey1, line 1: In other plugins/modules, we are using `schemas.py` for pydantic models and `models.py` for SQLAlchemy models, so `schemas.py` would be a better name for this file.

app/lib/http.py
```diff
@@ -0,0 +1,132 @@
...
+from types import TracebackType
+from typing import Any, Self
+
+from aiohttp import ClientPayloadError, ClientResponse, ClientSession
+
+from app.lib.enums import Auth, Method
+from app.lib.exceptions import AuthenticationError, LMSRequestError
+
+
+class BaseHttpClient:
```
abdulrafey1, line 12: If this class is only for providing HTTP authentication utility, then lets rename it to be more specific, something like `BaseHttpAuthenticationClient`

app/core_plugins/openedx/schemas.py
```diff
@@ -1,8 +1,9 @@
-from enum import Enum
 from typing import Any
```
abdulrafey1, line 1: same comment, `schemas.py` would be a more consistent name for this file

app/lib/exceptions.py
abdulrafey1: Altight, then lets rename it to be more specific, because it is only catering to HTTP errors

zaira-bibi:
@claude-review

abdulrafey1 (approved):
have a look at the pending comments and mark them resolved once done.
Everything looks good

app/core_plugins/canvas/client.py
zaira-bibi: i think `get_all` would be a better name, since `list` is a python builtin. what do you say?

app/lib/http.py
zaira-bibi: it's actually a generic HTTP client base and not specific to authentication.

zaira-bibi:
@claude-review

PR #413, opened 2026-06-08
refactor(plugins): remove dead code from SparkthPlugin

regisb:
## What
Remove dead code from `SparkthPlugin` — members and a constructor param with no callers anywhere in the codebase, plus the unused model registry that never actually drove Alembic table discovery.

## Changes
- refactor(plugins): drop `get_migrations_path`, `get_default_config`, `get_config_value`, and the unused `dependencies` constructor param/attribute
- refactor(plugins): drop the `_models` registry — `add_model()`, `get_models()`, and the now-unused `SQLModel` import
- refactor(plugins): drop `add_model(...)` calls from chat and slack plugins; keep model imports (the actual table-registration side effect) with an explicit `# noqa`
- refactor(migrations): remove the dead `get_models()` logging loop and unused `get_plugin_loader` import from `env.py`; clarify the docstring
- docs(plugins): update `PLUGIN_GUIDE.md` "Adding Database Models" to reflect import-based registration
- refactor(plugins): remove unused class attributes from SparkthPlugin as well as exceptions
- refactor(plugins): simplify plugin loading at app startup time and in migrations.

## How to Test
1. `make mypy` — clean.
2. `make lint.backend` — clean.
3. `make test.backend.pytest` — all pass.
4. Confirm autogenerate still sees plugin tables: in a Python shell, call `get_plugin_loader()` then inspect `SQLModel.metadata.tables` — `chat_conversations`, `chat_messages`, `slack_workspaces`, `slack_bot_response_logs` are present.

## Notes
No migration, no breaking change, no env var, no dependency change. A plugin's `table=True` classes register into `SQLModel.metadata` at module-import time (triggered by the loader), so autogenerate is unaffected by removing the model registry.

### DIFF
--- a/app/migrations/env.py
+++ b/app/migrations/env.py
@@ -10,7 +10,6 @@ from sqlmodel import SQLModel
 from app.core.config import get_settings
 from app.lib.log import get_logger
 from app.models import *  # noqa: F403
-from app.plugins import get_plugin_loader
 from app.services.plugin import get_plugin_service
 
 logger = get_logger(__name__)
@@ -24,7 +23,12 @@ config.set_main_option("sqlalchemy.url", db_url)
 
 # Import plugin models for Alembic autogenerate
 def import_plugin_models():
-    """Import models from all enabled plugins for Alembic to discover."""
+    """Load all enabled plugins so their SQLModel tables register for Alembic.
+
+    ``get_or_create_all`` instantiates the plugin loader, which imports each
+    plugin module; their table classes register into ``SQLModel.metadata`` at
+    import time, making them visible to autogenerate.
+    """
     plugin_service = get_plugin_service()
     try:
         # Create all plugin objects in database
@@ -34,13 +38,6 @@ def import_plugin_models():
             logger.warning("plugins table not yet created — skipping plugin model import")
             return
         raise
-
-    # Get models from each plugin
-    plugin_loader = get_plugin_loader()
-    for plugin_name, plugin in plugin_loader.get_loaded_plugins():
-        models = plugin.get_models()
-        if models:
-            logger.info(f"Loaded {len(models)} model(s) from plugin '{plugin_name}'")
 
 
 # this is the Alembic Config object, which provides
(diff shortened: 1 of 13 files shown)


app/migrations/env.py
```diff
@@ -24,7 +23,12 @@
 
 # Import plugin models for Alembic autogenerate
 def import_plugin_models():
-    """Import models from all enabled plugins for Alembic to discover."""
+    """Load all enabled plugins so their SQLModel tables register for Alembic.
```
hamza-56: Now that the iteration loop is gone, this function basically just calls `get_or_create_all`, which writes plugin rows to the DB. That feels a bit unexpected to find in `env.py`. Worth renaming to something more honest about what it does, or moving the DB writes back to startup so `env.py` only triggers the imports.

app/plugins/base.py
```diff
@@ -106,71 +101,47 @@ def __new__(
...
-        self,
-        name: str,
-        config_schema: Type[PluginConfig] | None = None,
-        is_core: bool = False,
-        version: str = "1.0.0",
-        description: str = "",
-        author: str = "",
-        dependencies: list[str] | None = None,
-    ):
+    def __init__(self, name: str, config_schema: Type[PluginConfig] | None = None):
```
hamza-56, line 134: Heads up for a follow-up: with `dependencies` gone, `PluginDependencyError` in `app/plugins/exceptions.py` has nothing left that can raise it. A few of its siblings (`PluginAlreadyLoadedError`, `PluginNotLoadedError`, `PluginConfigError`, `PluginNotFoundError`) also have zero raise sites in the codebase. Probably worth dropping them in the same cleanup.

app/services/plugin.py
```diff
@@ -147,7 +147,6 @@ async def get_or_create_all(self) -> None:
                     await self.get_or_create(
```
hamza-56: Not introduced by this PR, but worth flagging since this code path is now reached from `env.py` too: we open a new session per plugin and do one SELECT each (plus a possible INSERT/UPDATE). Batching into a single `select(Plugin).where(Plugin.name.in_(names))` plus one commit would be cleaner and atomic. Happy to leave for a follow-up.

abdulrafey1 (approved):

PR #414, opened 2026-06-08
refactor(rag): decouple RAG core from DriveFile — introduce neutral Document registry

### LINKED ISSUE #411: RAG core is coupled to the googledrive DriveFile model and cannot be reused by other plugins
## Summary

The internal RAG module (`app/rag/`) is bidirectionally coupled to the
`googledrive` plugin's `DriveFile` database model. RAG is meant to be a
self-contained core subsystem that any plugin can use through its public API
(`app/lib/rag.py`). Today it instead reaches directly into the `drive_files`
table and the `DriveFile` model, while the model in turn depends on a RAG type.

The practical consequence: a new plugin (e.g. a OneDrive or Dropbox connector)
**cannot use RAG**, because RAG queries `drive_files` and its chunk-link table
foreign-keys into `drive_files.id` — tables that only the `googledrive` plugin
owns. This is a concrete instance of the broader API-perimeter problem tracked
in #379 (`RagStatus` and `DriveFile` both appear in that issue's catalogue of
leaked symbols).

## The dependency cycle

```
app.models.drive  ->  app.lib.rag  ->  app.rag.*  ->  (lazy) app.models.drive
```

The cycle is currently papered over with `TYPE_CHECKING` and function-local
("lazy") imports inside `app/rag/`. Those workarounds hide a real architectural
cycle rather than removing it.

## Where it happens

**Back-edges — RAG reaching into the plugin (these should not exist):**

| File | Line | Coupling |
|---|---|---|
| `app/rag/models.py` | 15–17 | Bridge table `rag_drive_file_chunk_links` foreign-keys `drive_file_id -> drive_files.id`; the table is named after the plugin's concept. |
| `app/rag/store.py` | 194–207 | Lazy-imports `DriveFile`, JOINs `drive_files`, filters on `DriveFile.is_deleted`. |
| `app/rag/cleanup.py` | 10, 30, 52, 78 | Imports `DriveFile`; garbage-collects chunks based on soft-deleted `drive_files` rows. |
| `app/rag/retrieval/utils.py` | 24–51, 82–98 | Queries `DriveFile`; reads `rag_status`, `user_id`, `is_deleted` to gate retrieval. |
| `app/rag/utils.py` | 12, 34–40 | `resolve_source_name(drive_file: DriveFile)` is typed to the plugin model **and** applies Google-native-mime logic (`GOOGLE_NATIVE_MIMES`) inside core RAG. |

**Forward-edge — the plugin model carrying a RAG concern:**

| File | Line | Coupling |
|---|---|---|
| `app/models/drive.py` | 9, 66 | `DriveFile.rag_status: Optional[RagStatus]` — RAG ingestion status is stored on the plugin's own table. |

## The actual contract

Across all five back-edges, the only thing RAG needs from a "source" is six
fields: `id`, `user_id`, `name`, `mime_type`, `rag_status`, and `is_deleted`.
RAG currently obtains them by reaching into `drive_files`. Any other plugin
lacks that table, so RAG breaks for it.

## Why it matters

- Blocks third-party (and even first-party) plugins from reusing RAG — the
  stated goal of a decoupled core.
- Forces plugin-specific knowledge (`GOOGLE_NATIVE_MIMES`) into the core RAG
  module, violating single-responsibility.
- The lazy/`TYPE_CHECKING` imports are fragile and obscure the dependency graph.

## How to observe

Search for the back-references and the cycle workarounds:

```bash
grep -rn "DriveFile\|models.drive\|drive_files" app/rag/
```

Every match in `app/rag/` is a place where core RAG depends on the `googledrive`
plugin.

---

Related: #379 (plugins lack a well-defined core API).

abdulrafey1:
## Closes: [RAG core is coupled to the googledrive DriveFile model and cannot be reused by other plugins](https://github.com/edly-io/sparkth/issues/411)

Closes #411

## What

Introduce a neutral `Document` model in `app/core/documents/` so that `app/rag/*` never imports from any plugin. Plugins manage document identity and status through the core registry; the RAG pipeline depends only on `Document`, not `DriveFile`.

## Changes

- `refactor(rag)`: add `documents` table (`Document` model, `DocumentStatus` enum) and `DocumentChunkLink` bridging `rag_document_chunk_links`
- `migrations`: backfill `document_id` on existing `drive_files` rows; create `documents` table and `rag_document_chunk_links`
- `refactor(rag)`: update retrieval, exceptions, utils, and `lib/rag` to use `Document`
- `refactor(rag)`: `DriveFileChunkLink` renamed to `DocumentChunkLink`; store now writes `document_id`
- `refactor(rag)`: remove `rag_status` and `rag_error` from `DriveFile` model; status now lives on `Document`
- `refactor(plugins)`: update googledrive routes and `rag_mcp` tools to read `DocumentStatus` from the `Document` join
- `refactor(plugins)`: decouple chat and slack RAG lookups from `DriveFile.rag_status`; add `_to_document_ids` helper in chat, JOIN-based filtering in slack
- `migrations`: drop `rag_status`, `rag_error` columns from `drive_files` and drop old `rag_drive_file_chunk_links` table

## How to Test

1. Apply pending migrations: `make migrations`
2. Upload a file via the Google Drive plugin and wait for RAG ingestion to complete
3. `GET /api/v1/googledrive/files/{id}` — confirm `rag_status` is `"ready"` (sourced from `Document.status`)
4. Ask a question in the Chat plugin with the file attached — confirm RAG context is retrieved correctly
5. Confirm `app/rag/` has zero imports from `app.core_plugins.*` or `app.models.drive`:
   ```bash
   grep -r "core_plugins\|models.drive" app/rag/
   ```
   Expected: no output
6. Run the full test suite: `make test.backend.pytest`

## Notes

**Two migrations are required** — apply in order:
1. `e65be7f91ab2` — creates `documents` table, adds `document_id` to `drive_files`, creates `rag_document_chunk_links`, backfills existing rows
2. `696dd7cdaeb3` — drops `rag_status` and `rag_error` from `drive_files`; drops old `rag_drive_file_chunk_links` table

**API backward compatibility**: response field names `rag_status` and `rag_error` are preserved on all existing endpoints — only the source of truth has changed (now `Document.status` / `Document.error`).


app/core_plugins/chat/tests/test_conversation_attachments.py
```diff
@@ -8,9 +8,10 @@
 from sqlalchemy.exc import IntegrityError
 from sqlmodel.ext.asyncio.session import AsyncSession
 
+from app.core.documents.enums import DocumentStatus
+from app.core.documents.models import Document
```
regisb: Do we need to import from app.core in tests or can we instead import from app.lib?

Same question about tests below.

app/lib/documents.py
```diff
@@ -0,0 +1,17 @@
+"""Public API for the Document registry.
+
+All plugins and external modules manage document identity and status from here.
+"""
+
+from app.core.documents.enums import DocumentStatus
+from app.core.documents.models import Document
+from app.core.documents.service import create_document, get_document, soft_delete_document, update_document_status
+
+__all__ = [
```
regisb, line 10: Could we get some clarification about the usage of `__all__`? I'd be tempted to remove them all if they are not useful. (though not in this PR)

app/models/__init__.py
```diff
@@ -1,3 +1,4 @@
+import app.core.documents.models  # noqa: F401 — loads Document table into SQLModel metadata
```
regisb, line 1: Can we avoid adding stuff to this `__init__.py` file by importing from the full model path? Unless I'm mistaken the declarations in this file are not useful so we should avoid adding new stuff there.

A bigger cleanup of `__init__.py` files is out of the scope for this PR, I'm only trying to limit the amount of stuff that we add here.

app/models/drive.py
regisb, line 46: My understanding was that we would be moving this model to the googledrive plugin. Is this not the case?

app/migrations/versions/e65be7f91ab2_add_documents_table_and_document_id_to_.py
```diff
@@ -0,0 +1,123 @@
...
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index("ix_documents_user_id", "documents", ["user_id"])
+    op.create_index("ix_documents_is_deleted", "documents", ["is_deleted"])
+
+    # 2. Backfill: one Document row per DriveFile with a non-null rag_status.
+    # Cast rag_status enum to text so it stores cleanly into documents.status (varchar).
+    op.execute(
+        """
+        INSERT INTO documents (user_id, name, mime_type, status, error, is_deleted,
```
regisb: I find this raw SQL statement a little strange. Do we have a good reason to bypass the ORM here?

app/core_plugins/slack/utils.py
```diff
@@ -0,0 +1,19 @@
+"""Slack plugin utility helpers."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from app.rag import constants
+
+if TYPE_CHECKING:
```
regisb, line 9: Can we avoid `TYPE_CHECKING` at all? I see that there is a dependency on a core model: can we get rid of it entirely? Having two different codebases for typing and production is a sure footgun.

tests/rag/conftest.py
```diff
@@ -1,20 +1,58 @@
...
+from collections.abc import AsyncGenerator, Generator
+from typing import Any
 from unittest.mock import patch
 
 import pytest
+from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
+from sqlalchemy.pool import StaticPool
+from sqlmodel.ext.asyncio.session import AsyncSession
+
+import app.models  # noqa: F401
```
regisb: Why do we need to import this if we don't use the import?

tests/rag/conftest.py
```diff
@@ -1,20 +1,58 @@
...
+async def rag_engine() -> AsyncGenerator[AsyncEngine, None]:
+    """In-memory SQLite engine for RAG tests that need lightweight FK support."""
+    engine = create_async_engine(
+        DATABASE_URL,
+        connect_args={"check_same_thread": False},
+        poolclass=StaticPool,
+    )
+
+    from app.models.base import TimestampedModel  # noqa: F401
+    from app.models.user import User  # noqa: F401
```
regisb: Why the lazy import?

tests/rag/test_store_and_link.py
```diff
@@ -171,48 +166,31 @@ async def test_skips_already_linked_chunks(self) -> None:
 
     @pytest.mark.asyncio
     async def test_deleted_file_chunks_not_reused(self, session: AsyncSession) -> None:
-        """Chunks linked only to deleted DriveFiles must never be reused."""
-        from app.models.drive import DriveFile, DriveFolder
+        """Chunks linked only to deleted Documents must never be reused."""
+        from app.core.documents.enums import DocumentStatus
+        from app.core.documents.models import Document
         from app.models.user import User
```
regisb: Why the lazy imports?

app/lib/documents.py
abdulrafey1: so since these are only re-exported here, mypy flags them as unused. A few options to stop that flag are:

1. A `#noqa` comment
2. `from app.core.documents.models import Document as Document`
3. the `__all__` attribute

option 3 seemed the best to me, hence the above

app/models/drive.py
abdulrafey1: they will be moved to the `googledrive` plugin, but not in this PR, hopefully the next one.

app/core_plugins/slack/utils.py
abdulrafey1: Yes, this is something that is out of the scope of this PR.
The problem here is deeper, it is the `slack` plugin depending on a model which belongs to the `googledrive` plugin.

Once this dependency is resolved (in the next few smaller PRs), this import will no longer be required.

app/lib/documents.py
regisb: Right, let's think about this later then.

app/models/drive.py
```diff
@@ -6,7 +6,6 @@
 from sqlalchemy import UniqueConstraint
 from sqlmodel import Field, Relationship
 
-from app.lib.rag import RagStatus
```
regisb, line 9: For the record: this change alone removes 3 circular dependencies from the codebase:


     app.models.drive -> app.lib.rag -> app.rag.retrieval -> app.rag.retrieval.utils -> app.models.drive
     app.models.drive -> app.lib.rag -> app.rag.retrieval -> app.rag.retrieval.agent -> app.rag.utils -> app.models.drive
     app.models.drive -> app.lib.rag -> app.rag.retrieval -> app.rag.retrieval.agent -> app.rag.store -> app.models.drive

PR #415, opened 2026-06-08
refactor(rag): consolidate RAG MCP tooling under app/rag/mcp

regisb:
## What
Consolidate all RAG MCP tooling into a single `app/rag/mcp/` subpackage so the RAG module is self-contained; pure move + import refactor with no behaviour change.

## Changes
- refactor(rag): move `app/rag_mcp/{__init__,schemas,tools}.py` into `app/rag/mcp/`
- refactor(rag): move `app/rag/agent_tools.py` into `app/rag/mcp/agent_tools.py`
- refactor(rag): relocate tests to the mirrored `tests/rag/mcp/`; make `tests/rag/` a package so the chain resolves as `tests.rag.mcp.*` (avoids a top-level `mcp` on `sys.path` shadowing the installed FastMCP `mcp` package)
- refactor(rag): update imports/patch targets in `app/rag/agent.py`, `app/core_plugins/chat/intent_router.py`, and the affected tests
- docs(core): update the test-layout section in `CLAUDE.md`

## How to Test
1. `make mypy` — passes (239 files, no issues).
2. `make lint.backend` — all checks pass.
3. `make test.backend.pytest` — full suite green (1058 passed); in particular `tests/rag/mcp/test_tools.py`, `tests/rag/mcp/test_agent_tools.py`, and `app/core_plugins/chat/tests/test_intent_router.py`.
4. `grep -rnE "rag_mcp|app\.rag\.agent_tools" app tests` returns nothing.

## Notes
None — no migrations, breaking changes, env vars, or dependency bumps.


tests/rag/mcp/test_tools.py
```diff
@@ -7,15 +7,15 @@
...
-from app.rag_mcp.tools import (
+from app.rag.mcp.tools import (
     get_chunk_stats,
     get_document_structure,
     get_file_metadata,
     list_file_sections,
     list_user_files,
     search_section_by_keyword,
 )
+from app.rag.types import RagStatus
```
abdulrafey1, line 18: this is weird, `RagStatus` lives in `app.rag.enums` not `app.rag.types`, i have double checked on the main branch as well

app/rag/mcp/tools.py
```diff
@@ -8,9 +8,9 @@
 from app.lib.log import get_logger
 from app.models.drive import DriveFile
 from app.rag.db_models import DocumentChunk
+from app.rag.mcp.schemas import ChunkStats, DocumentSection, FileInfo, FileMetadata, SectionKey
 from app.rag.types import RagStatus
```
abdulrafey1, line 12: `RagStatus` lives in `app.rag.enums` not `app.rag.types`

abdulrafey1 (commented):
other than the 2 imports, everything looks fine

tests/rag/mcp/test_tools.py
regisb: Yeah but app.rag.types re-exports these types "for convenience". Hey you made [that change](https://github.com/edly-io/sparkth/pull/352/changes#diff-b9b36a29915caa0141cc9bf2c6fa80f84b15756de7a90460ea1e85d23c59b2eb)! (and I approved it :wink:) The assumption being that consumers of this API should import from "types", not "enums". That way, if we ever decide to change the type of RagStatus from an enum to something else, we don't have to change all the imports.

PR #417, opened 2026-06-08
refactor(plugins): unify plugin contributions on a hook primitive

regisb:
## Relates to: [Plugin API is inconsistent, closed to extension, and disconnected from the frontend](https://github.com/edly-io/sparkth/issues/372)

Addresses problems 1–3 of the issue — the missing unifying extension primitive. Problem 4 (linking the backend plugin to its frontend counterpart) is out of scope for this PR.

## What
Introduce a single `PluginHook` extension primitive and migrate plugin routes, MCP tools, and config schemas onto it, so a plugin contributes everything uniformly from its `__init__` instead of via bespoke base-class accessors and side-channel dicts.

## Changes
- feat(plugins): add the `PluginHook` / `PluginCollectionHook` primitive
- refactor(routes): migrate plugin routes to the `ROUTES` hook
- refactor(mcp): migrate plugin MCP tools to the `MCP_TOOLS` hook
- refactor(config): migrate plugin config schemas to the `CONFIG_SCHEMAS` hook
- refactor(mcp): register tools on `MCP_TOOLS` and drop the `@tool` decorator, `PluginMeta` metaclass, and `_tool_registry` — resolving the `app.lib` ↔ `app.plugins` circular import (no more `TYPE_CHECKING` workaround)
- test(openedx): drop a vestigial fixture that leaked a `MagicMock` into the global `CONFIG_SCHEMAS` hook and broke the lms-credentials tests under non-alphabetical test order

## How to Test
1. `uv run python -c "import app.main"` — succeeds with no import cycle.
2. `make test.backend` — lint + mypy + pytest (1072 passed) + format, all green.
3. `make mcp` — the server boots and logs the plugin tools registering (Canvas 29 + Open edX 10).
4. `make backend.up.dev` — plugin routes are mounted (e.g. chat under `/api/v1`, Slack under `/api/v1/slack`).

## Notes
- **Breaking for out-of-tree plugins:** the `@tool` decorator, `PluginMeta`, and `_tool_registry` are removed. Plugins now register tools with `MCP_TOOLS.add_item(self, Tool(handler, category=...))` in `__init__`, and a tool's description comes from its handler docstring (no separate `description` argument). See [`app/plugins/PLUGIN_GUIDE.md`](app/plugins/PLUGIN_GUIDE.md).
- No database migration, no new environment variables, no dependency changes.


app/lib/mcp/hooks.py
```diff
@@ -0,0 +1,145 @@
...
+    def description(self) -> str:
+        return (self.handler.__doc__ or "").strip()
+
+    @property
+    def input_schema(self) -> dict[str, Any]:
+        return generate_input_schema(self.handler)
+
+
+# MCP tools contributed by plugins, consumed by the MCP server (app/mcp/main.py)
+# and the chat tool registry (app/core_plugins/chat/tools.py).
```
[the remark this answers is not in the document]
regisb: I'm concerned that we are calling get_plugin_loader() in too many places and that we should get rid of those calls. Otherwise AI agents are going to pick on this pattern and replicate it in many places. Let's investigate.

app/mcp/main.py
```diff
@@ -117,45 +44,36 @@ def register_plugin_tools() -> None:
         logger.error(f"Failed to initialize plugin system for MCP: {e}")
```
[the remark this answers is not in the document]
regisb: This is out of the scope of this PR.

app/mcp/main.py
```diff
@@ -117,45 +44,36 @@ def register_plugin_tools() -> None:
...
-        registered_tools[validated_tool.name] = plugin_name
+        registered_tools[tool.name] = plugin_name
 
-        category_str = f" [{validated_tool.category}]" if validated_tool.category else ""
-        version_str = f" v{validated_tool.version}" if validated_tool.version != "1.0.0" else ""
-        logger.info(
-            f"  ✓ Registered tool '{validated_tool.name}'{category_str}{version_str} from plugin '{plugin_name}'"
-        )
+        category_str = f" [{tool.category}]" if tool.category else ""
+        logger.info(f"  ✓ Registered tool '{tool.name}'{category_str} from plugin '{plugin_name}'")
```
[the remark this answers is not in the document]
regisb: This is out of scope, let's ignore this.

app/lib/hooks.py
```diff
@@ -0,0 +1,46 @@
+import weakref
+from typing import Generic, Iterator, TypeVar
+
+from app.plugins.base import SparkthPlugin
```
hamza-56, line 4: This makes `app.lib` depend on `app.plugins`, which is the opposite of the documented direction (plugins import from lib). The hooks only need `.name`, so a small `Protocol` here would keep lib free of plugin imports and avoid reintroducing the cycle this PR removes.

*Generated with AI assistance (Claude).*

app/lib/config/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    lifespan, the standalone MCP server, or the migration runner), which is what
+    populates CONFIG_SCHEMAS; this iterator assumes that has already happened.
+    """
+    for plugin, config_schema in CONFIG_SCHEMAS.iter_items():
+        yield plugin.name, config_schema
+
+
+def get_plugin_config_schema(plugin_name: str) -> type[PluginConfig] | None:
+    """Return the config class a plugin contributed, looked up by plugin name."""
+    for name, config_schema in iter_plugin_config_schemas():
```
hamza-56, line 20: This is an O(n) scan per lookup, and `get_or_create_all` calls it once per plugin, so that path ends up O(n^2). Harmless at the current plugin count, but a name-indexed dict would make it O(1).

*Generated with AI assistance (Claude).*

app/lib/mcp/hooks.py
```diff
@@ -0,0 +1,145 @@
...
+    except (TypeError, NameError, AttributeError) as e:
+        logger.warning("Failed to generate input schema for %s: %s. Using empty schema.", func.__name__, e)
+        return {"type": "object", "properties": {}}
+
+
+def type_to_json_schema(py_type: type[Any]) -> dict[str, Any]:
+    """Convert a Python type to a JSON Schema type definition."""
+    # Check if it's a Pydantic BaseModel
+    try:
+        from pydantic import BaseModel
```
hamza-56: pydantic is a hard dependency of the project, so this import can live at the top of the file. That also makes the `ImportError` arm below dead code.

*Generated with AI assistance (Claude).*

tests/lib/mcp/test_tool.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any
+
+from app.lib.mcp.hooks import Tool, generate_input_schema
```
hamza-56, line 3: `app/lib` is core, not a plugin, so per the CLAUDE.md test layout these belong at `tests/lib/mcp/` (same rule as `app/rag/mcp/` -> `tests/rag/mcp/`). This branch already follows that for `tests/lib/test_hooks.py`, so moving this file and `test_schema_resolution.py` would keep things consistent.

*Generated with AI assistance (Claude).*

app/core_plugins/canvas/plugin.py
```diff
@@ -50,239 +53,304 @@ class CanvasPlugin(SparkthPlugin):
...
+                [
+                    canvas_list_questions,
+                    canvas_get_question,
+                    canvas_create_question,
+                    canvas_update_question,
+                    canvas_delete_question,
+                ],
+            ),
+        ]
+        for category, handlers in tools_per_category:
```
hamza-56, line 94: This registration loop is duplicated verbatim in `openedx/plugin.py`. Might be worth a small shared helper (e.g. in `app/lib/mcp/hooks.py`) so future plugins have one obvious pattern to follow.

*Generated with AI assistance (Claude).*

app/plugins/base.py
```diff
@@ -1,455 +1,49 @@
 """
```
hamza-56, line 1: Now that `@tool` is gone, CLAUDE.md line 27 still says `base.py (SparkthPlugin, @tool), loader.py`. Worth updating it to mention the hook primitives instead.

*Generated with AI assistance (Claude).*

app/core_plugins/chat/lms_credentials.py
```diff
@@ -3,12 +3,18 @@
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.core_plugins.chat.constants import LMS_RULES
+from app.lib.config import iter_plugin_config_schemas  # lazy import — avoids circular dep
```
abdulrafey1: this does not look like a lazy import, the comment is misleading

app/core_plugins/chat/lms_credentials.py
```diff
@@ -3,12 +3,18 @@
...
 
 
 def _lms_tool_prefixes() -> tuple[str, ...]:
-    from app.plugins import PLUGIN_CONFIG_CLASSES  # lazy import — avoids circular dep
+    """
+    Derive LMS tool-name prefixes from all registered plugin configs.
+
+    Any config class that overrides ``lms_tool_prefix`` contributes its prefix.
+    Lazy-imported to avoid circular dependencies at module load time.
```
abdulrafey1: the last line of the doc-string is also irrelevant now

app/core_plugins/chat/plugin.py
```diff
@@ -4,24 +4,16 @@
...
-
-        if not self.config_schema:
-            raise ValueError("ChatUserConfig is required")
-
-        logger.info("Chat plugin configuration validated")
-
-    def get_route_prefix(self) -> str:
-        return "/api/v1"
+        super().__init__(plugin_name)
+        ROUTES.add_item(self, (chat_router, "/api/v1", ["chat"]))
```
abdulrafey1: are we not using the `add_router` lib method we talked about in the meeting last week? or are we planning to shift to it in a separate PR?

app/core_plugins/canvas/plugin.py
```diff
@@ -50,239 +53,304 @@ class CanvasPlugin(SparkthPlugin):
...
+            MCP_TOOLS.add_items(self, [Tool(handler, category=category) for handler in handlers])
+
+
+async def canvas_authenticate(auth: AuthenticationPayload) -> dict[str, Any]:
```
abdulrafey1: we've added a new `tools.py` in the openedx plugin for the tools definitions, is there a reason we haven't done the same here?
If not then lets move these tool definitions to a `tools.py` for canvas.

app/lib/mcp/tests/test_schema_resolution.py
```diff
@@ -1,9 +1,8 @@
 from typing import Any
 
-import pytest
 from pydantic import BaseModel
 
-from app.plugins.base import SparkthPlugin
+from app.lib.mcp.hooks import resolve_schema_refs, type_to_json_schema
 
 
 # ---------------------------------------------------------------------------
```
abdulrafey1: lets remove all comments like these, im sure all of them have been added by Claude in previous iterations

app/lib/mcp/tests/__init__.py
abdulrafey1, line 1: shouldn't these tests be at `tests/lib/mcp`

app/lib/mcp/hooks.py
```diff
@@ -0,0 +1,145 @@
...
+from typing import Any, get_type_hints
+
+from app.lib.hooks import PluginCollectionHook
+from app.lib.log import get_logger
+
+logger = get_logger(__name__)
+
+
+@dataclass
+class Tool:
```
abdulrafey1, line 15: can we add output schema here too? 
I know Most of the tools infer output schemas from the type hint, but they also infer input schemas form the signature and input schema is added here, i think we can add the output schema here as well. WDYT?

app/lib/mcp/hooks.py
abdulrafey1: lets fix the lazy import

abdulrafey1 (commented):
I believe the branch needs rebasing, the diff from main might be a little off because of it. Nonetheless following are a few comments

app/lib/hooks.py
regisb: `app.plugins` is not a plugin, it's the part of the core library that is related to plugin management. In that sense, it's OK to import it.

app/lib/config/__init__.py
regisb: Good catch. This is totally avoidable since PluginHook objects store plugin objects in a dict-like structure. We should add a `BasePluginHook.get(plugin, default=None)` method and convert the first argument of `get_plugin_config_schema` to a SparkthPlugin object.

The catch is that we would need a function to easily switch from app.models.plugin.Plugin objects to SparkthPlugin. 

But at this point, SparkthPlugin are nothing more than objects with a name, so we should be able to get rid of them entirely and load plugins either at import time or by calling a simple function.

This points to a bigger flaw, which is the plugin loading mechanism. I had this in the back of my mind, and I want to take a look at this. But I'd like to keep this for a later PR. I'll open an issue.

app/lib/mcp/tests/__init__.py
regisb: yup, absolutely.

app/lib/mcp/hooks.py
regisb: I don't understand your suggestion because I don't know what is an output schema. Do you mean the pydantic schema of the tool output? Is it used anywhere? If not I'd rather not introduce a new, unused feature.

app/lib/mcp/hooks.py
regisb: Yes. Actually, I'd like to also get rid of the TypeError catch which is not covered by any unit test.

app/core_plugins/canvas/plugin.py
regisb: I moved the Open edX tools to a separate module to resolve a circular dependency issue, which wasn't happening here. But yeah, sure, I can do the same here.

app/core_plugins/canvas/plugin.py
regisb: I'm fine with duplicating this loop. It's not very verbose and it's very explicit.

app/core_plugins/chat/plugin.py
regisb: At the moment I feel like an `add_router` would add very little value. It would look like this:

```
def add_router(router, prefix, tags):
    ROUTES.add_item(self, (chat_router, "/api/v1", ["chat"]))
```

So it's one level of indirection with no clear benefit. If we start defining the route prefix and tags automatically, then it will make sense to add a function to the lib (and to remove the hook from the lib).

Do you agree?

app/lib/mcp/hooks.py
abdulrafey1: MCP tools can also have [output schemas](https://modelcontextprotocol.io/specification/draft/server/tools#output-schema), this helps the LLM know the structure of what they return. This helps LLMs in chaining the output of one tool to the input of another

> Providing an output schema helps clients and LLMs understand and properly handle structured tool outputs by:
Enabling strict schema validation of responses
Providing type information for better integration with programming languages
Guiding clients and LLMs to properly parse and utilize the returned data
Supporting better documentation and developer experience

app/core_plugins/chat/plugin.py
abdulrafey1: > if we start defining the route prefix and tags automatically

I think we should do this. Let me know if you have a reason to defer this implementation.

app/lib/mcp/hooks.py
regisb: Right, but right now we are not using output schemas, right?

app/core_plugins/chat/plugin.py
regisb: OK I ~have~ had one reason. The chat plugin uses the "api/v1" prefix. But the slack and googledrive plugins use "/api/v1/slack" and "/api/v1/googledrive". If we settle on "/api/v1/<plugin-name>", isn't this going to break some routes?

... just had a look at the chat router, and it uses a "/chat" prefix, so all chat routes are effectively prefixed with "/api/v1/chat". So I'm OK to implement this change :+1:

app/lib/mcp/hooks.py
abdulrafey1: On line 29 in `app/mcp/main.py` in the `register_plugin_tool` function we register tools for plugins, via the `mcp.tool` method (`FastMCP`) that comes from `app.mcp.server`

Following is the signature of the `FastMCP.tool` method:
```
def tool(
    self: FastMCP[Any],
    name_or_fn: AnyFunction,
    *,
    name: str | None = None,
    title: str | None = None,
    description: str | None = None,
    icons: list[Icon] | None = None,
    tags: set[str] | None = None,
    output_schema: dict[str, Any] | NotSetT | None = NotSet,
    annotations: ToolAnnotations | dict[str, Any] | None = None,
    exclude_args: list[str] | None = None,
    meta: dict[str, Any] | None = None,
    enabled: bool | None = None
) -> FunctionTool:
```

`output_schema` is part of the signature, if we pass this output schema to the tool, it would help the LLMs understand these tools better. In short, this will reduce the errors in LLMs creating workflows with these tools, where ever they are used.

**Different Topic**:
With that being said, while going through the codebase to write this message, I noticed that the `chat` plugin is converting all tools registered with the `MCP_TOOLS` hook to `langchain` tools (and this is the only place where `app.lib.mcp.hooks.Tool.input_schema` is being used). I don't quite understand the reason why the chat plugin is doing this.

@zaira-bibi is this being done for our course generation agent? If so, can we move our course generation agent into a separate module, it can live inside the chat plugin but with better organization. And we might need to double-check which tools do we want to use for this agent, using tools for all plugins doesn't seem right

app/core_plugins/chat/plugin.py
regisb: Done, please review the latest commit.

app/lib/mcp/hooks.py
regisb: Output schema will be addressed in a later issue: https://github.com/edly-io/sparkth/issues/436

abdulrafey1 (approved):
looks good

regisb:
I'm honestly very tired of rebasing this so I'll merge this now if you don't mind.

PR #419, opened 2026-06-09
refactor(mcp): drop standalone entrypoint, move registration to server

regisb:
## What

Remove the standalone MCP entrypoint (`python -m app.mcp.main`), which was unused
in production, and move plugin tool registration next to the FastMCP instance. The
MCP server only ever runs embedded in the FastAPI app, mounted at `/ai/mcp`.

## Changes

- refactor(mcp): drop the standalone entrypoint (`main`/`run_stdio`/`run_http`) and the now-dead `TransportMode` (`app/mcp/mode.py`)
- refactor(mcp): move `MCPToolDefinition`/`register_plugin_tools`/`_validate_and_register_tool` from `app/mcp/main.py` into `app/mcp/server.py`, next to the FastMCP instance
- docs(mcp): update `Makefile` (drop `make mcp`), `README.md`, `CLAUDE.md`, and architectural docs — the server now runs only embedded at `/ai/mcp`; Claude Desktop bridges to that HTTP endpoint via `mcp-remote`

## How to Test

1. `make mypy` — passes (245 source files, no issues).
2. `make lint.backend` — ruff clean.
3. `make test.backend.pytest` — 1049 passing.
4. Start the backend (`make backend.up.dev`) and confirm plugin MCP tools register on startup (log: "MCP server ready with N total tool(s) registered") and are reachable at `http://127.0.0.1:7727/ai/mcp`.

## Notes

- **Breaking (dev/integration only):** the standalone MCP server (`make mcp` /
  `python -m app.mcp.main`, including `--transport=stdio`) is removed. The MCP
  server now runs only embedded in the FastAPI app at `/ai/mcp`. The README's
  Claude Desktop setup is updated to bridge to that HTTP endpoint via
  `mcp-remote`.
- No migrations, no new env vars, no dependency changes.


regisb:
@claude-review

app/mcp/server.py
```diff
@@ -27,3 +34,150 @@ async def get_course_generation_prompt_tool(
...
+        registered_tools: Dictionary tracking already registered tools
+
+    Returns:
+        True if tool was successfully registered, False otherwise
+    """
+    try:
+        validated_tool = MCPToolDefinition(**tool_def)
+    except ValidationError as e:
+        logger.warning(f"Invalid tool definition from plugin '{plugin_name}': {e}")
+        return False
```
zaira-bibi: this needs to re-raise the `ValidationError`, so the except block at line 119 can catch it.

app/mcp/server.py
```diff
@@ -27,3 +34,150 @@ async def get_course_generation_prompt_tool(
...
+    4. Checks for naming conflicts
+    5. Registers tools with the FastMCP server
+
+    Note: Assumes plugins are already loaded by the plugin lifespan manager.
+    """
+    try:
+        plugin_loader = get_plugin_loader()
+
+        loaded_plugins = plugin_loader.get_loaded_plugins()
+        loaded_plugin_names = [name for name, _plugin in loaded_plugins]
```
zaira-bibi: `loaded_plugin_names` should be moved after the `if not loaded_plugins` guard.

app/mcp/server.py
```diff
@@ -27,3 +34,150 @@ async def get_course_generation_prompt_tool(
...
+                if plugin_tool_count > 0:
+                    logger.info(
+                        f"✓ Plugin '{plugin_name}' registered {plugin_tool_count} tool(s)"
+                        + (f" ({plugin_failed_count} failed)" if plugin_failed_count > 0 else "")
+                    )
+                elif plugin_failed_count > 0:
+                    logger.warning(f"Plugin '{plugin_name}' failed to register {plugin_failed_count} tool(s)")
+
+            except (AttributeError, RuntimeError) as e:
+                logger.error(f"Failed to process MCP tools from plugin '{plugin_name}': {e}")
```
zaira-bibi: i think we should have a counter for this case as well, so the summary log below gives the complete picture.

zaira-bibi (changes requested):

app/mcp/server.py
regisb: I feel like we should just let the application crash whenever an MCP tool fails to load. Was there ever a case when a tool failed to load and you wanted the app to keep running?

app/mcp/server.py
regisb: I just migrated the code from the previous implementation, without modification. Here again I think we should get rid of the excessive exception catching. What do you think?

app/mcp/server.py
zaira-bibi: makes sense

zaira-bibi (approved):
on board with dropping the extra exception handling. rest looks good!

app/mcp/server.py
regisb: OK I'll do it.

app/mcp/server.py
regisb: OK I'll do it.

Issue #403, opened 2026-06-04
Frontend duplicates the backend API contract instead of deriving it


regisb:
## Summary

The frontend has no single source of truth for the backend's REST API contract.
Backend endpoint paths (and the request/response shapes that go with them) are
hardcoded as string literals and hand-written TypeScript interfaces, duplicated
across many files. Because nothing ties these to the backend's actual routes,
any change to a backend URL, request body, or response shape must be manually
mirrored in the frontend — and when it isn't, the mismatch is invisible until it
fails at runtime (a 404 or a malformed payload) rather than at build time.

## Where it happens

Backend paths (`/api/v1/...`) are referenced as raw strings in **13 source
files** (26 occurrences):

- Centralised-ish API modules: `frontend/lib/api.ts`, `frontend/lib/llm-api.ts`,
  `frontend/lib/slack-api.ts`, `frontend/lib/drive.ts`,
  `frontend/lib/file_upload.ts`, `frontend/lib/plugins/context.tsx`
- Inline `fetch()` calls scattered through the chat plugin:
  `frontend/plugins/chat/hooks/useChatStream.ts`,
  `frontend/plugins/chat/hooks/useConversation.ts`,
  `frontend/plugins/chat/hooks/useConversations.ts`,
  `frontend/plugins/chat/hooks/useChatInput.ts`,
  `frontend/plugins/chat/hooks/useCatalogDefaults.ts`,
  `frontend/plugins/chat/components/ChatInterfaceInner.tsx`,
  `frontend/lib/auth-context.tsx`

Representative examples:

```ts
// frontend/lib/api.ts:119
response = await fetch(`${API_BASE_URL}/api/v1/auth/login`, { ... });

// frontend/plugins/chat/hooks/useChatStream.ts (inline, no shared module)
const res = await fetch("/api/v1/chat/completions", { ... });
```

The request/response types for these calls are likewise hand-maintained
TypeScript interfaces in each module, with no link to the Pydantic/SQLModel
schemas they mirror on the backend.

## Why it matters

This is the classic split-brain between a decoupled frontend and backend: the
two share a contract, but the contract lives in two places and is enforced in
neither.

Compared to established good practice for a typed frontend talking to an
OpenAPI-capable backend (FastAPI publishes a complete schema at
`/openapi.json`):

- **Single source of truth.** The contract should be derived from the backend,
  not re-typed by hand on the client. Today it is duplicated.
- **Drift caught at build, not in production.** When a route or payload changes,
[...]

PR #427, opened 2026-06-10
refactor(frontend): centralise inline api fetches into lib modules

hamza-56:
## What

First PR for #403: backend paths were fetched inline from the chat hooks, ChatInterfaceInner, and auth-context, so a single endpoint change meant a multi-file hunt. Every frontend API path now lives in a lib/* module, which the upcoming typed-client migration will convert one module at a time.

## Changes

- refactor(frontend): add lib/chat-api module for chat endpoints (+ tests)
- refactor(frontend): support abort signals in fetchProviderCatalog (+ tests)
- refactor(frontend): add getCurrentUser to lib/api (+ tests)
- refactor(frontend): route auth-context user fetch through lib/api
- refactor(frontend): move chat plugin fetches into lib modules
- test(frontend): pin chat-api auth header and abort-signal forwarding
- merge main: adopted the #425 attachments decoupling, so chat-api speaks document semantics (attachDocument/detachDocument, document_id, document_ids)

## How to Test

1. `make test.frontend` (vitest: 10 files / 84 tests, includes the new lib/chat-api, lib/llm-api, and lib/api specs)
2. `make frontend.build` (acts as the TypeScript gate until #403's PR-3 adds tsc to CI)
3. In the app: log in, send a chat message (streams), reload a conversation, attach/detach a Drive file. Note: I verified via the unit suite, build, and grep (no inline /api/v1 paths remain in converted files), but did not run the manual browser flow; please exercise step 3 during review.

## Notes

- No new dependencies, no backend changes, no migrations.
- Behavior-preserving except two logging-only changes: non-401 /user/me failures now console.error (previously silent), and non-ok provider-catalog responses now console.warn (previously a silent return).
- lib/chat-api deliberately throws plain Error to preserve the original inline call-site behavior, while lib/api and lib/llm-api use ApiRequestError; error-shape normalisation is deferred to the typed-client PRs of #403.
- The pre-login "Bearer null" quirk is intentionally preserved, documented in lib/chat-api, and pinned by a test.
- main was merged in after #425 landed (it reworked the same chat hooks); the conflict resolution reapplies the centralisation on top of the documentId-keyed logic.


frontend/lib/tests/api.test.ts
```diff
@@ -0,0 +1,40 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
```
abdulrafey1, line 1: I have some suggestions on the organization of the codebase.

1. I can see that almost all api files house multiple things, such as interfaces and endpoint handlers. I propose, instead of `frontend/lib/chat-api.ts`, we create a `frontend/lib/chat` directory and inside we create multiple files like `interfaces.ts`, `routes.ts` (filenames can be better).
2. Can we move all `*.test.ts` files to a `frontend/lib/tests` directory, instead of having the apis and their tests together.

Let me know if the current organization is a requirement of next.js, then we can ignore this

hamza-56: Good call, did both. Split `chat-api.ts` into `lib/chat/{types,client,index}.ts` with a barrel re-export so callers still import from `@/lib/chat`, and moved every `*.test.ts` into `lib/tests/` with absolute `@/lib/...` paths. Not a Next.js constraint, just inertia. Landed in 35de3eb.

frontend/lib/llm/client.ts
abdulrafey1, line 1: are we not breaking `llm-api.ts` into multiple files? or are we planning to do this in a later PR?

hamza-56: Done in dd8438e, same `{types,client,index}` split as `lib/chat`. `drive.ts`, `slack-api.ts`, and `lib/api.ts` are still single-file. 

frontend/lib/chat/client.ts
```diff
@@ -0,0 +1,150 @@
...
+  similarity_threshold: number;
+  conversation_id?: string;
+  document_ids?: number[];
+}
+
+// Token may be null pre-login; the backend rejects "Bearer null" with 401,
+// which callers already handle. Matches the previous inline behavior exactly.
+function authHeaders(token: string | null): HeadersInit {
+  return { Authorization: `Bearer ${token}` };
+}
```
zaira-bibi, line 21: could a TODO be added here noting that `authHeaders` should eventually require a non-null token? the above comment explains the legacy reason, but it might be easy for future callers to miss that passing null produces a silently malformed header.

frontend/lib/chat-api.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+export interface OutgoingChatMessage {
+  role: string;
+  content: string | object[];
+  attachment?: { name: string; size: number };
+}
+
+export interface ChatCompletionRequestBody {
+  llm_config_id: number | undefined;
```
zaira-bibi: this should probably be `llm_config_id?: number`. the rest of the optional fields in this interface use the `?` style, so this one seems inconsistent.

frontend/lib/chat-api.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export interface PersistedAttachment {
+  id: number;
+  name: string;
+  size: number | null;
+}
+
+export interface OutgoingChatMessage {
+  role: string;
+  content: string | object[];
```
zaira-bibi: `object[]` is considered a weak type in TypeScript as it allows arbitrary properties to be accessed without type checking. 

might be a good place to narrow the array type. if it carries typed content parts (e.g. image or text blocks), it should be typed accordingly. if the shape is genuinely unknown, `unknown[]` is safer because it forces callers to narrow before use.

frontend/lib/chat-api.ts
```diff
@@ -0,0 +1,150 @@
...
+export async function getConversation(
+  token: string | null,
+  conversationId: string,
+  signal?: AbortSignal,
+): Promise<ApiConversation> {
+  const r = await fetch(`${API_BASE}/conversations/${conversationId}`, {
+    headers: authHeaders(token),
+    signal,
+  });
+  if (!r.ok) throw new Error(`Load conversation failed with status ${r.status}`);
```
zaira-bibi: pre-existing: the `if (!r.ok)` throw caters to HTTP errors, but there's no handling for network errors which may come up as raw `TypeError`. 

now that this is a named lib function, worth wrapping the `fetch` in a try/catch. this applies to all the functions in the file that call `fetch` directly

zaira-bibi (changes requested):
the two lib files (`api.ts` and `chat-api.ts`) have different error contracts:

- `getCurrentUser` in `api.ts` throws `ApiRequestError` with structured field errors and a status code, consistent with the rest of that file. 
- all the functions in `chat-api.ts` throw plain `new Error(...)` 

not something to fix in this PR, but worth a follow-up or at least a comment in `chat-api.ts` noting the deliberate difference.

regisb (commented):
I think I'm not the right person to review this...

hamza-56:
@claude-review

abdulrafey1 (approved):

PR #428, opened 2026-06-10
refactor(core): extract db-free assemble_app factory

hamza-56:
## What

Second PR for #403: plugin routes were registered inside the lifespan after a DB call, so the OpenAPI schema could not be derived offline (a bare import saw only the core paths). Route registration now lives in a sync, DB-free assemble_app() factory that runs at import time; the lifespan keeps the stateful work (plugin DB seeding, MCP startup, static mount, plugin unload).

## Changes

- refactor(core): extract db-free assemble_app factory from lifespan
- test(core): patch consumer binding in assemble_app db-free guard
- test(plugins): drop conftest route registration made redundant by assemble_app
- docs: describe assemble_app route registration in architecture notes

## How to Test

1. `make test.backend` (includes the new tests/core/test_assemble_app.py: plugin routes present without lifespan, DB-free regression guard, full openapi schema)
2. `uv run python -c "from app.main import assemble_app; print(len(assemble_app().openapi()['paths']))"` prints 48 paths (all plugin endpoints included) with no server or DB running
3. `make backend.up.dev` boots clean; /docs lists plugin endpoints

## Notes

- No migrations, no new env vars, no request-handling behavior change; routes simply register at import time instead of server startup.
- Intentional fail-loud change: a plugin whose get_routes() raises ImportError/RuntimeError/OSError now surfaces at import time instead of being logged and skipped at startup (the loader-level failure path is still caught and logged). Per-plugin AttributeError/TypeError/ValueError handling is unchanged.
- The deleted conftest sentinel hack (tests/lib/routes.py) existed only to fake import-time registration; its circular-import workaround in the chat conftest is moot because app.main itself now imports plugin routes.
- Verified locally: full suite 1080/1080, mypy --strict, ruff, offline schema dump. Live `make backend.up.dev` boot was NOT exercised (Docker daemon not running here); the lifespan body is byte-identical to before, but please boot once during review.
- Unblocks #403 PR-3 (OpenAPI dump script + frontend codegen + CI drift check).


app/main.py
```diff
@@ -99,18 +78,56 @@ async def lifespan(application: FastAPI) -> AsyncIterator[None]:
             yield
 
 
-app = FastAPI(lifespan=lifespan)
```
regisb, line 101: I don't understand the difference with the previous implementation: app creation still depends on plugin_lifespan, which calls `plugin_service.get_or_create_all()`, so it makes a call to the database. What am I missing?

EDIT: the routes creation now happens before (or even without) database access, so we can generate the openapi spec offline.

abdulrafey1 (approved):

app/main.py
```diff
@@ -99,18 +78,56 @@ async def lifespan(application: FastAPI) -> AsyncIterator[None]:
...
+    plugin_loader = get_plugin_loader()
+    try:
+        loaded_plugins = plugin_loader.get_loaded_plugins()
+    except (ImportError, RuntimeError, OSError) as e:
+        logger.error(f"Plugin initialization failed: {e}")
+        return
+
+    loaded_plugin_names = [name for name, _plugin in loaded_plugins]
+    if loaded_plugins:
+        logger.info(f"Loaded {len(loaded_plugins)} plugin(s): {', '.join(loaded_plugin_names)}")
```
zaira-bibi: `loaded_plugin_names` should be computed inside the `if loaded_plugins:` guard. or better yet, since it is not being used anywhere else, just inline it in the `logger.info` call.

zaira-bibi (approved):
just one nit, rest looks good!

regisb (approved):
I wish I could delay this PR until [mine](#417) got merged so I don't have to deal with rebase but I'll live with it :)

hamza-56:
@claude-review

app/main.py
hamza-56: Addressed in b7fb924: moved the join inside the `if loaded_plugins:` guard and inlined it into the `logger.info` call.

PR #429, opened 2026-06-10
feat: openapi codegen pipeline with drift check and tsc gate

hamza-56:
## What

Third PR for #403, stacked on #428: the frontend's view of the API contract is now generated from the backend's OpenAPI schema instead of being re-typed by hand, and CI fails when the generated types are stale or when frontend type errors slip in (nothing type-checked the frontend in CI before).

## Changes

- feat(core): add offline openapi schema dump script (+ subprocess tests for sentinels and byte-determinism)
- feat(frontend): generate api types from the backend openapi schema (openapi-typescript dev dep, committed lib/api/generated.ts, make gen.api)
- feat(frontend): add typed openapi-fetch client with auth middleware (+ 5 tests)
- refactor(frontend): split client middleware and drop unneeded vitest option
- ci: add api-contract drift job and frontend typecheck gate
- docs: document make gen.api in essential commands

## How to Test

1. `make gen.api && git diff --exit-code frontend/lib/api/generated.ts` (idempotent regeneration)
2. `make test.backend` (dump-script tests) and `make test.frontend` (client middleware tests + the new tsc step)
3. Delete any route from a backend router locally, run `make gen.api`: the generated.ts diff appears; revert.

## Notes

- Stacked on #428 (assemble_app); review/merge that first. GitHub retargets this PR to main when the base branch is deleted.
- openapi.json is a gitignored build artifact; frontend/lib/api/generated.ts IS committed so backend contract changes land in the same PR diff (the api-contract job enforces freshness).
- The generated file is kept byte-stable by running oxfmt inside gen:api (no oxlint/oxfmt ignore configs exist in this repo); openapi-typescript and oxfmt versions are pinned via bun.lock and CI installs with --frozen-lockfile.
- The client throws the existing ApiRequestError (reusing lib/api's formatApiError) instead of introducing a new error class, deviating from the spec sketch deliberately so converted modules keep the error contract components already handle.
- Known quirk for the follow-up PRs: lib/api.ts (the file) shadows the new lib/api/ barrel for the "@/lib/api" import specifier; the call-site conversion PRs will fold api.ts into the directory.
- No call sites converted yet; that starts in the next PRs (one lib module each).


abdulrafey1 (approved):

frontend/lib/api/generated.ts
```diff
@@ -0,0 +1,3547 @@
+/**
```
regisb, line 1: FTR this corrresponds to versioning decision 2 in this comment:

https://github.com/edly-io/sparkth/issues/403#issuecomment-4621367365

We don't store openapi.json under version control, but we store generated.ts because we want to see API changes.

Makefile
```diff
@@ -79,6 +79,11 @@ frontend.install.dev: ## Install exact frontend dependencies from lockfile
 frontend.up.dev: ## Run frontend dev server (hot reload)
 	cd frontend && bun run dev
 
+.PHONY: gen.api
+gen.api: ## Regenerate frontend API types from the backend OpenAPI schema
+	uv run python scripts/dump_openapi.py > openapi.json
+	cd frontend && bun run gen:api
```
regisb: I propose the following changes:

1. Rename the  `gen.api` make target to `frontend.build.api`.
2. Move gen:api command to Makefile: `make frontend.build.api` with the `frontend.build` target depending on `frontend.build.api`. After all, we don't need package.json scripts if we have a Makefile, and make handles dependencies better than package.json.
3. Introduce a `make test.frontend.api`: which will run `openapi-typescript --check`. Have `test.frontend` depend on `test.frontend.api`.
4. Remove the awkward `git diff` test from CI: instead, API checks will be performed as part of `test.backend`. That way we don't have to introduce a new CI job.

hamza-56:
@claude-review

frontend/lib/api/client.ts
```diff
@@ -0,0 +1,42 @@
...
+      message: `Request failed (HTTP ${response.status}). Please try again.`,
+      fieldErrors: {} as Record<string, string>,
+    };
+    try {
+      const envelope: ApiError = await response.clone().json();
+      formatted = formatApiError(envelope);
+    } catch {
+      // Non-JSON error body: keep the generic message.
+    }
+    throw new ApiRequestError(formatted, response.status);
```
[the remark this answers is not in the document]
hamza-56: Good catch. Fixed in c5e52a3: `errorMiddleware` now extracts the structured detail via `isStructuredDetail` (now exported from `lib/api.ts`) and forwards it as the third `ApiRequestError` arg, and `client.test.ts` has a round-trip test for `403 { detail: { code: "email_not_verified", email: ... } }` that asserts both `err.code` and `err.data`.

Makefile
hamza-56: Applied (1), (2), and (3) in d39d606:

- `gen.api` is now `frontend.build.api`, with the codegen pipeline (dump + `openapi-typescript` + `oxfmt`) inlined in the Makefile; `gen:api` is gone from `frontend/package.json`.
- `frontend.build` now depends on `frontend.build.api`.
- New `test.frontend.api` regenerates and diffs `frontend/lib/api/generated.ts`; `test.frontend` depends on it.

For (4), I dropped the standalone `api-contract` CI job (6592d07) but folded the drift gate into `test.frontend.api` rather than `test.backend`. The check needs both the Python dump and the JS toolchain; the frontend job already has bun, so adding uv + a backend install there is cheaper than putting bun + frontend deps into the backend job. Result is the same as your proposal (no separate CI job), just a different host.

One note on the drift mechanism itself: `openapi-typescript --check` exists, but our committed `generated.ts` is post-`oxfmt`, so a raw `--check` always shows drift even on a clean checkout. `test.frontend.api` therefore runs the full pipeline into the working tree and then `git diff --exit-code`, which is the same thing the old CI job did, just moved.

Makefile
```diff
@@ -143,7 +149,16 @@ test.backend.format: ## Run backend formatting tests
 	$(MAKE) lint.format.backend check=1
 
 .PHONY: test.frontend
-test.frontend: lint.frontend lint.frontend.react-doctor test.frontend.vitest test.frontend.format ## Run frontend linting, react-doctor, unit and formatting tests
+test.frontend: lint.frontend lint.frontend.react-doctor test.frontend.api test.frontend.typecheck test.frontend.vitest test.frontend.format ## Run frontend linting, react-doctor, api drift, typecheck, unit and formatting tests
+
+.PHONY: test.frontend.api
+test.frontend.api: frontend.build.api ## Fail when generated.ts is stale vs the current backend OpenAPI schema
+	@git diff --exit-code frontend/lib/api/generated.ts \
```
regisb: If the only reason that we need to run `git diff` instead of `openapi-typescript --check` is the automated `oxfmt` I suggest we get rid of it for this file specifically and ignore it in our format check. Both approaches work for me so I'll go with whatever you suggest.

regisb (approved):

Makefile
hamza-56: Good call. Done in 08ce3af:

- Dropped the post-codegen `oxfmt` pass from `frontend.build.api`.
- Excluded `lib/api/generated.ts` via `frontend/.oxfmtrc.json` (`ignorePatterns`), and added `--no-error-on-unmatched-pattern` to the frontend-format lefthook so a commit touching only the generated file does not abort.
- `test.frontend.api` now runs `openapi-typescript --check` directly: dump the schema, then check the committed `generated.ts` against it.

PR #432, opened 2026-06-11
refactor(chat): decouple chat plugin from RAG by introducing a new public API function

abdulrafey1:
## What
Decouples the chat intent router from RAG MCP internals by exposing ingested document structure through the public RAG API.

## Changes
- refactor(chat): use `app.lib.rag` for RAG ingested document structure lookup
- refactor(rag): move ingested document structure lookup into RAG utilities
- refactor(mcp): make the MCP document-structure tool delegate to the RAG utility
- test(rag): cover public API exposure and ingested document structure behavior

## How to Test
1. `make test.backend.pytest`
2. `make lint.backend`
3. `make test.backend.format`
4. `make mypy`

## Notes
No migrations, frontend changes, env vars, or dependency changes. The MCP tool name remains `get_document_structure`; only the internal/public Python API name is clarified as `get_rag_ingested_document_structure`.

_This pull request description was written with the assistance of an LLM (Codex)._


regisb (commented):
Good direction overall — the decoupling is clean and the test migration is well-structured. Two items worth addressing before merge:

---

### 1. `DocumentSection` repr format change may affect RAG agent accuracy (medium)

**File:** `app/rag/mcp/agent_tools.py:37` / `app/rag/types.py`

The PR moves `DocumentSection` from a Pydantic `BaseModel` (in `schemas.py`) to a stdlib `@dataclass` (in `types.py`). LangChain serializes tool outputs by calling `json.dumps()` first and falling back to `str()`. Both types fall to `str()`, but their `__repr__` differs:

- **Pydantic BaseModel:** `source_name='doc.pdf' chapter='Chapter 1'`
- **stdlib dataclass:** `DocumentSection(source_name='doc.pdf', chapter='Chapter 1')`

The RAG agent system prompt instructs the LLM to copy `source_name` and section fields *exactly as returned by `get_document_structure`*. This tool is now the only one in `build_search_tools` that returns a non-Pydantic type — all others (`list_user_files`, `get_file_metadata`, `get_chunk_stats`, etc.) still return `schemas.*` Pydantic models.

Worth either aligning `DocumentSection` with the rest (keep as Pydantic, or convert the field description in `schemas.py` to a simple `@dataclass` wrapper that stays serialisation-compatible), or adding an explicit `__repr__`/`__str__` so the output format is stable and intentional.

---

### 2. Ordering contract dropped from `get_rag_ingested_document_structure` docstring (low)

**File:** `app/rag/utils.py`

The new function's docstring reads:

> Return ordered section metadata generated from the ingested RAG chunks. Returns an empty list when the document does not exist for the user.

The old `tools.py` docstring documented that rows are ordered by `func.min(DocumentChunk.id)` (document insertion order, not alphabetical), and that this is what makes `position_index` meaningful for positional reasoning ("second half", "last chapter"). The RAG agent system prompt still relies on this contract.

Without the docstring, a future change to the `ORDER BY` clause has no stated invariant to preserve, and the positional reasoning silently breaks. Suggest restoring the ordering explanation:

```python
"""Return ordered section metadata generated from the ingested RAG chunks.

Sections are ordered by the minimum chunk id within each (chapter, section,
subsection) group — i.e. document insertion order — so position_index reliably
reflects the section's position in the original document.

Returns an empty list when the document does not exist for the user.
"""
```

---

regisb (approved):
feel free to address the agent's comments or not. For me this is good to go.

PR #433, opened 2026-06-11
refactor(rag): make RAG retrieval independent of user ID

abdulrafey1:
## What
Remove user ownership checks from RAG retrieval so callers validate access before passing document IDs into RAG.

## Changes
- refactor(rag): make retrieval, agent tools, and ingested structure lookup document-scoped
- refactor(plugins): validate chat document ownership/readiness before invoking RAG
- test(rag): update RAG, chat, and Slack retrieval coverage for document-scoped calls
- refactor(rag): drop user_id from RAGIntentRouter.decide() and resolve_rag_intent() — param was unused
- refactor(rag): remove redundant validate_ready_user_documents pre-check from chat paths — agentic_retrieve_context validates internally
- refactor(rag): remove dead list_user_documents tool — no production callers remain
- docs(rag): update agentic_retrieve_context docstring to be explicit that callers don't need to validate readiness
- refactor(rag): add TODO for batching per-document structure lookups in intent_router

## How to Test
1. `make test.backend.pytest`
2. `make lint.backend`
3. `make mypy`
4. `make test.backend.format`

## Notes
No migration, dependency, or env var changes.

### DIFF
--- a/app/core_plugins/chat/intent_router.py
+++ b/app/core_plugins/chat/intent_router.py
@@ -36,14 +36,12 @@ class RAGIntentRouter:
         *,
         query: str,
         attached_documents: list[Document],
-        user_id: int,
     ) -> RAGRoutingDecision:
         """Decide whether to run RAG retrieval for this turn.
 
         Args:
             query: The user's query text.
             attached_documents: Documents attached to the conversation.
-            user_id: The user ID for the current chat request.
 
         Returns:
             RAGRoutingDecision with should_retrieve and reason.
@@ -55,6 +53,7 @@ class RAGIntentRouter:
         attachment_summary = ""
         if attached_documents:
             documents = [doc for doc in attached_documents if doc.id is not None]
+            # TODO: batch-lookup document structures in a single query instead of one coroutine per document
             results = await asyncio.gather(
                 *[get_rag_ingested_document_structure(document_id=cast(int, doc.id)) for doc in documents],
                 return_exceptions=True,
--- a/app/core_plugins/chat/intent_router.py
+++ b/app/core_plugins/chat/intent_router.py
@@ -43,7 +43,7 @@ class RAGIntentRouter:
         Args:
             query: The user's query text.
             attached_documents: Documents attached to the conversation.
-            user_id: The user ID passed to the public RAG ingested-document structure API.
+            user_id: The user ID for the current chat request.
 
         Returns:
             RAGRoutingDecision with should_retrieve and reason.
@@ -56,10 +56,7 @@ class RAGIntentRouter:
         if attached_documents:
             documents = [doc for doc in attached_documents if doc.id is not None]
             results = await asyncio.gather(
-                *[
-                    get_rag_ingested_document_structure(user_id=user_id, document_id=cast(int, doc.id))
-                    for doc in documents
-                ],
+                *[get_rag_ingested_document_structure(document_id=cast(int, doc.id)) for doc in documents],
                 return_exceptions=True,
             )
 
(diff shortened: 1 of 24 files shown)


app/core_plugins/chat/intent_router.py
```diff
@@ -43,7 +43,7 @@ async def decide(
         Args:
             query: The user's query text.
             attached_documents: Documents attached to the conversation.
-            user_id: The user ID passed to the public RAG ingested-document structure API.
+            user_id: The user ID for the current chat request.
```
hamza-56: `user_id` is unused in this method now, this was its last consumer. Might as well drop the param here and from `resolve_rag_intent`, which only exists to forward it.

app/core_plugins/chat/intent_router.py
```diff
@@ -56,10 +56,7 @@ async def decide(
         if attached_documents:
             documents = [doc for doc in attached_documents if doc.id is not None]
             results = await asyncio.gather(
```
hamza-56, line 57: Not new in this PR, but since this code is being touched anyway: this opens a session and runs a query per attached document. A batched lookup taking `document_ids` would do it in one round trip.

app/core_plugins/chat/routes/helpers.py
```diff
@@ -191,6 +193,29 @@ async def _retrieve_rag_chunks(
         ) from exc
 
 
+async def validate_ready_user_documents(
```
hamza-56: This is almost line-for-line `validate_documents_ready` from `app/rag/retrieval/utils.py`, just with the `user_id` predicate added. Could we give that one an optional `user_id` filter and expose it via `app/lib/rag.py` instead? Saves the next plugin (Slack?) from copying it again.

app/rag/retrieval/__init__.py
```diff
@@ -36,17 +34,17 @@ async def agentic_retrieve_context(
...
+        DocumentNotFoundError: a document is missing or soft-deleted.
         RAGNotReadyError: a document exists but is not READY.
         RAGRetrievalError: retrieval failed.
     """
     if not document_ids:
         return []
 
     async with session_scope() as session:
-        await validate_documents_ready(session, user_id, document_ids)
+        await validate_documents_ready(session, document_ids)
```
hamza-56, line 46: On the chat path this now runs twice per request: the route already did `validate_ready_user_documents`, and this re-runs the same `SELECT ... WHERE id IN (...)`. Fine as defense in depth, but if the shared helper idea lands, pre-validated callers could skip this.

app/rag/mcp/agent_tools.py
```diff
@@ -12,45 +12,36 @@
...
         document_id: Document ID the search is scoped to, injected where relevant.
 
     Returns:
         LangChain tools exposing only the arguments the agent should choose.
     """
 
-    async def list_user_documents() -> list[schemas.DocumentInfo]:
-        return await tools.list_user_documents(user_id=user_id)
-
     async def get_document_metadata() -> schemas.DocumentMetadata | None:
```
hamza-56, line 25: With the `list_user_documents` wrapper gone, `tools.list_user_documents` has no production callers left (only its own tests). Worth removing it too, or leaving a note if it is being kept for future MCP exposure.

app/core_plugins/chat/routes/helpers.py
abdulrafey1: The entire purpose of the PR is to make rag retrieval independent of user ID scope. So we cant add user_id to `app.rag.retrieval.utils.validate_documents_ready`

Second we cant let other plugins use internal implementations of the rag module. For now we are ok with some plugins duplicating some utility logic, because not everything can be put into the shared public lib.

app/core_plugins/chat/intent_router.py
abdulrafey1: this is outside of the scope of this PR. I can add a TODO comment so this does not get lost.

hamza-56 (approved):

PR #440, opened 2026-06-12
refactor(frontend): convert lib/api to the typed openapi client

hamza-56:
## What

Fifth PR for #403 and the first call-site conversion: the auth, whitelist, email-verification and current-user calls go through the shared typed openapi-fetch client, so their paths and payloads are now compile-checked against the backend schema.

## Changes

- refactor(frontend): convert all nine lib/api endpoint functions to api.METHOD(...) calls and replace the hand-written request/response interfaces with generated schema re-exports (UserLogin, Token, UserCreate, User, GoogleAuthUrl, WhitelistedEmailResponse)
- refactor(frontend): fold lib/api.ts into lib/api/ (errors.ts for the FastAPI error vocabulary, middleware.ts for auth + error middlewares, client.ts for the client wiring, endpoints.ts for the calls); the file no longer shadows the barrel for "@/lib/api"
- refactor(frontend): let explicit Authorization headers win over the stored token in authMiddleware, so token parameters keep their meaning

## How to Test

1. `cd frontend && bun run vitest run lib/tests/api.test.ts lib/api/client.test.ts` (24 tests pin per-endpoint paths, methods, bodies, auth headers and error contracts)
2. `make test.frontend` (oxlint, react-doctor, api drift gate, tsc --noEmit, vitest, format check; all green locally)
3. Manual golden path: log in, dashboard shows the user, admin whitelist add/remove an entry, wrong password shows the inline error.

## Notes

- No backend change; generated.ts is untouched and the drift gate stays green.
- Consumer imports are unchanged: lib/api/index.ts re-exports every name lib/api.ts used to export.
- CurrentUser is now the generated User schema. That surfaced two dead reads in app/dashboard/layout.tsx (user.avatar, user.plan): the backend never sends either field, so the sidebars' fallbacks were already what rendered. The layout now builds userInfo without them; runtime output is identical.
- Accepted minor divergence: when an error response body is not valid JSON, the old per-endpoint fallback strings ("Login failed. Please try again.", ...) are replaced by the shared middleware's generic "Request failed (HTTP n). Please try again." FastAPI always sends JSON envelopes, so this only fires for proxy-level failures.
- Browser verification was skipped this session; beyond the unit suite, the live error envelopes (401 wrong password, structured 403 email_not_verified) were verified against a running backend through the Next dev proxy.


hamza-56:
@claude-review

frontend/lib/api/endpoints.ts
```diff
@@ -0,0 +1,120 @@
...
+  return { Authorization: `Bearer ${token}` };
+}
+
+// errorMiddleware turns every non-ok response into an ApiRequestError; anything
+// else rejecting here is a transport failure (DNS, refused connection, ...).
+function wrapConnectionError(error: unknown): never {
+  if (error instanceof ApiRequestError) throw error;
+  const message = error instanceof Error ? error.message : "Unknown error";
+  throw new ApiRequestError({
+    message: `Unable to connect to server: ${message}`,
```
zaira-bibi: this will missclassify `AbortError` as a connection failure once any endpoint receives an `AbortSignal`. maybe add a `DOMException`/`AbortError` guard here.

frontend/lib/api/endpoints.ts
```diff
@@ -0,0 +1,120 @@
...
+  });
+}
+
+export async function login(data: LoginRequest): Promise<LoginResponse> {
+  try {
+    const { data: token } = await api.POST("/api/v1/auth/login", { body: data });
+    return token as LoginResponse;
+  } catch (error) {
+    wrapConnectionError(error);
+  }
```
zaira-bibi: 8 endpoint functions share identical try/catch structure. a higher-order `call()` helper would centralise it.

zaira-bibi (approved):
just two minor suggestions, rest looks good!

frontend/lib/tests/api/client.test.ts
abdulrafey1, line 1: Can we move this test file to `lib/tests/api/client.test.ts` and place the `lib/tests/api.test.ts` in the same directory as well

hamza-56: Done in 7352710 — moved both `client.test.ts` and `api.test.ts` into `lib/tests/api/`, and switched `client.test.ts` from the relative `./client` import to the `@/lib/api/client` alias so it matches the other `lib/tests` modules.

frontend/lib/api/endpoints.ts
hamza-56: Done in 35a5fb9 — pulled the shared try/catch into a `call()` helper that the eight simple endpoints now use. `resendVerificationEmail` keeps its own try/catch since its mapping (429 `rate_limited` sentinel, fixed message otherwise) differs.

frontend/lib/api/endpoints.ts
hamza-56: Good catch. 35a5fb9 adds an `AbortError` guard to `wrapConnectionError` (matches a `DOMException` or `Error` whose `name === "AbortError"`) so a cancelled request propagates untouched instead of being relabelled "Unable to connect to server", with a test covering it.

frontend/lib/api/endpoints.ts
```diff
@@ -0,0 +1,120 @@
...
+  const message = error instanceof Error ? error.message : "Unknown error";
+  throw new ApiRequestError({
+    message: `Unable to connect to server: ${message}`,
+    fieldErrors: {},
+  });
+}
+
+export async function login(data: LoginRequest): Promise<LoginResponse> {
+  try {
+    const { data: token } = await api.POST("/api/v1/auth/login", { body: data });
```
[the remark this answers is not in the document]
hamza-56: The cast now lives in one place (the `call()` helper in 35a5fb9). I kept it as `data as T` rather than asserting presence: the void endpoints (`removeWhitelistEntry`, `verifyEmail`) legitimately return no body, so a blanket empty-body throw would break them, and this conversion deliberately preserves the existing contract.

hamza-56:
@claude-review

abdulrafey1 (approved):

PR #441, opened 2026-06-12
refactor(frontend): convert lib/llm to the typed openapi client

hamza-56:
## What

Sixth PR for #403: the LLM config and provider-catalog calls go through the shared typed openapi-fetch client, compile-checked against the backend schema.

## Changes

- refactor(frontend): convert all seven lib/llm functions to api.METHOD(...) calls (configs list/create/update/rotate-key/set-active/delete, provider catalog)
- refactor(frontend): replace the six hand-written interfaces in lib/llm/types.ts with generated schema re-exports (LLMConfigResponse, LLMConfigListResponse, ProviderInfo, ProviderCatalogResponse, LLMConfigCreate, LLMConfigUpdate)

## How to Test

1. `cd frontend && bun run vitest run lib/tests/llm.test.ts` (12 tests pin paths, methods, bodies, auth headers, abort passthrough and error contracts)
2. `make test.frontend` (oxlint, react-doctor, drift gate, tsc, vitest, format)
3. Manual: AI Keys page lists/creates/renames/deactivates/deletes a key; chat and slack config modals load the provider catalog.

## Notes

- Stacked on #440 (lib/api fold); review/merge that first. GitHub retargets this PR when the base branch is deleted, and repo CI runs once it targets main.
- Function signatures and lib/llm/index.ts are unchanged, so the six consumer files need no edits.
- Envelope errors now carry the response status and structured detail (the old llmFetch dropped both); AbortError still passes through unwrapped for the catalog prefetch.
- ProviderCatalogResponse tightens default_provider/default_model from "string | null" to "string", matching what the backend actually declares; tsc is clean across consumers.
- Same accepted divergence as #440: non-JSON error bodies get the middleware's generic message instead of the old per-action fallback strings.

### DIFF
--- a/frontend/lib/llm/client.ts
+++ b/frontend/lib/llm/client.ts
@@ -1,4 +1,4 @@
-import { api, ApiRequestError } from "@/lib/api";
+import { api, bearer, rethrowOrWrapConnectionError } from "@/lib/api";
 import type {
   CreateLLMConfigPayload,
   LLMConfig,
@@ -6,22 +6,6 @@ import type {
   ProviderCatalogResponse,
   UpdateLLMConfigPayload,
 } from "@/lib/llm/types";
-
-function bearer(token: string): { Authorization: string } {
-  return { Authorization: `Bearer ${token}` };
-}
-
-// errorMiddleware turns every non-ok response into an ApiRequestError; aborts
-// are flow control and pass through; anything else is a transport failure.
-function rethrowOrWrapConnectionError(error: unknown): never {
-  if (error instanceof ApiRequestError) throw error;
-  if (error instanceof DOMException && error.name === "AbortError") throw error;
-  const message = error instanceof Error ? error.message : "Unknown error";
-  throw new ApiRequestError({
-    message: `Unable to connect to server: ${message}`,
-    fieldErrors: {},
-  });
-}
 
 export async function fetchLLMConfigs(
   token: string,
--- a/frontend/lib/llm/client.ts
+++ b/frontend/lib/llm/client.ts
@@ -1,4 +1,4 @@
-import { ApiRequestError, type ApiError, formatApiError } from "@/lib/api";
+import { api, ApiRequestError } from "@/lib/api";
 import type {
   CreateLLMConfigPayload,
   LLMConfig,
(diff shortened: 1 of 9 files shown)


hamza-56:
@claude-review

abdulrafey1:
Something that is not specific to this PR but overall, can we take out the `api/v1` to some generic constant because that seems to be defined every URL.

Or will the migration to 1openapi-fetch` completely take the URL out of the question? if yes then we can ignore this

frontend/lib/llm/client.ts
```diff
@@ -7,72 +7,50 @@ import type {
...
-const API_BASE = "/api/v1/llm";
-
-function jsonHeaders(token: string): HeadersInit {
-  return {
-    "Content-Type": "application/json",
-    Accept: "application/json",
-    Authorization: `Bearer ${token}`,
-  };
+function bearer(token: string): { Authorization: string } {
+  return { Authorization: `Bearer ${token}` };
```
zaira-bibi: `bearer` and `rethrowOrWrapConnectionError` are defined almost identically in both `lib/api/endpoints.ts` and `lib/llm/client.ts`. can we export both from `lib/api/`? as `llm/client.ts` already imports from there the call sites stay clean.

frontend/lib/tests/llm.test.ts
```diff
@@ -1,28 +1,166 @@
...
+  name: "Work key",
+  provider: "anthropic",
+  model: "claude-sonnet-4-20250514",
+  masked_key: "sk-...abcd",
+  is_active: true,
+  created_at: "2026-01-01T00:00:00Z",
+  last_used_at: null,
+};
+
+function mockFetch(body: unknown, status = 200) {
```
zaira-bibi: same as the above comment for the `mockFetch`/`sentRequest` test helpers, which are duplicated across `api.test.ts` and `llm.test.ts`. a `lib/tests/test-utils.ts` would cover both.

zaira-bibi (changes requested):

frontend/lib/llm/client.ts
hamza-56: Done. `bearer()` now lives in `lib/api/client.ts` and `rethrowOrWrapConnectionError()` in `lib/api/errors.ts`, both re-exported from `lib/api`; `lib/llm/client.ts` imports them and drops its local copies (7b5a5ff, 6775f95).

The unified connection-error helper keeps the `AbortError` guard, so the catalog prefetch's abort still propagates unwrapped.

frontend/lib/tests/llm.test.ts
hamza-56: Done. `mockFetch`/`sentRequest` moved to `lib/tests/test-utils.ts`, and both `api.test.ts` and `llm.test.ts` now import them from there (7b5a5ff, 6775f95).

abdulrafey1 (approved):

zaira-bibi (approved):

PR #442, opened 2026-06-12
refactor(frontend): convert lib/slack-api to the typed openapi client

hamza-56:
## What

Seventh PR for #403: the Slack plugin's five calls (OAuth status/authorize/disconnect, logs, RAG sources) go through the shared typed openapi-fetch client.

## Changes

- refactor(frontend): convert lib/slack-api.ts to api.METHOD(...) calls and replace its hand-written response interfaces with generated re-exports (namespaced slack ConnectionStatusResponse/AuthorizationUrlResponse, BotResponseLogItem, ConnectionLogItem, LogsResponse, RagSourcesResponse)
- fix(frontend): coalesce the now-optional next_cursor to null in LogStream pagination actions

## How to Test

1. `cd frontend && bun run vitest run lib/tests/slack.test.ts` (9 new tests pin paths, query serialization, the plain-Error contract and raw network-error passthrough)
2. `make test.frontend`
3. Manual: Slack plugin page shows connection status, connect/disconnect works, log stream paginates, doc source picker lists sources.

## Notes

- Stacked sibling of #441 on #440 (lib/api fold); merges after #440. CI runs once retargeted to main.
- This module's error contract differs from lib/api and lib/llm: it throws plain Error objects with "Failed to <action>: <detail>" messages and lets network failures propagate untouched. Both behaviors are preserved and now pinned by tests.
- FetchLogsOptions stays hand-written (client-side option bag); openapi-fetch omits undefined query params, matching the old conditional URLSearchParams exactly.
- Same accepted divergence as the previous conversions for non-JSON error bodies.

_This PR description was written with the assistance of an LLM (Claude)._

PR #443, opened 2026-06-12
refactor(frontend): convert lib/drive to the typed openapi client

hamza-56:
## What

Eighth PR for #403: the Google Drive plugin's fourteen calls (OAuth, folders, files, browse, RAG status) go through the shared typed openapi-fetch client.

## Changes

- refactor(frontend): convert lib/drive.ts to api.METHOD(...) calls and replace hand-written interfaces with generated re-exports (DriveFolderResponse, DriveFileResponse, DriveBrowseResponse, SyncStatusResponse, FolderRagStatusResponse, namespaced googledrive ConnectionStatusResponse)
- refactor(frontend): multipart upload via custom bodySerializer with unset Content-Type (fetch adds the boundary); blob download via parseAs
- fix(frontend): coalesce newly-optional fields in consumers (useRagStatusPolling rag_status/rag_error, DriveFilePicker mime_type/size)

## How to Test

1. `cd frontend && bun run vitest run lib/tests/drive.test.ts` (21 tests pin paths, methods, bodies, pagination queries, multipart boundary, blob download and the plain-Error contract)
2. `make test.frontend`
3. Manual: Google Drive plugin connects, browses Drive, syncs a folder, uploads/renames/downloads/deletes a file, RAG status chips update.

## Notes

- Stacked sibling of #441/#442 on #440 (lib/api fold); merges after #440. CI runs once retargeted to main.
- Error contract preserved: plain Error with "Failed to <action>: <detail>" plus console.error, network failures propagate untouched (both pinned by tests).
- RagStatus stays a runtime const, now typed by the generated DocumentStatus enum (values verified identical).
- PaginatedResponse<T> stays a hand-written generic (structurally identical to the generated concrete envelopes) so fetchAllPages keeps its typing.
- The upload test pins multipart via the boundary header; reading the Request body back hangs under jsdom (jsdom File vs undici stream), noted inline.
- Same accepted non-JSON-body divergence as the sibling conversions.

### DIFF
--- a/frontend/lib/drive.ts
+++ b/frontend/lib/drive.ts
@@ -1,53 +1,15 @@
-const API_BASE_URL = "/api/v1/google-drive";
+import { api, ApiRequestError, type Schema } from "@/lib/api";
 
-export interface ConnectionStatus {
-  connected: boolean;
-  email?: string;
-  expires_at?: string;
-}
-
-export interface DriveFolder {
-  id: number;
-  drive_folder_id: string;
-  name: string;
-  parent_id?: string;
-  file_count: number;
-  last_synced_at?: string;
-  sync_status: string;
-}
-
-export interface DriveFile {
-  id: number;
-  drive_file_id: string;
-  document_id?: number | null;
-  name: string;
-  mime_type?: string;
-  size?: number;
-  modified_time?: string;
-  last_synced_at?: string;
-  rag_status?: RagStatus | null;
-  rag_error?: string | null;
-}
-
-export interface DriveFolderWithFiles extends DriveFolder {
-  files: DriveFile[];
-}
-
-export interface DriveBrowseItem {
-  id: string;
-  name: string;
-  mime_type: string;
-  is_folder: boolean;
-  modified_time?: string;
-  size?: number;
-}
-
-export interface SyncStatus {
-  folder_id: number;
-  sync_status: string;
-  last_synced_at?: string;
-  error?: string;
-}
+export type ConnectionStatus =
+  Schema<"app__core_plugins__googledrive__schemas__ConnectionStatusResponse">;
+export type DriveFolder = Schema<"DriveFolderResponse">;
+export type DriveFile = Schema<"DriveFileResponse">;
+export type DriveFolderWithFiles = Schema<"DriveFolderWithFilesResponse">;
+export type DriveBrowseItem = Schema<"DriveBrowseItem">;
+export type DriveBrowseResponse = Schema<"DriveBrowseResponse">;
+export type SyncStatus = Schema<"SyncStatusResponse">;
+export type FileRagStatus = Schema<"FileRagStatusResponse">;
+export type FolderRagStatusResponse = Schema<"FolderRagStatusResponse">;
 
 export interface PaginatedResponse<T> {
   items: T[];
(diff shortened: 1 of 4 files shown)


frontend/lib/drive.ts
```diff
@@ -56,64 +18,58 @@ export interface PaginatedResponse<T> {
...
 
-async function handleError(message: string, response: Response) {
-  const text = await response.text();
-  let detail = text;
-  try {
-    const json = JSON.parse(text);
-    if (json.detail) detail = json.detail;
-  } catch {
-    // use raw text
+export const RagStatus = {
```
abdulrafey1, line 28: could there be a way where we could get these from the backend?
Defining them here separately would put this at risk of drifting apart from the backend over time.

frontend/lib/drive.ts
```diff
@@ -1,53 +1,15 @@
...
-export interface SyncStatus {
-  folder_id: number;
-  sync_status: string;
-  last_synced_at?: string;
-  error?: string;
-}
+import { api, ApiRequestError, type Schema } from "@/lib/api";
+
+export type ConnectionStatus =
+  Schema<"app__core_plugins__googledrive__schemas__ConnectionStatusResponse">;
```
abdulrafey1, line 4: why does this have the entire path and the rest below dont?

frontend/lib/drive.ts
```diff
@@ -56,64 +18,58 @@ export interface PaginatedResponse<T> {
...
-    if (json.detail) detail = json.detail;
-  } catch {
-    // use raw text
+export const RagStatus = {
+  Queued: "queued",
+  Processing: "processing",
+  Ready: "ready",
+  Failed: "failed",
+} as const;
+export type RagStatus = Schema<"DocumentStatus">;
```
abdulrafey1, line 34: lets add a TODO comment here that we need to update the terminology of frontend to follow the new Document API usage as well. 

`RagStatus` was changed to `DocumentStatus` in the backend but I assume here we are no changing the name to keep it compatible with rest of the code, but we should do this in a separate PR

frontend/lib/drive.ts
```diff
@@ -56,64 +18,58 @@ export interface PaginatedResponse<T> {
...
-    headers: {
-      Accept: "application/json",
-      Authorization: `Bearer ${token}`,
-    },
-  });
-
-  if (!response.ok) {
-    await handleError("Failed to get connection status", response);
+  try {
+    const { data } = await api.GET("/api/v1/googledrive/oauth/status", {
```
abdulrafey1: can we extract `/api/v1/googledrive` in a constant on top of the file and reuse it everywhere below?

frontend/lib/drive.ts
hamza-56: There is no way to get the runtime values from the backend client: openapi-typescript emits types only, and types are erased at compile time, so the values have to live somewhere in the frontend. Addressed the drift risk in ab215e1 instead: the object is now `satisfies Record<string, Schema<"DocumentStatus">>`, so if the backend enum ever changes, regenerating the client makes the type check fail right here.

frontend/lib/drive.ts
hamza-56: Name collision: both the googledrive and slack plugins define a `ConnectionStatusResponse` (and an `AuthorizationUrlResponse`), so the OpenAPI generator disambiguates the duplicates with the full module path. Schemas with unique names keep the short form. The slack conversion (#442) has the same pattern from the other side.

frontend/lib/drive.ts
hamza-56: Added the TODO in ab215e1. Exactly that: the type already aliases the backend's `DocumentStatus`, the `RagStatus` name is kept so the rest of the frontend compiles unchanged, and the rename will be its own PR.

frontend/lib/drive.ts
hamza-56: I'd keep the full literals here. The other PRs in this stack actually remove their `API_BASE` constants (#441 llm, #442 slack, #446 chat) as part of the conversion, because openapi-fetch keys each request on the literal path from the generated schema: the string is type-checked against the backend, so it cannot drift, and full paths stay greppable. A `const` template literal would still type-check, but then drive.ts would be the one module in the stack doing it differently. If you feel strongly, I can apply it across all six PRs for consistency.

frontend/lib/drive.ts
hamza-56: Followed up on this: opened #449 off main to rename the colliding models to plugin-prefixed names (`SlackConnectionStatusResponse` / `GoogleDriveConnectionStatusResponse`, and the same for `AuthorizationUrlResponse`), which makes the generated client use the short names. Once #449 lands, this PR and #442 rebase and switch `Schema<"app__core_plugins__...">` to the short keys.

abdulrafey1 (approved):

PR #444, opened 2026-06-12
refactor(frontend): convert lib/file-upload to the typed openapi client

hamza-56:
## What

Ninth PR for #403: the parser upload call goes through the shared typed openapi-fetch client, and the module sheds its snake_case filename.

## Changes

- refactor(frontend): convert lib/file_upload.ts to api.POST with the multipart bodySerializer pattern and rename it to lib/file-upload.ts (the one snake_case outlier in lib/)
- test(frontend): pin path, method, multipart boundary, optional bearer header, bare-detail plain-Error contract and raw network passthrough (4 tests)

## How to Test

1. `cd frontend && bun run vitest run lib/tests/file-upload.test.ts`
2. `make test.frontend`
3. Manual: attach a local file in chat; the parsed text lands in the input attachment.

## Notes

- Stacked sibling of #441/#442/#443 on #440; merges after #440. CI runs once retargeted to main.
- UploadResponse stays hand-written: the backend route has no response_model, so the generated 200 type is unknown. Adding one is a possible backend follow-up.
- Theoretical divergence: when called without a token while a token sits in storage, the shared authMiddleware now injects it (the old code sent no header). The only caller always passes the token of the logged-in user, so behavior is unchanged in practice.


hamza-56:
@claude-review

abdulrafey1 (approved):

PR #445, opened 2026-06-12
refactor(frontend): convert lib/plugins/context to the typed openapi client

hamza-56:
## What

Tenth PR for #403: the plugin framework's three user-plugins calls (list, config update, enable/disable) go through the shared typed openapi-fetch client; the dynamic callApi passthrough deliberately stays raw fetch.

## Changes

- refactor(frontend): convert fetchUserPlugins, updatePluginConfigApi and togglePluginApi to api.METHOD(...) calls
- refactor(frontend): drop the dead description field from UserPluginState (the backend UserPluginResponse never sends it; ListItem now reads only the registry definition's description, which is what always rendered)
- docs(frontend): document why callApi keeps raw fetch (runtime-built paths cannot be statically typed; the spec's one exception)

## How to Test

1. `cd frontend && bun run vitest run lib/tests/plugins-context.test.ts` (5 provider-driven tests pin the mount fetch, both toggle endpoints, the config PUT body, and the error contract)
2. `make test.frontend`
3. Manual: settings page lists plugins, toggling enable/disable works, saving a plugin config persists, dashboard shows enabled plugin cards.

## Notes

- Stacked sibling of #441..#444 on #440; merges after #440. CI runs once retargeted to main.
- Error contract: plain Errors with the existing prefixes; the list/toggle messages now carry the middleware-formatted detail instead of the raw response body text (e.g. "Failed to fetch plugins: Forbidden" rather than the JSON envelope verbatim), and updateConfig surfaces the bare detail exactly as before.
- Token params stay explicit per the stack-wide decision; a follow-up can drop them everywhere at once and lean on the auth middleware.
- UserPluginState.config keeps its pre-existing Record<string, string> typing; tightening it against the wire's unknown values is plugin-framework scope, not call-conversion scope.

### DIFF
--- a/frontend/components/settings/ListItem.tsx
+++ b/frontend/components/settings/ListItem.tsx
@@ -84,7 +84,7 @@ export default function PluginListItem({
             </div>
             <p className="text-sm text-muted mb-2">Sparkth</p>
             <p className="text-sm text-muted-foreground leading-relaxed">
-              {pluginDef?.description || plugin.description}
+              {pluginDef?.description}
             </p>
 
             {toggleError && (
--- a/frontend/lib/plugins/context.tsx
+++ b/frontend/lib/plugins/context.tsx
@@ -9,6 +9,7 @@ import {
   useMemo,
   ReactNode,
 } from "react";
+import { api, ApiRequestError } from "@/lib/api";
 import {
   PluginDefinition,
   PluginConfig,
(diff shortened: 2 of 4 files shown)


hamza-56:
@claude-review

abdulrafey1 (approved):

PR #446, opened 2026-06-12
refactor(frontend): convert lib/chat to the typed openapi client

hamza-56:
## What

Eleventh and final conversion PR for #403: the chat plugin's seven calls go through the shared typed openapi-fetch client, including the SSE completions request, whose path, body and auth header are now compile-checked.

## Changes

- refactor(frontend): convert lib/chat/client.ts to api.METHOD(...) calls (completions via parseAs: "stream" returning the raw Response; conversations, attachments, last-message)
- refactor(frontend): type the completions body as the generated ChatCompletionRequest and drop similarity_threshold from the wire (the backend schema never had it; pydantic ignored it)
- refactor(frontend): replace useChatStream's hand-rolled error-envelope parsing with the structured ApiRequestError (the bespoke validation formatter is gone; the friendly "No AI Key selected" mapping now keys off fieldErrors)

## How to Test

1. `cd frontend && bun run vitest run lib/tests/chat.test.ts` (19 tests pin paths, bodies, the Bearer-null quirk, abort/network handling, soft failures and error message shapes)
2. `make test.frontend`
3. Manual: send a chat message and watch it stream; switch conversations; attach/detach a drive document; remove the AI key selection and send (friendly error shows).

## Notes

- Last stacked sibling of #441..#445 on #440; merges after #440. CI runs once retargeted to main.
- Surfaced drift, now fixed: similarity_threshold never reached the backend (not in ChatCompletionRequest), so the "Try with less strict matching" retry was always a plain retry and remains one; llm_config_id is required by the schema while the hook can hold undefined, which is documented at the call site and produces the same 422 -> friendly-message flow as before (a UI guard is a possible follow-up).
- The hook now sends temperature: 0.7 explicitly (the schema default; openapi-typescript marks defaulted fields required for senders). Wire behavior is unchanged.
- ApiMessage/ApiConversation stay hand-written narrowings (role/message_type unions, rag_sections/tool_calls shapes): those are JSON dicts on the wire, the same category as the SSE event payloads the spec deliberately leaves hand-typed.
- ConversationSummary/PersistedAttachment/ChatCompletionRequestBody are now generated re-exports; OutgoingChatMessage is deleted.
- Completions error-path divergences: validation arrays format via the shared formatApiError, and non-JSON bodies get the middleware generic message instead of "Something went wrong. Try again.".

### DIFF
--- a/frontend/lib/chat/client.ts
+++ b/frontend/lib/chat/client.ts
@@ -1,3 +1,4 @@
+import { api, ApiRequestError } from "@/lib/api";
 import type {
   ApiConversation,
   ApiMessage,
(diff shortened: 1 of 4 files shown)


hamza-56:
@claude-review

abdulrafey1 (approved):

PR #447, opened 2026-06-12
refactor(rag): make RAG ingestion independent of user ID

abdulrafey1:
## What
Remove `user_id` from the RAG ingestion pipeline so RAG operates on documents only, not ownership — mirroring what was done for retrieval in #433.

## Changes
- refactor(rag): remove `user_id` param from `ingest_document`
- refactor(rag): remove `user_id` param from `store_and_link_chunks`
- refactor(rag): remove `user_id` param from `ChunkStoreService.store_chunks`, `delete_by_source`, and `get_sources`
- refactor(rag): remove `user_id` field from `DocumentChunk` model
- migrations: drop `user_id` column, FK constraint, and indexes from `rag_document_chunks`
- refactor(plugins): update Google Drive caller to drop `user_id` from `ingest_document` call
- test: update all affected test files

## How to Test
1. `make test.backend` — all tests should pass
2. `make mypy` — no type errors
3. `make migrations` — migration applies cleanly
4. Ingest a document via the Google Drive plugin and verify it is processed successfully

## Notes
Migration required: drops `user_id` column and `idx_rag_chunks_user_source` index from `rag_document_chunks`. Downgrade path included.


app/rag/store.py
```diff
@@ -129,37 +122,26 @@ def _key_condition(key: dict[str, str | None]) -> Any:
...
 
         Note: this method flushes but does not commit. The caller is
         responsible for committing (or rolling back) the transaction.
         """
-        stmt = (
-            delete(DocumentChunk)
-            .where(col(DocumentChunk.user_id) == user_id)
-            .where(col(DocumentChunk.source_name) == source_name)
-        )
+        stmt = delete(DocumentChunk).where(col(DocumentChunk.source_name) == source_name)
```
zaira-bibi, line 132: from my understanding, if two users have uploaded the same document and one of them calls `delete_by_source`, then it would delete the other's chunks as well. is that right?

regisb (approved):
Looks good!

app/rag/store.py
abdulrafey1: oh thats a great catch, the entire thing was wired to not let this happen.

1. The document deletion was only supposed to delete the `link` between an ingested chunk and the document, never the actual chunk from the DB
2. The `rag-cleanup` job was meant to run periodically to cleanup all the orphaned chunks.

Seems like in the refactoring, the soft deletion changed into hard deletion. I have opened a new issue #456 to tackle this in a later PR

PR #448, opened 2026-06-12
refactor(rag): extract shared chunk formatting utilities to app.lib.rag

abdulrafey1:
## What
Eliminates code duplication between the chat and slack plugins by moving `format_source_block` and `group_by_source` to a shared module exposed through the public RAG API.

## Changes
- `feat(rag)`: add `format_source_block` and `group_by_source` to `app/rag/formatting.py` and expose via `app/lib/rag.py`
- `refactor(chat)`: remove local definitions of `format_source_block`/`group_by_source` from `helpers.py` and `completions.py`; import from `app.lib.rag`
- `refactor(slack)`: delete `_format_context` (near-duplicate of chat's implementation); replace call site with `format_source_block`/`group_by_source` from `app.lib.rag`

## How to Test
1. `make test.backend.pytest` — 1115 tests pass
2. `make mypy` — no type errors
3. `make lint.backend` — no lint errors
4. `grep -rn "def format_source_block\|def group_by_source\|_format_context" app/` — only returns definitions in `app/rag/formatting.py`

## Notes
None.


tests/rag/test_formatting.py
```diff
@@ -0,0 +1,87 @@
+"""Unit tests for RAG chunk formatting utilities."""
+
+from app.rag.formatting import format_source_block, group_by_source
+from app.rag.types import RetrievedChunk
```
zaira-bibi: should they import from `app.lib.rag` instead?

zaira-bibi (approved):
just one minor nit, rest is A-OK!

app/rag/formatting.py
```diff
@@ -0,0 +1,31 @@
...
+    for i, chunk in enumerate(chunks, 1):
+        parts = [p for p in [chunk.chapter, chunk.section, chunk.subsection] if p]
+        label = " / ".join(parts) if parts else "General"
+        lines.append(f"--- Excerpt {i} (Section: {label}) ---")
+        lines.append(chunk.content.strip())
+        lines.append("")
+    return "\n".join(lines)
+
+
+def group_by_source(chunks: list[RetrievedChunk]) -> dict[str, list[RetrievedChunk]]:
```
regisb: This function is not very interesting, I don't think it should be part of a library. Do we really need it? Can we take a step back and rethink how RAG consumers make use of it? I don't understand enough of the code to make a suggestion.

app/rag/formatting.py
```diff
@@ -0,0 +1,31 @@
+"""Utilities for formatting RetrievedChunk objects into human-readable context blocks."""
+
+from app.rag.types import RetrievedChunk
+
+
+def format_source_block(source_name: str, chunks: list[RetrievedChunk]) -> str:
```
regisb: This is not a great name for a library function. What's a "source"? What's a "block"? These are generic terms that don't mean much here. Maybe "format_prompt"? or "convert_chunks_to_string"?

abdulrafey1: ok so this might be something deeper that we need to discuss. A `source` is essentially a `document`.

We all understand`document` with ease now since the Document API has been implemented, but these functions predate that. Before that, for RAG implementation we used `source` terminology, so RAG ingested chunks have a `source_name` or a `source`

There could be 2 fixes:
- Rename this function to `format_document_chunk_as_context` and the `source_name` argument to `document_name` and call it a day
OR
- Have a bigger refactor (in a defered PR) where we entirely migrate from the `source` terminology to `document` terminology.

What do you suggest?

Update: I have implemented option 1 for now, if its ok with you we can implement option 2 later on

app/rag/formatting.py
abdulrafey1: So the name and the doc-string might need an update to be more verbose but this is a good utility function.
given a list of chunks (most likely after being retrieved), this function gives the utility to group them based on the document (source) they belong to.

Both the chat and slack plugin use this, I believe this might be handy for future plugins too, let me know if you think its too small of a utility and we are better of letting plugins define this on their own, even it it means a little duplication.

Update: I have updated the function name to `group_retrieved_chunks_by_document`, the argument name to `retrieved_chunks` and have also updated the doc-string, the verbosity makes it a bit easier to understand

hamza-56 (approved):

app/rag/formatting.py
abdulrafey1: @regisb let me know about your thoughts on this ^

PR #459, opened 2026-06-17
chore(test): run local e2e against an ephemeral sqlite database

hamza-56:
## What
Run the local Playwright E2E suite against its own ephemeral SQLite database, created fresh and deleted at the start of every run, instead of the shared dev Postgres, so test runs never mix with development data. Builds on the original idempotent seed, now rebased onto `main` and moved to the async engine.

## Changes
- test: add `scripts/e2e_init_db.py` to build the schema via `SQLModel.metadata.create_all` (migrations can't run on SQLite: pgvector extension + Postgres enums)
- test: add `scripts/e2e-backend.sh` wrapper that deletes `e2e.db`, builds the schema, seeds, then serves uvicorn on :7727
- test(frontend): Playwright `webServer` brings up the ephemeral SQLite backend + frontend (gated on `PLAYWRIGHT_NO_WEBSERVER`, so CI is untouched); `workers: 1` for SQLite's single writer
- refactor(test): seed and schema-init use the async engine (`session_scope` / `run_sync(create_all)`), matching `main`
- fix(test): dispose the async engine so the SQLite scripts don't hang at interpreter shutdown
- chore(test): drop the standalone `make seed-e2e` target (seeding now runs inside the backend); ignore `e2e.db`
- docs(test): fold `frontend/tests/README.md` into a short section in the root `README.md`

## How to Test
1. Install browsers once: `make test.e2e.install`.
2. Start the backing services (Mailpit + Redis; Postgres goes unused): `make services.up`. Stop `make backend.up.dev` if it is running, since the suite owns :7727.
3. `make test.e2e` builds a fresh `e2e.db`, seeds the admin user + `@example.com` whitelist, runs the specs, and tears the backend/frontend down. Confirm the dev Postgres row counts are unchanged afterward.

## Notes
- CI is unchanged: it still runs against Postgres with real migrations (`PLAYWRIGHT_NO_WEBSERVER=1`).
- Limitation: SQLite has no pgvector table, so RAG-ingestion flows can't be exercised in local E2E (the current specs don't need it).
- Addresses this PR's review comments: async engine, throwaway SQLite DB, no separate test README, and the `make seed-e2e` naming (the target is gone).


hamza-56:
@claude-review

abdulrafey1 (approved):

scripts/seed_e2e.py
```diff
@@ -0,0 +1,61 @@
...
+to the values in `frontend/tests/config.ts` and can be overridden via the
+`FIRST_SUPERUSER`, `FIRST_SUPERUSER_EMAIL`, and `FIRST_SUPERUSER_PASSWORD`
+environment variables.
+"""
+
+import os
+
+from sqlmodel import Session, select
+
+from app.core.db import get_engine
```
regisb: I've migrated the app to use async engines exclusively. You'll have to rewrite this.

CLAUDE.md
```diff
@@ -91,6 +91,7 @@ make test.backend.format    # Run backend formatting tests
 make test.frontend          # Run all frontend tests
 make test.frontend.vitest   # Run unit tests with vitest
 make test.frontend.format   # Run frontend formatting tests
+make seed-e2e               # Seed E2E superuser + @example.com whitelist (idempotent; test.e2e runs it automatically)
```
regisb: Target should be "test.e2e.seed".

frontend/tests/README.md
```diff
@@ -32,12 +32,13 @@ sign-up) override `storageState` to start with a fresh context.
```
regisb: Can we use this opportunity to rewrite and simplify this README?

1. We should not have a separate README.
2. Instructions should be simple enough that we could include them in the main README.

Makefile
```diff
@@ -163,8 +163,12 @@ test.frontend.vitest: ## Run frontend unit tests (make test.frontend.vitest [pat
 test.frontend.format: ## Run frontend formatting tests
 	$(MAKE) lint.format.frontend check=1
 
+.PHONY: seed-e2e
+seed-e2e: ## Seed the E2E superuser + @example.com whitelist (idempotent; rerun after services.clean)
+	uv run python scripts/seed_e2e.py
+
 .PHONY: test.e2e
-test.e2e: ## Run Playwright E2E tests (needs the app running + seeded admin; see frontend/tests/README.md)
+test.e2e: seed-e2e ## Run Playwright E2E tests (needs the app running; seeds the admin user first; see frontend/tests/README.md)
```
regisb: I do think that e2e tests should target a different database than the local postgres database. Maybe a throwable sqlite database that would be deleted on every re-run.

scripts/seed_e2e.py
hamza-56: Done. Rebased the branch onto `main` and moved both `seed_e2e.py` and the new `scripts/e2e_init_db.py` onto the async engine: the seed now goes through `session_scope` (mirroring `app/cli/users.py`) and the schema build uses `async_engine` + `run_sync(SQLModel.metadata.create_all)` (mirroring `app/testing.py`).

One gotcha worth flagging for any future async one-shot script on SQLite: the engine has to be disposed before the process exits, otherwise aiosqlite's connection worker thread stays alive and the process hangs at interpreter shutdown (it never hangs on Postgres). The `e2e-backend.sh` wrapper runs these scripts before uvicorn, so that hang blocked the API from starting until I added the dispose (6e0e721).

CLAUDE.md
hamza-56: There's no standalone seed target to rename anymore. Seeding now runs automatically inside the ephemeral E2E backend (`scripts/e2e-backend.sh`), so I dropped `make seed-e2e` entirely. `make test.e2e` builds the throwaway DB, seeds it, and serves, all in one go.

frontend/tests/README.md
hamza-56: Done. Deleted `frontend/tests/README.md` and folded a short "End-to-end tests" section into the root `README.md` (install browsers, `make services.up`, then `make test.e2e` / `make test.e2e.ui`). Repointed the `Makefile` and `CLAUDE.md` references to it.

Makefile
hamza-56: Done, this is the main change in the updated PR. Local `make test.e2e` now runs against its own ephemeral SQLite database (`e2e.db`) that is deleted at the start of every run, so it never reads or writes the dev Postgres data. Playwright spins up a throwaway backend on :7727 (schema built with `create_all`, since the pgvector/enum migrations can't run on SQLite) plus the frontend, then tears both down. CI stays on Postgres with real migrations.

hamza-56:
@claude-review

zaira-bibi (approved):

regisb:
Tests pass in this PR only because the scripts/ directory is not covered by mypy/format tests. Otherwise we would have realised that the `async_engine` import in e2e_init_db.py is no longer functional. But it's strange that unit tests did go through. On my computer unit tests fail:

```
__________________________________________________________________ ERROR collecting tests/scripts/test_e2e_init_db.py __________________________________________________________________
ImportError while importing test module '/home/data/regis/projets/code/work/edly/sparkth/tests/scripts/test_e2e_init_db.py'.
Hint: make sure your test modules/packages have valid Python names.                                                                                                                     Traceback:                                                                                                                                                                              
/home/regis/.pyenv/versions/3.14.2/lib/python3.14/importlib/__init__.py:88: in import_module 
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/scripts/test_e2e_init_db.py:10: in <module>     
    from scripts.e2e_init_db import init_schema                                                                                                                                         scripts/e2e_init_db.py:19: in <module>                                                      
    from app.core.db import async_engine
E   ImportError: cannot import name 'async_engine' from 'app.core.db' (/home/data/regis/projets/code/work/edly/sparkth/app/core/db.py)
```

In addition, we have multiple scripts in scripts/ which seem to all have different naming conventions. Some use dashes others use underscores (e2e-backend.py vs e2e_init_db.py); some use "e2e" as a prefix, others as a suffix. We should harmonize this.

Issue #420, opened 2026-06-09
[EPIC] Permission system needed for fine-grained access control


regisb:
## Problem

As Sparkth evolves toward a new product direction, we need a fine-grained, customizable permission system to control:
- User and group-level access to features (e.g., chat, plugins)
- Action-level permissions (create, edit, delete, manage)
- Page/dashboard access
- Course-level or resource-specific access based on attributes (e.g., "student in course ID 123")

Currently, we only have an `is_superuser` boolean on the User model, which is insufficient for role-based workflows.

### RBAC (Role-Based Access Control)
**Concept:** Users are assigned roles; roles have sets of permissions.
- Example: `alice` → `teacher` role → permissions {`chat.create`, `plugin.manage`}

**Pros:**
- Simple to model and implement
- Scales well — grant/revoke via role membership, not per-user
- Group hierarchies map naturally to role hierarchies
- Intuitive for organizational structures

**Cons:**
- Static permissions per role — difficult to grant exceptions
- Can't express rules like "access to course ID X" without creating roles per course
- Inflexible for attribute-based rules

### ACL (Access Control List)
**Concept:** Direct mapping of who can do what on a specific resource.
- Example: `alice` can `read` `document_123`; `bob` cannot

**Pros:**
- Extremely fine-grained, handles one-off exceptions easily
- No abstraction layer

**Cons:**
- Doesn't scale — hundreds of users × resources = thousands of ACL entries
- Hard to reason about permissions ("what can alice do?")
- No group/hierarchy support without flattening

### ABAC (Attribute-Based Access Control)
**Concept:** Permissions based on attributes of the user, resource, and context.
- Example: "user can access course if `user.role == 'instructor'` AND `course.id IN user.assigned_courses`"

**Pros:**
- Extremely flexible — can express any permission rule
- Handles course/resource-level access naturally
- Scalable — rules, not entries

**Cons:**
- Complex to implement and reason about
- Policy evaluation can be expensive at scale
- Hard to audit ("why can alice see this course?")

## Uncertainty

The product definition isn't finalized. We suspect RBAC may be insufficient for course-level or resource-level access control (e.g., "give instructor alice access to course X", "give student bob access to course Y"). This might require ABAC or a hybrid approach.

## Open Questions

1. Should we invest in a dedicated framework (Casbin, Ory, Permit.io)?
2. Can we get away with a simpler custom implementation?
3. What are the performance and maintenance tradeoffs?

[...]

PR #460, opened 2026-06-18
feat(permissions): add scoped-rbac engine (phase 1, framework only)

abdulrafey1:
## What

Adds the **scoped-RBAC permission engine (Phase 1, framework only)** — the spine of the permission system tracked in #420. It is **purely additive**: no existing endpoint changes behaviour and `is_superuser` remains a fast-path bypass. Correctness is proven by unit tests; nothing is wired into a real route yet (that is a follow-up — see Notes).

## Changes

- `feat(permissions)`: package scaffold — `constants.py` (`SCOPE_GLOBAL`, `ROLE_ADMIN`), `exceptions.py` (`RoleNotFound`)
- `feat(permissions)`: `Role`, `RolePermission`, `RoleAssignment` models — `scope_type`/`scope_id` columns ship now (default `global`); assignments are soft-deleted
- `feat(permissions)`: `PermissionService.can` for the global scope (with the `is_superuser` fast-path)
- `feat(permissions)`: `assign_role` / `revoke_role` (revoke = soft delete)
- `feat(permissions)`: public facade `app/lib/permissions.py`
- `feat(permissions)`: `RequirePermission` FastAPI dependency (callable class — carries permission + scope)
- `feat(migrations)`: `role` / `role_permission` / `role_assignment` tables (`7ee9891441d3`)
- `test(permissions)`: cover `RequirePermission` path-param scope resolution

## How to Test

1. `make services.up && make migrations` — the new migration applies cleanly (single head `7ee9891441d3`, parent `19422defc371`).
2. `uv run pytest tests/permissions/ -v` — 17 tests pass (service + dependency).
3. `make test.backend.pytest` — full backend suite green (1185 passed; nothing regressed).
4. `make mypy` and `make lint.backend` — clean.

## Notes

- **Migration:** adds 3 tables; single head, `down_revision` `19422defc371`, reversible. No edits to existing migrations.
- **Scope:** only the **global** scope is evaluated this phase. `scope_type`/`scope_id` ship now (default `global`) so Phase 2 (scope hierarchy + cascade) needs **no** assignment-table migration.
- **Follow-ups (separate, stacked PRs):**
  1. Migrate the existing `require_superuser`-gated endpoints to `RequirePermission` (+ permission catalogue + admin-role seeding). **This PR is stacked on top of the current one.**
  2. Phase 2 — scope hierarchy / `scope_chain` cascade.
  3. Later — groups; ABAC conditional access.
- **No breaking changes, no new environment variables.**


abdulrafey1:
@claude-review

abdulrafey1:
@claude-review

app/core/permissions/constants.py
```diff
@@ -0,0 +1,4 @@
+SCOPE_GLOBAL = "global"
```
regisb, line 2: What is this variable? It is exposed in app.lib.permissions. What is it supposed to represent?

app/core/permissions/exceptions.py
```diff
@@ -0,0 +1,6 @@
+class RoleNotFound(Exception):
```
regisb, line 1: Shouldn't this inherit from RuntimeError?

app/lib/permissions.py
```diff
@@ -0,0 +1,11 @@
+"""Public API for the permissions (scoped RBAC) framework.
+
+Application code and plugins import the permissions surface from here, never from
+``app.permissions.*`` directly.
+"""
+
+from app.permissions.constants import SCOPE_GLOBAL
+from app.permissions.exceptions import RoleNotFound
+from app.permissions.service import PermissionService
```
regisb: We expose just three objects in app.lib.permissions, but in reality the entire PermissionService object becomes an API. So all three methods become part of the library, though this is not explicit. I don't see why we should expose a service, and not three functions (that would all take an extra "session" argument).

app/core/permissions/exceptions.py
abdulrafey1: RuntimeError is Python's "error that doesn't fall into any other category" bucket the stdlib raises it for things like "dict changed size during iteration" or recursion-limit hits. A "role not found" is a categorizable, expected condition (a lookup miss), not an uncategorized runtime fault. Subclassing RuntimeErrormislabels it

app/core/permissions/constants.py
abdulrafey1: The root scope: a role assigned here applies platform-wide and names no object.

A comment has also been added to this line

abdulrafey1:
@claude-review

app/core/permissions/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+
+from sqlmodel import col, select
+from sqlmodel.ext.asyncio.session import AsyncSession
+
+from app.core.permissions.exceptions import RoleNotFound
+from app.core.permissions.models import Role, RoleAssignment, RolePermission
+from app.models.user import User
+
+
+async def can(
```
zaira-bibi, line 14: this method matches on `scope` and `scope_object_id` and doesn't check the parents (`PermissionScope.get_parents()` is defined but isn't called anywhere).

 but the documentation at `scope.py:10` _("cascade to its descendants")_  and README _("global … applies everywhere")_ describes a cascading system which doesn't exist **yet**. if it's not in this PR's scope, then let's either fix the docs or add a TODO so the readers aren't misled.

app/api/v1/auth.py
```diff
@@ -349,3 +350,40 @@ async def require_superuser(
...
+        current_user: User = Depends(get_current_user),
+        session: AsyncSession = Depends(get_async_session),
+    ) -> User:
+        """Authorize the current user for the bound permission, or raise 403.
+
+        Resolves the scope object id from the request's path params at call time, asks the
+        permission engine whether the user may act, and returns the user so the route
+        receives the authenticated principal. Raises 403 when the permission is absent.
+        """
+        scope_object_id = request.path_params.get(self.scope_param) if self.scope_param else None
```
zaira-bibi, line 400: could we raise 500 here when `self.scope_param` is set but missing from `request.path_params`? right now that case silently collapses to `scope_object_id=None`, which makes the non-global check return 403 to every user. 


```suggestion
        if self.scope_param and self.scope_param not in request.path_params:
    logger.error(
        "RequirePermission(%r) scope_param %r missing from path params %s on %s; "
        "denying all requests",
        self.permission, self.scope_param, list(request.path_params), request.url.path,
    )
    raise ...
scope_object_id = request.path_params.get(self.scope_param) if self.scope_param else None
```

app/core/permissions/scope.py
```diff
@@ -0,0 +1,32 @@
...
+    The plugin registers it from its ``__init__`` with
+    ``PERMISSION_SCOPE.add_item(self, PermissionScope(<scope_name>, parent=<parent_scope_object>))``.
+
+    A scope is a named kind of boundary a role can be assigned at (e.g. ``global``,
+    ``course``, ``quiz``). Each scope may have a single ``parent`` scope, forming a
+    chain from a specific kind up to a broader one; a scope with no parent is a root.
+    The hierarchy lets a role granted at a parent scope cascade to its descendants.
+    """
+
+    def __init__(self, name: str, parent: "PermissionScope" | None = None) -> None:
```
zaira-bibi: `"PermissionScope" | None` evaluates as `str | None` rather than a single forward ref. It's fine at runtime today (3.14 defers annotations) and mypy accepts it, but `typing.get_type_hints(PermissionScope.__init__)` raises `TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'`, so this will break the moment this class is introspected (pydantic field, FastAPI dep, etc.). Quote the whole union instead: `"PermissionScope | None"`.

_Caught by Claude_

zaira-bibi (changes requested):

zaira-bibi (approved):
LGTM!

Issue #462, opened 2026-06-18
[Permission System] Frontend admin UI gates on a coarse admin flag, not the user's permissions


abdulrafey1:
## Problem

The frontend gates admin-only UI on a single coarse admin flag rather than on the user's actual permissions.

With the scoped-RBAC permission system, admin-ness on the frontend is a coarse boolean (`is_admin`, derived from holding the global `admin` role) — returned in the `/user` API response and consumed in:

- `frontend/app/dashboard/admin/whitelist/WhitelistPage.tsx` — blocks the page unless the user is an admin
- `frontend/components/AppSidebar.tsx` and `MobileSidebar.tsx` — show the admin section only for admins
- `frontend/app/dashboard/layout.tsx` — threads the flag down

This does not reflect the backend's granular permissions. A user granted a specific permission (e.g. `email.whitelist.read`) but **not** the full `admin` role cannot see the whitelist admin page, even though the backend would authorize their requests. The frontend's notion of "who sees admin UI" is therefore coarser than — and can diverge from — the backend's actual authorization.

## Why it matters

The point of the granular permission system is that access is per-permission. The frontend should surface a feature's admin UI to exactly the users the backend would authorize for it, so that granting `email.whitelist.read` to a non-admin actually reveals the whitelist page. As more permission-gated features are added, a single coarse admin flag will not scale.

_This issue description was written with the assistance of an LLM (Claude)._

abdulrafey1:
## Proposed approach

Expose the authenticated user's effective permission keys to the frontend and gate UI on specific permissions instead of the coarse admin flag.

- **Backend:** include the user's granted permission keys (e.g. `["email.whitelist.read", ...]`, resolved from their role assignments) in the `/user` response. Consider caching, since it is read on every page load.
- **Frontend:** replace `user.is_admin` checks with a `hasPermission("email.whitelist.read")`-style helper. Gate each admin page/section on the specific permission it needs; show the admin sidebar section when the user has any permission backing a section within it.

This keeps the frontend's visible affordances aligned with what the backend will actually authorize, and scales as new permission-gated features are added.

_This proposed approach was written with the assistance of an LLM (Claude)._

zaira-bibi:
Closed by #558.

PR #463, opened 2026-06-18
feat(permissions): drop is_superuser, derive is_admin from roles

abdulrafey1:
## What

Fully retires `is_superuser`. Admin-ness is now derived from the permission system: `GET /user/me` returns a computed `is_admin` (= the user holds the global `admin` role), and the frontend gates admin UI on that. The `is_superuser` column, model field, schema field, CLI flag behaviour, and every usage are gone — only the permission system authorizes now.

> **Base is `main`.** This branch's original permission foundation (a standalone `app/permissions` package, an `app/lib/permissions.py` facade, a `PermissionService` class, `scope_type`/`scope_id`) was superseded by the scoped-RBAC engine that landed on `main` via #460/#461 (`app/core/permissions`, the `app/lib/permissions` package, free-function API, `scope`/`scope_object_id`). Merging `main` adopts that foundation wholesale; the `is_superuser` retirement is re-expressed on top of it. The net diff vs `main` is exactly that retirement.

## Changes

- `feat(permissions)`: add free-function `has_role(user, role_name, scope, scope_object_id, session)` to the engine, exposed via `app.lib.permissions`
- `feat(api)`: `GET /user/me` returns a computed `is_admin` (schema `is_superuser` → `is_admin`, defaulting `False` so register/login report non-admin)
- `feat(cli)`: `create-user --admin`/`--superuser` assigns the global `admin` role before the single commit (atomic — a missing `admin` role aborts and leaves no orphaned user)
- `feat(frontend)`: gate admin UI (sidebar, whitelist page) on `is_admin`; regenerate the API types
- `feat(migrations)`: `3364c02e8569` drops the `is_superuser` column (downgrade re-adds it `NOT NULL` + `server_default false`); chains onto `main`'s head `b8bf876007e2`, single head
- `test`: cover `has_role`, the `/user/me` `is_admin` derivation, and `create-user --admin`; drop `is_superuser` from fixtures and delete the now-meaningless superuser tests

## How to Test

1. `make services.up && make migrations` — applies `3364c02e8569` (single head; drops the column; reversible).
2. `make test.backend.pytest` — 1227 passed. `make test.frontend.vitest` — 158 passed.
3. `make mypy` / `make lint` / `make test.frontend.format` — clean.
4. **`/user/me`:** a user holding the global `admin` role returns `is_admin: true`; without it, `false`; the `admin` role at a non-global scope does not count.
5. **CLI:** `roles assign-role <user> admin` then `GET /user/me` shows `is_admin: true`. `create-user --admin` assigns the admin role (and exits cleanly, persisting nothing, if the `admin` role hasn't been seeded yet).
6. **Migration (Postgres):** `alembic upgrade head` drops `user.is_superuser`; `alembic downgrade -1` re-adds it; `upgrade head` drops it again. Verified end-to-end.

## Notes

- **Coupled backend + frontend change** (the API field rename forces the frontend rename) — both ship here, no broken admin UI on the branch.
- After this PR `is_superuser` is **fully gone**; the only remaining references are historical migrations (`c697de92eb38`, `b8bf876007e2`) and this migration's `downgrade`.
- **Follow-up (issue #462):** gate frontend admin UI on the user's *specific* permissions rather than the coarse `is_admin` flag.
- No new environment variables. Migration reversible. No existing migration edited.


abdulrafey1:
@claude-review

abdulrafey1:
@claude-review

abdulrafey1:
@claude-review

app/cli/users.py
```diff
@@ -45,15 +46,28 @@ async def _create_user(
...
+        await session.flush()
+        if superuser:
+            assert user.id is not None
+            try:
+                await assign_role(user.id, "admin", "global", None, session)
+            except RoleNotFound:
+                typer.secho(
+                    "Admin role not found — run migrations to seed it before creating an admin user.",
+                    fg=typer.colors.RED,
+                )
```
zaira-bibi, line 64: maybe add a log here as well?

app/api/v1/user.py
```diff
@@ -27,4 +30,6 @@ def get_user(
             detail="No user authenticated.",
         )
 
-    return current_user
+    # "global" is the root scope; admin-ness is membership of the admin role there.
+    is_admin = await has_role(current_user, "admin", "global", None, session)
```
zaira-bibi, line 34: `"global"` is a hard-coded literal here. shouldn't we use `SCOPE_GLOBAL` constant instead?

app/cli/users.py
```diff
@@ -45,15 +46,28 @@ async def _create_user(
...
             email_verified_at=utc_now() if email_verified else None,
         )
         session.add(user)
+        # Flush to obtain user.id, then grant the admin role before the single commit
+        # so a missing admin role aborts the whole operation and leaves no orphaned user.
+        await session.flush()
+        if superuser:
+            assert user.id is not None
+            try:
+                await assign_role(user.id, "admin", "global", None, session)
```
zaira-bibi, line 59: hardcoded literals here as well. let's use constants for maintainability.

tests/permissions/test_dependency.py
zaira-bibi, line 80: let's add a test to verify that a user holding the global `admin` role is still denied when the role hasn't been granted the specific checked permission.

app/core/permissions/__init__.py
```diff
@@ -38,6 +38,30 @@ async def can(
...
+        .join(Role, col(Role.id) == col(RoleAssignment.role_id))
+        .where(
+            RoleAssignment.user_id == user.id,
+            Role.name == role_name,
+            RoleAssignment.is_deleted == False,
+            RoleAssignment.scope == permission_scope,
+            RoleAssignment.scope_object_id == scope_object_id,
+        )
+        .limit(1)
+    )
```
zaira-bibi, line 78: both `has_role()` and `can()` have duplicate query structure so there's a divergence risk on any future scope hierarchy changes.

zaira-bibi (changes requested):

app/api/v1/user.py
abdulrafey1: For now we will keep using the string literals, but i agree that there needs to be a better approach. Its not in the scope of this PR since the decision would have to be a bit more deeper and thoughtful that just using a constant. (Most likely fetching permissions/scopes form relevant registries).

This issue is documented in #469

app/cli/users.py
abdulrafey1: answered above

app/cli/users.py
abdulrafey1: a log for a cli command? 
We already echo to the user about the role not being found, do we need to log it separately?

zaira-bibi: i think a single `logger.warning()` alongside it would cover the case where this command is ever called unattended. optional and not blocking.

zaira-bibi (approved):

PR #464, opened 2026-06-18
refactor(db): injectable engine + drop session_scope test mocks

regisb:
## What
Make the async DB engine an injectable provider and, on that foundation, remove every `session_scope` mock from the test suite so tests exercise the real query path against a fresh in-memory database.

## Changes
- refactor(db): replace the import-time `async_engine` constant with a lazy `get_engine()`/`dispose_engine()` singleton and a single `open_session` provider that `session_scope` delegates to (the one seam tests override)
- refactor(core): per-test harness in `app/testing.py` — migrate the in-memory DB on first session use, dispose the engine per test for isolation; drop the old `SessionFactory`/`get_async_session` override scheme
- test(rag): convert the cleanup/utils/mcp-tools/lib RAG tests from `AsyncMock` session stubs to seeded rows + real-query assertions
- test(slack): drive the `_dispatch_event` tests through the real `session_scope`, asserting on persisted `BotResponseLog` rows
- test(plugins): convert the chat-title and Google Drive folder tests to the real DB (real `ChatService`, seeded `Conversation`/`DriveFolder`/`DriveFile`/`Document`)
- test(cli,services): drop the redundant session-injection patches in the CLI and plugin-service tests

## How to Test
1. `make test.backend.pytest` — full backend suite passes (1168 tests).
2. `make mypy` — strict type-check is clean.
3. `make lint.backend` — ruff is clean.
4. Spot-check isolation: `uv run pytest tests/rag/test_cleanup.py app/core_plugins/slack/tests/test_routes.py -q` — these now seed real rows and assert on DB state.

## Notes
Tests-and-harness plus a small production seam (`app.core.db.open_session`); no schema/migration changes, no new env vars, no dependency bumps. CI runs against SQLite, which the new in-memory-per-test design targets directly. The only remaining `session_scope` patch is the intentional bootstrap guard in `tests/core/test_assemble_app.py` (asserts app assembly never opens a session).

### DIFF
--- a/app/core/db.py
+++ b/app/core/db.py
@@ -1,8 +1,10 @@
-from sqlalchemy.ext.asyncio import create_async_engine
+from collections.abc import AsyncGenerator
+from contextlib import asynccontextmanager
+
+from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
+from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.core.config import get_settings
-
-settings = get_settings()
 
 
 # Helper function to convert database URL to async driver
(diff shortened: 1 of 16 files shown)


abdulrafey1 (approved):
Tried to review this to the best of my abilities 😅 
Looks good!

zaira-bibi (approved):
looks good!

PR #465, opened 2026-06-19
feat(analytics): set up separate analytics database with TimescaleDB

zaira-bibi:
## What
Add a separate analytics database (TimescaleDB) alongside the application database, including a dedicated Alembic migration lineage, async session helpers, and Docker wiring — no analytics tables yet, just the foundation.

## Changes
- **feat(analytics):** new `app/analytics/` package — `db.py` (engine/session providers) and `models.py` (separate `analytics_metadata` registry); engine follows the same lazy-injectable provider pattern as `app/core/db.py` so the test suite can monkeypatch `open_analytics_session` without mocking the engine
- **feat(core):** add `ANALYTICS_DATABASE_URL` setting
- **feat(db):** add `analytics_session_scope` and `get_analytics_session` in `app/lib/db`; delegates to `app.analytics.db.open_analytics_session` (the testable seam)
- **feat(migrations):** consolidate both Alembic lineages under `app/migrations/`: `core/` (app DB, previously `app/migrations/`) and a new `analytics/`.
- **feat(migrations):** initial analytics migration to enable TimescaleDB extension on PostgreSQL
- **chore(docker):** switch Postgres image to `timescale/timescaledb-ha:pg18`; add initdb script to create `sparkth_analytics` logical DB
- **test:** analytics tests live in `tests/analytics/` (not `tests/core/`)
- **docs:** update `CLAUDE.md` and `Makefile` to document two-database setup

## How to Test
1. Pull the branch and run `make services.clean && make services.up` to recreate the Docker volume with the new TimescaleDB image.
2. Run `make migrations` — both `alembic upgrade head` commands should complete without error; the second applies the `enable_timescaledb_extension` migration on `sparkth_analytics`.
3. Run `make db-shell-analytics` and verify the `sparkth_analytics` database exists and `\dx` lists `timescaledb`.
4. Run `make test.backend.pytest` — all tests should pass, including `tests/analytics/test_analytics_db.py` and `tests/analytics/test_analytics_migrations.py`.

## Notes
- New required env var: `ANALYTICS_DATABASE_URL` — already present in `.env` with a working dev default (`postgresql://…/sparkth_analytics`). Production deployments must add it before deploying.
- Docker volume must be wiped on first use (`make services.clean`) because the init script only runs on an empty data directory and the Postgres image changes from `pgvector/pgvector:pg18` to `timescale/timescaledb-ha:pg18`.
- No analytics tables are defined yet — this PR is the foundation phase only.
- `DATABASE_URL` will be renamed to something more specific (e.g. `APP_DATABASE_URL`) in a follow-up; a TODO is left in `.env`.


zaira-bibi:
@claude-review

app/core/db.py
```diff
@@ -15,7 +15,15 @@ def _get_async_database_url(url: str) -> str:
...
-async_engine = create_async_engine(_async_db_url, echo=False, pool_pre_ping=True, **_pool_kwargs)
+def _create_engine(url: str) -> AsyncEngine:
+    """Build an async engine from a database URL, applying a connection pool for non-SQLite backends."""
+    async_url = _get_async_database_url(url)
+    pool_kwargs = dict(pool_size=3, max_overflow=2, pool_recycle=1800) if not async_url.startswith("sqlite") else {}
+    return create_async_engine(async_url, echo=False, pool_pre_ping=True, **pool_kwargs)
+
+
+# App database engine
+async_engine = _create_engine(settings.DATABASE_URL)
```
abdulrafey1: This is something that is being removed in #464
We might need to update the approach here

app/core/config.py
```diff
@@ -10,6 +10,7 @@ class Settings(BaseSettings):
     # still win over both.
     model_config = SettingsConfigDict(env_file=(".env", ".env.local"), env_file_encoding="utf-8", extra="ignore")
     DATABASE_URL: str
```
abdulrafey1, line 12: Not in this PR, but in a later PR we should rename this to `SPARKTH_DATABASE_URL` or `APP_DATABASE_URL` since we have an `ANALYTICS_DATABASE_URL` we cant name this just `DATABASE_URL`, that would be too generic

abdulrafey1:
I dont think `migrations/core` is an accurate name, I believe we can rename it to `app` or `sparkth`, `core` might be misleading because it would imply that these are migrations related to only the app's core, which is not the case since it has migrations for plugins as well.

app/analytics/models.py
abdulrafey1, line 1: Does this file work the same way as `app/models/__init__.py`, If we want alembic migrations to know about a table we have to import it here?

abdulrafey1 (approved):
Everything looks good, just asked one question.
One thing that is worth mentioning is that, should the `analytics` directory be moved into the `core` directory?
Analytics are meant to a part of the core that is shipped with the app. Not sure about it, I think it should be moved, let me know about your thoughts. We can defer it to a later PR

app/analytics/models.py
zaira-bibi: both files have the same end goal, but they work slighly differently. importing the model class in `app/models/__init__.py` is enough because SQLModel models auto-register on the global `SQLModel.metadata` at definition time. but for analytics, we want alembic to target `analytics_metadata` directly (imported from this file). so analytics tables register on it only if they're defined with `metadata=analytics_metadata`. 

that said, the model file still needs to be imported here because Python needs to run it before Alembic can see it in `analytics_metadata`.

zaira-bibi:
> Analytics are meant to a part of the core that is shipped with the app. Not sure about it, I think it should be moved, let me know about your thoughts. We can defer it to a later PR

Definitely. I'll open another PR for this change.

PR #466, opened 2026-06-22
fix(analytics): refactor emission write path

zaira-bibi:
## What
Hardens the analytics write path introduced in #468: introduces a typed schema base class, tightens registry validation, and extends endpoint and test coverage.

> Stacked on #468. GitHub will retarget this to main once that lands.

## Changes
- `feat(analytics)`: add `AnalyticsEvent` base Pydantic model in `schemas/base.py` so event schemas share `event_version` and `occurred_at` without duplication
- `feat(analytics)`: extend registry with `UnknownEventTypeError` and strict unknown-field rejection on validation
- `feat(api)`: extend `POST /api/v1/events` with typed response model and structured 422 error body
- `test(analytics)`: add `tests/analytics/conftest.py` with `analytics_client` fixture
- `test(analytics)`: extend `test_registry` and `test_events_endpoint` coverage

## How to Test
1. `make services.up && make migrations && make backend.up.dev`
2. `POST /api/v1/events` with a valid `user.logged_in` payload — expect `201`
3. `POST /api/v1/events` with an unregistered `event_type` — expect `422` with `detail` field
4. `POST /api/v1/events` with extra unknown fields in `payload` — expect `422`
5. `make test.backend.pytest` — all tests pass

## Notes
None

_This PR description is written by an LLM (Claude)_

PR #468, opened 2026-06-23
feat(analytics): add event emission gateway write path

zaira-bibi:
## What
Adds the analytics write path: a versioned event schema registry, a `raw_events` hypertable, a validating `ingest_event` gateway, and a `POST /api/v1/events` endpoint — with login as the first real producer.

## Changes
- `feat(analytics)`: add versioned Pydantic event schema registry (`app/analytics/registry.py`, `app/analytics/schemas/`)
- `feat(analytics)`: add `raw_events` table and hypertable migration (`app/analytics/models.py`, `app/migrations_analytics/versions/23ceb3cb8918_*`)
- `feat(analytics)`: add `ingest_event` gateway with schema validation (`app/analytics/gateway.py`)
- `feat(api)`: add `POST /api/v1/events` endpoint (`app/api/v1/events.py`)
- `feat(api)`: emit `user.logged_in` v1 event on login, fire-and-forget (`app/api/v1/auth.py`)
- `feat(core)`: expose `ingest_event` / `UnknownEventTypeError` via `app/lib/analytics` public API
- `test(analytics)`: add tests for registry, gateway, raw_events model, events endpoint, login producer, and migration

## How to Test
2. Log in via `POST /api/v1/auth/login` — confirm no error is returned
3. `POST /api/v1/events` with body `{"event_type": "user.logged_in", "event_version": 1, "occurred_at": "<iso8601>", "actor_id": "1", "payload": {"user_id": 1}}` — expect `201`
4. `POST /api/v1/events` with an unregistered `event_type` — expect `422`
5. `make test.backend.pytest` — all tests pass

## Notes
- Requires `make migrations` to create the `raw_events` hypertable (analytics DB migration `23ceb3cb8918`)
- Event types are open strings (not a closed enum) so future producers can register schemas without touching core
- `ingest_event` is fire-and-forget at the call site in auth; a gateway error logs a warning but does not fail the login request

_This PR description is written by Claude_

### DIFF
--- a/app/analytics/models.py
+++ b/app/analytics/models.py
@@ -6,13 +6,30 @@ database and drives the app's Alembic autogenerate). All analytics tables and
 their Alembic migrations target this dedicated :class:`~sqlalchemy.MetaData`
 instead.
 
-This module currently defines only the empty registry — the foundation phase
-adds no tables. TimescaleDB hypertables and continuous aggregates are added in
-later phases.
+This module defines the registry and the first analytics table, ``raw_events``
+(the append-only event store). TimescaleDB hypertable creation and continuous
+aggregates are handled by the analytics migrations / later phases.
 """
 
+import sqlalchemy as sa
 from sqlalchemy import MetaData
+from sqlalchemy.dialects.postgresql import JSONB
 
 # Registry every analytics table attaches to. The analytics Alembic env
-# (app/migrations_analytics/env.py) uses this as its target_metadata.
+# (app/migrations/analytics/env.py) uses this as its target_metadata.
 analytics_metadata = MetaData()
+
+# Append-only, time-partitioned store of validated analytics events. The analytics
+# migration turns this into a TimescaleDB hypertable (partitioned on occurred_at)
+# on PostgreSQL; on SQLite (tests) it stays a plain table. Event-specific fields
+# live in ``payload`` — never as top-level columns — so the table stays generic.
+raw_events = sa.Table(
+    "raw_events",
+    analytics_metadata,
+    sa.Column("occurred_at", sa.DateTime(timezone=True), nullable=False),
+    sa.Column("received_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+    sa.Column("event_type", sa.Text(), nullable=False),
+    sa.Column("event_version", sa.Integer(), nullable=False),
+    sa.Column("actor_id", sa.Text(), nullable=True),
+    sa.Column("payload", sa.JSON().with_variant(JSONB(), "postgresql"), nullable=False),
+)
--- a/app/analytics/registry.py
+++ b/app/analytics/registry.py
@@ -16,10 +16,36 @@ from app.analytics.exceptions import UnknownEventTypeError
 
 
 class EventRegistry:
-    """In-memory registry of versioned event payload schemas."""
+    """In-memory registry of versioned event payload schemas.
+
+    Singleton: ``EventRegistry()`` always returns the same instance, so every
+    import path shares one set of registered schemas. Core default events are
+    registered on first construction; call ``EventRegistry()`` during app
+    assembly (``assemble_app``) to ensure the registry is ready before the
+    first request arrives.
+    """
+
+    _instance: "EventRegistry | None" = None
+
+    def __new__(cls) -> "EventRegistry":
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
 
     def __init__(self) -> None:
-        self._schemas: dict[tuple[str, int], type[BaseModel]] = {}
+        # __init__ runs on every EventRegistry() call, even when __new__ returns
+        # the existing singleton — guard so re-construction never wipes registrations.
+        if not hasattr(self, "_schemas"):
+            self._schemas: dict[tuple[str, int], type[BaseModel]] = {}
+            self._register_defaults()
+
+    def _register_defaults(self) -> None:
+        """Register core events shipped with Sparkth."""
+        # Lazy import avoids a module-level dependency from registry → schemas.
+        from app.analytics.schemas.v1 import AssessmentSubmitted, UserLoggedIn
+
+        self.register("assessment.submitted", 1, AssessmentSubmitted)
+        self.register("user.logged_in", 1, UserLoggedIn)
 
     def register(self, event_type: str, version: int, schema: type[BaseModel]) -> None:
         """Register ``schema`` as the validator for ``(event_type, version)``."""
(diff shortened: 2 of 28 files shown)


zaira-bibi:
@claude-review

app/api/v1/analytics/routes.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    try:
+        await ingest_event(
+            session,
+            event.event_type,
+            event.version,
+            event.payload,
+            str(current_user.id) if current_user.id is not None else None,
+            event.occurred_at,
+        )
```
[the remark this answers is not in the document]
zaira-bibi: Handled in #466

app/api/v1/events.py
```diff
@@ -0,0 +1,63 @@
...
+logger = get_logger(__name__)
+
+router = APIRouter()
+
+
+class EmitEventRequest(BaseModel):
+    event_type: str
+    version: int
+    payload: dict[str, Any]
+    occurred_at: datetime | None = None
```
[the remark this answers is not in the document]
zaira-bibi: Handled in #466

app/analytics/schemas/v1/assessment_submitted.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+
+from pydantic import BaseModel
+
+
+class AssessmentSubmitted(BaseModel):
+    learner_id: str
+    competency_id: str
+    score: float
+    passed: bool
```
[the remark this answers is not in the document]
zaira-bibi: Handled in #466

zaira-bibi:
@claude-review

app/analytics/schemas/v1/__init__.py
abdulrafey1, line 1: lets re-export the models in this init file so they are easier to imoprt

app/analytics/models.py
```diff
@@ -6,13 +6,30 @@
...
 # Registry every analytics table attaches to. The analytics Alembic env
-# (app/migrations_analytics/env.py) uses this as its target_metadata.
+# (app/migrations/analytics/env.py) uses this as its target_metadata.
 analytics_metadata = MetaData()
+
+# Append-only, time-partitioned store of validated analytics events. The analytics
+# migration turns this into a TimescaleDB hypertable (partitioned on occurred_at)
+# on PostgreSQL; on SQLite (tests) it stays a plain table. Event-specific fields
+# live in ``payload`` — never as top-level columns — so the table stays generic.
+raw_events = sa.Table(
```
abdulrafey1, line 26: What does this variable mean, what does it store?

app/analytics/registry.py
```diff
@@ -0,0 +1,41 @@
...
+(plugins) can register their own events without editing core. The string is the
+canonical dot-separated name stored in ``raw_events.event_type``.
+"""
+
+from pydantic import BaseModel
+
+from app.analytics.exceptions import UnknownEventTypeError
+
+
+class EventRegistry:
```
abdulrafey1, line 18: lets make this class a singleton instead of defining `event_registry` below, because somewhere this class could be re-initialized, a singleton class would cater to that.

app/analytics/schemas/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+Importing this package populates :data:`app.analytics.registry.event_registry`
+with every known event schema. Modules that need a populated registry (the
+gateway) import this package for its registration side effect.
+"""
+
+from app.analytics.registry import event_registry
+from app.analytics.schemas.v1.assessment_submitted import AssessmentSubmitted
+from app.analytics.schemas.v1.user_logged_in import UserLoggedIn
+
+event_registry.register("assessment.submitted", 1, AssessmentSubmitted)
```
abdulrafey1: lets register default event (events that are shipped with the core) on the `__init__` method of `EventRegistry` and then the registry should be initialized when the app is being assembled in `main.py`

This way the `EventRegistry` will be ready on app startup

app/api/v1/events.py
```diff
@@ -0,0 +1,63 @@
...
+from app.lib.db import get_analytics_session
+from app.lib.log import get_logger
+from app.models.user import User
+
+logger = get_logger(__name__)
+
+router = APIRouter()
+
+
+class EmitEventRequest(BaseModel):
```
abdulrafey1: lets have an `events` module and add a `schemas.py` and a `routes.py` file to it instead of having everything in one file

app/api/v1/analytics/routes.py
```diff
@@ -0,0 +1,63 @@
...
+
+class EmitEventResponse(BaseModel):
+    accepted: bool
+
+
+@router.post("/", response_model=EmitEventResponse, status_code=status.HTTP_202_ACCEPTED)
+async def emit_event(
+    event: EmitEventRequest,
+    current_user: User = Depends(get_current_user),
+    session: AsyncSession = Depends(get_analytics_session),
```
hamza-56, line 23: Worth pinning down the trust boundary here before anything starts relying on this path.

We authenticate the caller and stamp `actor_id` ourselves (good), but `event_type`, `payload`, and `occurred_at` all come straight from the client for any registered schema. So any logged-in user can POST `assessment.submitted` with someone else's `learner_id`, a score of their choosing, and a backdated `occurred_at`, and it lands in `raw_events` under their own `actor_id`. Schema validation checks the shape, not whether the caller is allowed to assert that payload.

Either direction is fine, it just needs to be a deliberate call:

- gate which `event_type`s this generic route accepts and let trusted server-side producers (like the login hook) emit the rest, or
- treat it as intentional and document that only `actor_id` and `received_at` are server-trusted, so downstream consumers know payload identity claims are unverified.

Not blocking, but since this is the single write path it's the natural place to decide it.

zaira-bibi: PR #466 addresses this by gating which event types the HTTP route accepts.

app/analytics/models.py
zaira-bibi: this is the append-only analytics event store that holds every validated event emitted by the platform.

hamza-56 (approved):

app/api/v1/analytics/routes.py
abdulrafey1, line 1: lets rename the file to `routes.py` to stay consistent with rest of the codebase

abdulrafey1 (commented):
One comment, lets also update any relevant documentation if necessary.

abdulrafey1 (approved):
Everything looks good, just one last thing, why cant we have a class based SQLAlchemy model for the `raw_events`. Why does it need to be plain SQLAlchemy code.
All the other models in our codebase are python classes, why not this one

zaira-bibi:
> Why does it need to be plain SQLAlchemy code. All the other models in our codebase are python classes, why not this one

I didn't make `raw_events` a SQLModel instance as there's no need for ORM features: no identity map, no lazy-loading, no relationships, no `.get()`. It's a pure insert sink. plain SQLAlchemy is lighter and more explicit therefore a better fit for high-throughput append-only tables.

let me know if you think otherwise.

Issue #469, opened 2026-06-23
[Permission System] Misspelled permission or scope names are accepted and fail silently


abdulrafey1:
## Problem

The permission registries materialize the known permission strings (`PermissionsRegistry`)
and the scope tree (`PermissionScopesRegistry`), but the authorization functions never
consult them. `can()` and `assign_role()` (`app/core/permissions/__init__.py`) accept
free-form permission and scope strings and match them directly against the database, so an
unregistered (e.g. misspelled) name is accepted without complaint.

## Where

- `app/core/permissions/__init__.py` — `can()` and `assign_role()`.
- The only database guard is the `ck_role_assignment_scope` CHECK on `role_assignment`,
  which enforces the global↔NULL invariant — **not** that a scope name is registered.

## How to observe

- `assign_role(user_id, role, "typoscope", "x", session)` persists a row fine — the CHECK
  only requires a non-global scope to carry a `scope_object_id`, and nothing checks
  `"typoscope"` against `PermissionScopesRegistry`.
- `RequirePermission("assignmnt.grade", ...)` (note the typo) denies forever: no
  `role_permission` row matches the misspelled string, so the user gets a silent 403 with no
  signal that the *name* is wrong.

## Why it matters

Typos in permission/scope names become silent 403s or undetected bad assignment rows — hard
to diagnose. The behaviour fails *closed* (a typo never grants access), so this is a
robustness/debuggability gap, not a security hole.

## Context

Surfaced in review of the scoped-RBAC engine (#460), part of the permission system (#420).
The registry is intentionally advisory in phase 1 (framework only, nothing wired to real
routes); this can be addressed in the phase that seeds and assigns real roles.

_This issue description was written with the assistance of an LLM (Claude)._

abdulrafey1:
## Proposed fix

Wire the authorization path to validate names against the registries and fail loudly on an
unknown one:

- `assign_role()` — raise `PermissionScopeNotFound` (and optionally validate the permission)
  when the scope/permission isn't registered, before persisting.
- The permission check — validate the requested permission/scope and raise
  `PermissionNotFound` / `PermissionScopeNotFound` rather than returning a plain `False`.

### Constraints to respect

- **Init ordering:** `RequirePermission` instances are constructed during route
  registration, *before* `initialize_*` loads plugin contributions into the registries. So
  validation must run at **request time** (inside `can()`), or as a **post-startup
  consistency check** — not at `RequirePermission` construction, which would false-negative
[...]

PR #471, opened 2026-06-24
refactor(permissions): harden registries and role assignment

abdulrafey1:
## What
Harden the permission registries and role assignment in response to code review: dedupe the registry singletons and fix a race in `assign_role`.

## Changes
- refactor(permissions): extract a `SingletonRegistry` base so the two registries no longer repeat the `instance`/`__new__`/init-once boilerplate
- refactor(permissions): add `reset()` (clear + re-seed defaults) — `clear()` alone left the registry empty, dropping the `global` root and breaking later child-scope registration; mainly for test isolation
- fix(permissions): make `assign_role` race-safe — a concurrent duplicate insert now rolls back in a savepoint and re-queries the winner instead of surfacing an `IntegrityError` as a 500 (stays idempotent)
- refactor(permissions): drop the redundant `session.add` in `revoke_role`'s loop (rows are already session-tracked)
- test(permissions): re-seed the registries on teardown in `test_registry` so the module doesn't leave the shared singletons empty for a later module to trip over (review follow-up)

## How to Test
1. `make test.backend.pytest` — 1256 passed. New tests: registry `reset()` re-seed + child-scope-registers-after-reset, `assign_role` concurrent-insert recovery.
2. `make mypy` / `make lint` / `make test.backend.format` — clean.
3. **Registry:** `clear()` then add a child scope → fails (parent dropped); `reset()` then add a child scope → works (defaults restored).

## Notes
- No migration, no breaking change, no new env var, no dependency bump.
- Reviewer suggestion **5a (swap the permissions list for a set) was deliberately skipped**: the registry holds a handful of strings so O(n) lookup is irrelevant, and a set would drop the insertion order `all()` returns — a behavior change for no real gain.
- Reviewer suggestion **#4 (validate `RequirePermission`'s permission/scope strings) is deferred to #469** — the cleaner home for the check is `can()` itself (the single choke point: return `False` → 403 for a real denial, raise → 500 for an unregistered name). See the design comment on #469.

### DIFF
--- a/app/core/permissions/__init__.py
+++ b/app/core/permissions/__init__.py
@@ -4,6 +4,7 @@ The public surface is re-exported from ``app.lib.permissions``
 """
 
 from sqlalchemy import ColumnElement
+from sqlalchemy.exc import IntegrityError
 from sqlmodel import col, select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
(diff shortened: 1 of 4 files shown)


abdulrafey1:
@claude-review

hamza-56 (approved):

zaira-bibi (approved):

PR #473, opened 2026-06-24
feat(permissions): add role-management API

abdulrafey1:
## What

Add a role-management REST API — role CRUD plus per-role permission grants — under
`/api/v1/permissions`, built on the refactored, hook-sourced permission core.

## Changes

- feat(permissions): declare `role.create/read/update/delete` via `Permission.create()` in `app.core.permissions`, exported from `app.lib.permissions`
- feat(permissions): add the role CRUD + grant engine in `app/core/permissions/roles.py`
- feat(api): add the role-management endpoints in `app/api/v1/permissions/`, gated with `ROLE_*.require_in_global_scope()` and registered on the v1 router
- feat(permissions): add `RoleAlreadyExists` / `RoleInUse` exceptions
- feat(migrations): grant the four `role.*` permissions to the seeded `admin` role
- docs(permissions): document the runtime role-management API and the shipped `role.*` permissions in the README
- chore(frontend): regenerate `frontend/lib/api/generated.ts`

## How to Test

1. `make test.backend.pytest` — full backend suite green (1296 passed). Role coverage: `tests/permissions/test_roles.py` (20) and `tests/api/v1/test_permissions.py` (14).
2. `make mypy` and `make lint.backend` — both clean.
3. End-to-end: as a user holding `role.*` at the global scope, `POST /api/v1/permissions/roles` creates a role and `POST /api/v1/permissions/roles/{id}/permissions` grants a registered permission; a user without `role.*` receives `403`.

## Notes

- **Stacked PR.** Base is `rafey/shift-require-permission` (#478, itself stacked on #477). Review/merge #477 then #478 first; once they land on `main`, this branch restacks onto `main`.
- Migration `86c94c01a092` grants `role.*` to `admin`; its `down_revision` is the current head `3364c02e8569`. No new env var.
- Gating uses `Permission.require_in_global_scope()` — the `RequirePermission` class removed in #478 is not used.

### DIFF
--- a/app/api/v1/permissions/routes/roles.py
+++ b/app/api/v1/permissions/routes/roles.py
@@ -1,0 +1,150 @@
+"""Role sub-router (mounted at ``/roles``): role CRUD and managing a role's permission grants.
+
+Roles are created and edited here by end users with the ``role.*`` permissions. Permissions and
+scopes themselves are NOT manageable through this API — they are declared in code (platform
+defaults or plugin hooks); this API only assigns the already-registered permissions to roles.
+"""
+
+from fastapi import APIRouter, Depends, HTTPException, status
+from sqlmodel.ext.asyncio.session import AsyncSession
+
+from app.api.v1.permissions.schemas import RoleCreate, RolePermissionAdd, RoleResponse, RoleUpdate
+from app.core.permissions import roles as role_engine
+from app.core.permissions.models import Role
+from app.lib.db import get_async_session
+from app.lib.permissions import ROLE_CREATE, ROLE_DELETE, ROLE_READ, ROLE_UPDATE
+from app.lib.permissions.exceptions import PermissionNotFound, RoleAlreadyExists, RoleInUse, RoleNotFound
+
+router = APIRouter()
+
+
+def _role_id(role: Role) -> int:
+    """Return the role's primary key, raising if it is unset.
+
+    A persisted/refreshed role always has an id; a real guard (not a bare ``assert``, which
+    ``python -O`` strips) is what keeps the invariant enforced in optimized builds.
+    """
+    if role.id is None:
+        raise RuntimeError(f"Role has no id: {role!r}")
+    return role.id
+
+
+def _build_role_response(role: Role, permissions: list[str]) -> RoleResponse:
+    """Assemble a RoleResponse from a role and its already-fetched permission grants."""
+    return RoleResponse(id=_role_id(role), name=role.name, description=role.description, permissions=permissions)
+
+
+async def _role_to_response(role: Role, session: AsyncSession) -> RoleResponse:
+    """Build a RoleResponse for one persisted role, fetching its permission grants."""
+    permissions = await role_engine.get_role_permissions(_role_id(role), session)
+    return _build_role_response(role, permissions)
+
+
+@router.get(
+    "",
+    response_model=list[RoleResponse],
+    dependencies=[Depends(ROLE_READ.require_in_global_scope())],
+)
+async def list_roles(session: AsyncSession = Depends(get_async_session)) -> list[RoleResponse]:
+    """List all roles. Returns a list of RoleResponse (id, name, description, permissions).
+
+    Grants are fetched for every role in a single batched query.
+    """
+    all_roles = await role_engine.list_roles(session)
+    permissions_by_role = await role_engine.get_permissions_for_roles([_role_id(role) for role in all_roles], session)
+    return [_build_role_response(role, permissions_by_role.get(_role_id(role), [])) for role in all_roles]
+
+
+@router.post(
+    "",
+    response_model=RoleResponse,
+    status_code=status.HTTP_201_CREATED,
+    dependencies=[Depends(ROLE_CREATE.require_in_global_scope())],
+)
+async def create_role(payload: RoleCreate, session: AsyncSession = Depends(get_async_session)) -> RoleResponse:
+    """Create a role. Returns the created RoleResponse (id, name, description, permissions)."""
+    try:
+        role = await role_engine.create_role(payload.name, payload.description, session)
+    except RoleAlreadyExists as exc:
+        raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(exc)) from None
+    return await _role_to_response(role, session)
+
+
+@router.get(
+    "/{role_id}",
+    response_model=RoleResponse,
+    dependencies=[Depends(ROLE_READ.require_in_global_scope())],
+)
+async def get_role(role_id: int, session: AsyncSession = Depends(get_async_session)) -> RoleResponse:
+    """Fetch a role by id. Returns its RoleResponse (id, name, description, permissions)."""
+    try:
+        role = await role_engine.get_role(role_id, session)
+    except RoleNotFound as exc:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(exc)) from None
+    return await _role_to_response(role, session)
+
+
+@router.patch(
+    "/{role_id}",
+    response_model=RoleResponse,
+    dependencies=[Depends(ROLE_UPDATE.require_in_global_scope())],
+)
+async def update_role(
+    role_id: int, payload: RoleUpdate, session: AsyncSession = Depends(get_async_session)
+) -> RoleResponse:
+    """Update a role's name and/or description. Returns the updated RoleResponse."""
+    try:
+        role = await role_engine.update_role(role_id, payload.name, payload.description, session)
+    except RoleNotFound as exc:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(exc)) from None
+    except RoleAlreadyExists as exc:
+        raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(exc)) from None
+    return await _role_to_response(role, session)
+
+
+@router.delete(
+    "/{role_id}",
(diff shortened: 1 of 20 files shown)


abdulrafey1:
@claude-review

app/api/v1/permissions/routes.py
```diff
@@ -0,0 +1,151 @@
...
+    """Build a RoleResponse for a persisted role, attaching its permission grants."""
+    assert role.id is not None
+    permissions = await get_role_permissions(role.id, session)
+    return RoleResponse(id=role.id, name=role.name, description=role.description, permissions=permissions)
+
+
+@router.get(
+    "/available",
+    response_model=list[str],
+    dependencies=[Depends(RequirePermission("role.read", "global"))],
```
regisb: Why is the permission here "role.read" and not "permission.read"?

app/api/v1/permissions/routes.py
```diff
@@ -0,0 +1,151 @@
...
+        role = await create_role(payload.name, payload.description, session)
+    except RoleAlreadyExists as exc:
+        raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(exc)) from None
+    return await _role_to_response(role, session)
+
+
+@router.get(
+    "/roles/{role_id}",
+    response_model=RoleResponse,
+    dependencies=[Depends(RequirePermission("role.read", "global"))],
```
regisb: Why is the permission here not scoped to the current role? I.e:

```
RequirePermission("role.read", "role", role_id)
```

app/api/v1/permissions/routes.py
```diff
@@ -0,0 +1,151 @@
...
+    permissions = await get_role_permissions(role.id, session)
+    return RoleResponse(id=role.id, name=role.name, description=role.description, permissions=permissions)
+
+
+@router.get(
+    "/available",
+    response_model=list[str],
+    dependencies=[Depends(RequirePermission("role.read", "global"))],
+)
+async def list_available_permissions() -> list[str]:
```
regisb: The names of the different views are a little weird and inconsistent. I suggest:

```
list_roles                    #instead of list_available_permissions
list_permissions       # instead of list_all_roles
get_role                     # instead of get_one_role
create_role                # instead of create_new_role
update_role               # instead of update_existing_role
delete_role_permission # instead of revoke_role_permission
etc.
```

app/api/v1/permissions/routes/roles.py
```diff
@@ -0,0 +1,151 @@
...
+@router.get(
+    "/roles/{role_id}",
+    response_model=RoleResponse,
+    dependencies=[Depends(RequirePermission("role.read", "global"))],
+)
+async def get_one_role(role_id: int, session: AsyncSession = Depends(get_async_session)) -> RoleResponse:
+    """Fetch a role by id. Returns its RoleResponse (id, name, description, permissions)."""
+    try:
+        role = await get_role(role_id, session)
+    except RoleNotFound as exc:
```
regisb, line 82: Every view has to catch the same exceptions the same way. I suggest we create a decorator that catches all these exceptions and returns consistent HTTP 40x errors. Then wrap every view in this decorator and get rid of the exception handling within the views.

EDIT: now that I think of it, isn't there a more fastapi-native way of handling this? For instance with dependencies?

app/api/v1/permissions/routes.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+@router.get(
+    "/roles",
+    response_model=list[RoleResponse],
+    dependencies=[Depends(RequirePermission("role.read", "global"))],
+)
+async def list_all_roles(session: AsyncSession = Depends(get_async_session)) -> list[RoleResponse]:
+    """List all roles. Returns a list of RoleResponse (id, name, description, permissions)."""
+    return [await _role_to_response(role, session) for role in await list_roles(session)]
```
hamza-56: **N+1 query** — `list_all_roles` runs one query to fetch all roles, then calls `_role_to_response` for each role, which fires a separate `get_role_permissions()` SELECT per role. For N roles this is N+1 round-trips to the database.

```python
# current: N+1
return [await _role_to_response(role, session) for role in await list_roles(session)]
```

app/api/v1/permissions/routes.py
```diff
@@ -0,0 +1,151 @@
...
+    remove_role_permission,
+    update_role,
+)
+from app.lib.db import get_async_session
+
+router = APIRouter()
+
+
+async def _role_to_response(role: Role, session: AsyncSession) -> RoleResponse:
+    """Build a RoleResponse for a persisted role, attaching its permission grants."""
```
hamza-56: `assert` is stripped when Python runs in optimized mode (`python -O`), so `assert role.id is not None` becomes a silent no-op in production builds. Use a proper guard instead:

```python
if role.id is None:
    raise RuntimeError(f"Role has no id: {role!r}")
```

Or, since this is always called with a freshly persisted/refreshed `Role`, the check can be removed entirely if the ORM contract guarantees `id` is set after `session.refresh(role)`.

app/api/v1/permissions/routes/roles.py
```diff
@@ -0,0 +1,151 @@
...
+    "/roles/{role_id}",
+    status_code=status.HTTP_204_NO_CONTENT,
+    dependencies=[Depends(RequirePermission("role.delete", "global"))],
+)
+async def delete_existing_role(role_id: int, session: AsyncSession = Depends(get_async_session)) -> None:
+    """Delete a role (refused with 409 while it still has active assignments). Returns 204 No Content."""
+    try:
+        await delete_role(role_id, session)
+    except RoleNotFound as exc:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(exc)) from None
```
hamza-56, line 115: `add_role_permission()` already calls `get_role()` internally (validates and fetches the role). After it returns, this handler calls `get_role(role_id, session)` again to build the response, a redundant DB round-trip.

`add_role_permission` could return the role, or `_role_to_response` could accept only the data it needs. Simplest option: pass the role directly by having `add_role_permission` return it, or refactor `_role_to_response` to take `role_id: int` instead of a `Role` object so the second fetch stays but is clearly intentional.

app/api/v1/permissions/schemas.py
```diff
@@ -0,0 +1,25 @@
...
+
+from pydantic import BaseModel, Field
+
+
+class RoleCreate(BaseModel):
+    name: str = Field(min_length=1, max_length=50)
+    description: str | None = Field(default=None, max_length=255)
+
+
+class RoleUpdate(BaseModel):
```
hamza-56, line 13: With both `name` and `description` optional, a client can `PATCH /roles/{id}` with `{}` or `{"description": null}` and get a 200 with no change made. This burns a DB round-trip and silently succeeds, which may surprise callers.

Consider adding a model validator that rejects a fully-empty update:

```python
from pydantic import model_validator

class RoleUpdate(BaseModel):
    name: str | None = Field(default=None, min_length=1, max_length=50)
    description: str | None = Field(default=None, max_length=255)

    @model_validator(mode="after")
    def at_least_one_field(self) -> "RoleUpdate":
        if self.name is None and self.description is None:
            raise ValueError("at least one of name or description must be provided")
        return self
```

app/api/v1/permissions/routes.py
abdulrafey1: Currently there is only one scope, which is the `GLOBAL` scope. Creating a new scope for roles is out of the scope of this PR.

I've opened #490 (a ROLE scope), as a child of GLOBAL. Once this lands, these endpoints will be re-gated to the appropriate scopes.

app/api/v1/permissions/routes/roles.py
abdulrafey1: How about we define a new hook that stores exceptions and their handlers. Both the `core` and `plugins` can add to that hook and in `main.py`, we can register all exceptions along with their handlers via `app.add_exception_handler(exception, handler)`. This can be done in a private function just like `_register_plugin_routes` and call it in `assemble_app`

LMK what you think about this

app/api/v1/permissions/schemas.py
```diff
@@ -0,0 +1,25 @@
...
+    # Both optional: a field left unset is unchanged (PATCH semantics).
+    name: str | None = Field(default=None, min_length=1, max_length=50)
+    description: str | None = Field(default=None, max_length=255)
```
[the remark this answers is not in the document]
abdulrafey1: We will come back to this if clearing description ever becomes a problem.

app/api/v1/permissions/routes/roles.py
abdulrafey1: This is something that should definitely be addressed, but not in this PR.
#493 addresses the deeper concern, which is the lack of a centralized exception handling system, and once that has been implemented #495 aims at migrating the permissions/roles API to the new centralized exception handling system.

For this PR I suggest we keep the try/except blocks as is

app/api/v1/permissions/routes.py
```diff
@@ -0,0 +1,151 @@
...
+    add_role_permission,
+    available_permissions,
+    create_role,
+    delete_role,
+    get_role,
+    get_role_permissions,
+    list_roles,
+    remove_role_permission,
+    update_role,
+)
```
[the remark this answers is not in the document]
regisb: I think there's a misunderstanding in CLAUDE.md. It's quite OK to import core from other core libraries. What's not OK is to import core from plugins.

regisb (approved):
approved!

PR #474, opened 2026-06-29
refactor(permissions): hook-sourced Permission/PermissionScope objects

abdulrafey1:
## What
Reshape the (not-yet-shipped) scoped-RBAC framework into a self-registering object model: permissions and scope kinds are declared as `Permission` / `PermissionScope` objects collected by name-unique hooks that are the single source of truth — replacing the plain-string + registry-seeding approach.

## Changes
- feat(permissions): add `UniqueCollectionHook` — a flat, name-keyed hook that raises `ValueError` on a duplicate name (`app/lib/hooks.py`)
- refactor(permissions): declare permissions/scopes via `Permission.create()` / `PermissionScope.create()`, which self-register into the `PERMISSIONS` / `PERMISSION_SCOPES` hooks
- refactor(permissions): make the hooks the single source of truth — the registries seed nothing and only read the hooks at startup; delete `defaults.py`
- refactor(permissions): `can` / `has_role` / `assign_role` / `revoke_role` and `RequirePermission` take `Permission` / `PermissionScope` objects (reading `.name`); the whitelist routes pass the declared `EMAIL_WHITELIST_*` + `GLOBAL` objects
- refactor(permissions): rename `scope.py` → `scopes.py`; replace the `SCOPE_GLOBAL` string constant with the `GLOBAL` scope object; delete `constants.py`
- refactor(permissions): house the hooks in `app.core.permissions` and decouple them from `core` (typed via the shared `HasName` bound), so the dependency runs one way (`app.lib.permissions` → `app.core.permissions`) with no circular imports
- docs(permissions): update the README "Permission Management System" section, `PLUGIN_GUIDE.md`, `CLAUDE.md`, and `architectural_patterns.md`
- test(permissions): cover `UniqueCollectionHook`; update the registry / service / dependency / whitelist / user / cli tests

## How to Test
1. `make test.backend.pytest` — full backend suite passes (1255 tests).
2. `make mypy` — `mypy --strict` is clean (322 files). This is the guarantee that every caller was converted to objects: passing a `str` where a `Permission` / `PermissionScope` is now expected is a type error.
3. `make lint.backend` — ruff check + format pass.
4. Behaviour spot-check: declaring two permissions (or scopes) with the same name raises `ValueError` at import; the whitelist endpoints (`GET` / `POST` / `DELETE /api/v1/whitelist`) still authorize via `RequirePermission(EMAIL_WHITELIST_*, GLOBAL)`.

## Notes
- **No migration.** The `RoleAssignment` scope CHECK constraint still emits the literal `global` (now via `GLOBAL.name`), so the database schema is unchanged.
- **No end-user behaviour change** — `can()` authorizes identically; this is an authoring / internal-API reshape of the not-yet-shipped framework.
- **Import-surface changes:** removed `app.core.permissions.constants` (`SCOPE_GLOBAL`) and `app.core.permissions.defaults`. Import `GLOBAL` — and `Permission` / `PermissionScope` / `EMAIL_WHITELIST_*` — from `app.lib.permissions`. The permission/scope hooks now live in `app.core.permissions.hooks` (re-exported from `app.lib.permissions`).
- No new env vars or dependencies.
- **Merged latest `main`** (incl. `#471` registry hardening). Kept this branch's no-seeding object model and dropped `#471`'s `DEFAULT_*`/`reset()` seeding (replaced by the hook-as-single-source model here); **integrated `#471`'s `assign_role` TOCTOU race-fix + `_find_active_assignment`** into `app/core/permissions/utils.py`.


abdulrafey1:
@claude-review

app/lib/hooks.py
```diff
@@ -44,3 +52,22 @@ def iter_items(self) -> Iterator[tuple[SparkthPlugin, T]]:
...
+
+    def __init__(self) -> None:
+        self._items: dict[str, N] = {}
+
+    def add_item(self, item: N) -> None:
+        if item.name in self._items:
+            raise ValueError(f"Duplicate hook item: {item.name}")
+        self._items[item.name] = item
+
+    def iter_items(self) -> Iterator[N]:
```
regisb: nit.: this should be named `iter_values`. `iter_items` would also yield the names.

app/lib/hooks.py
```diff
@@ -44,3 +52,22 @@ def iter_items(self) -> Iterator[tuple[SparkthPlugin, T]]:
         for plugin, plugin_items in self._iter_plugin_items():
             for plugin_item in plugin_items:
                 yield plugin, plugin_item
+
+
+class UniqueCollectionHook(Generic[N]):
```
regisb: In "PluginCollectionHook" the "Collection" part was meant to represent the fact that each plugin was collecting a list of objects. This is not the case here, as each name is associated with a single item. So I suggest to rename "UniqueCollectionHook". For instance, to "SingleNamedItemHook".

app/cli/roles.py
```diff
@@ -4,14 +4,12 @@
 from sqlmodel import select
 
 from app.lib.db import session_scope
-from app.lib.permissions import RoleNotFound
+from app.lib.permissions import GLOBAL, PermissionScope, RoleNotFound
 from app.lib.permissions import assign_role as grant_role
```
regisb, line 10: Just noticed this:

```
from app.lib.permissions import assign_role as grant_role
```

Doesn't this say that we named the function wrong in the first place?

I realize this is not part of this PR, so feel free to ignore.

app/core/permissions/__init__.py
```diff
@@ -1,171 +1,28 @@
...
+from app.core.permissions.hooks import PERMISSIONS
 
-The public surface is re-exported from ``app.lib.permissions``
-"""
 
-from sqlalchemy import ColumnElement
-from sqlalchemy.exc import IntegrityError
-from sqlmodel import col, select
-from sqlmodel.ext.asyncio.session import AsyncSession
+class Permission:
```
regisb, line 4: This is an important class, let's document it.

app/core/permissions/hooks.py
```diff
@@ -0,0 +1,15 @@
+from app.lib.hooks import HasName, UniqueCollectionHook
+
+# The hooks are typed by the shared ``HasName`` bound rather than the concrete
+# ``Permission`` / ``PermissionScope`` classes, so this module stays free of any
+# app.core.permissions import.
```
regisb: This construct is a little weird. If `Permission` and `PERMISSIONS` truly depend on each other, let's just put them both in the same module, no? This would allow us to get rid of this "HasName" layer of indirection.

app/core/permissions/__init__.py
```diff
@@ -1,171 +1,28 @@
...
+    #     ...
 
-    Shared by the read-side checks (``can`` and ``has_role``) so they resolve a scope
-    identically — the single place to change when scope hierarchy lands, which keeps the
-    two from diverging.
 
-    TODO(scope hierarchy): matches the exact (scope, scope_object_id) only — a role granted
-    at a parent scope does not yet cascade to descendants (see PermissionScope.get_parents()).
-    Cascading is the end goal and lands here.
+# Core Permissions shipped with the application.
```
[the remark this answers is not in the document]
abdulrafey1: These comments will be implement in the next PR

app/lib/permissions/registry.py
```diff
@@ -117,19 +101,22 @@ def get(self, name: str) -> PermissionScope:
...
 
 def initialize_permission_scopes_registry() -> None:
-    """Load every permission scope contributed through the PERMISSION_SCOPE hook into the registry."""
+    """Load every permission scope contributed through the PERMISSION_SCOPES hook into the registry."""
     registry = PermissionScopesRegistry()
-    for _, permission_scope in PERMISSION_SCOPE.iter_items():
-        registry.add(permission_scope)
+    for permission_scope in PERMISSION_SCOPES.iter_items():
+        # The hook is typed by the shared HasName bound; every item is in fact a
+        # PermissionScope (added via PermissionScope.create), and the registry needs the
```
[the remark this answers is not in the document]
abdulrafey1: this should not be an issue

tests/permissions/test_registry.py
```diff
@@ -26,54 +24,31 @@ def permission_scopes() -> PermissionScopesRegistry:
 
 
 @pytest.fixture(autouse=True)
-def _reset_registries() -> Generator[None, None, None]:
-    # Clear before each test for an empty baseline; restore the platform defaults afterwards
-    # so this module never leaves the shared singletons empty for a later module to trip over.
+def _reset_registries() -> None:
     PermissionsRegistry().clear()
     PermissionScopesRegistry().clear()
```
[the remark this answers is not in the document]
abdulrafey1: The registries will be modified in the next PR, this will no longer be an issue

scripts/e2e_init_db.py
```diff
@@ -16,7 +16,7 @@
 from sqlalchemy.ext.asyncio import AsyncEngine
 from sqlmodel import SQLModel
 
-from app.core.db import async_engine
+from app.core.db import get_engine
```
regisb: I'm surprised to see this change here. Commented in this other PR: https://github.com/edly-io/sparkth/pull/459#issuecomment-4842269255 Same for the change to seed_e2e.py below.

app/core/permissions/scopes.py
```diff
@@ -30,3 +39,10 @@ def get_parents(self) -> list["PermissionScope"]:
             return []
         grand_parents = self.parent.get_parents()
         return [self.parent, *grand_parents]
+
+
+# Core Permisson Scopes shipped with the application, ordered root-first so each is registered after its parent.
+
+# The root scope: it applies platform-wide and names no object. This is the canonical
+# instance plugins should reference as a parent so the whole tree shares one root.
+GLOBAL = PermissionScope.create("global")
```
regisb, line 58: Why the rename from SCOPE_GLOBAL to GLOBAL? When we import this object (as in `from app.lib.permissions import GLOBAL`) neither the name nor the import path tell us that this is a scope object.

app/api/v1/whitelist.py
```diff
@@ -14,7 +20,7 @@
 @router.get(
     "/",
     response_model=list[WhitelistedEmailResponse],
-    dependencies=[Depends(RequirePermission("email.whitelist.read", "global"))],
+    dependencies=[Depends(RequirePermission(EMAIL_WHITELIST_READ, GLOBAL))],
```
regisb, line 23: I think this tells us that we need a `RequireGlobalPermission` function. That way we would not have to import the GLOBAL object.

app/cli/roles.py
```diff
@@ -32,12 +32,21 @@ async def _assign_role(identifier: str, role: str, scope: str, scope_object_id:
...
         typer.secho("--scope-object-id is required for non-global scopes", fg=typer.colors.RED)
         raise typer.Exit(code=1)
-    if scope == _SCOPE_GLOBAL and scope_object_id is not None:
+    if scope == GLOBAL.name and scope_object_id is not None:
         typer.secho("--scope-object-id is not allowed for the global scope", fg=typer.colors.RED)
         raise typer.Exit(code=1)
+    # Validate the scope kind against the registered vocabulary so a mistyped --scope
+    # fails loudly instead of persisting a no-op assignment.
+    get_plugin_loader()
+    initialize_permission_scopes_registry()
```
regisb, line 47: The fact that we need to explicitly call this function really tells us that we need to get rid of it. The registry only duplicates the data that is container in the PERMISSION_SCOPES hook.

app/core/permissions/models.py
```diff
@@ -44,9 +44,9 @@ class RoleAssignment(TimestampedModel, SoftDeleteModel, table=True):
         # object, every other scope must. Enforced in the database so no code path
         # can persist a contradictory pair.
         CheckConstraint(
-            f"(scope = '{SCOPE_GLOBAL}' AND scope_object_id IS NULL) "
+            f"(scope = '{GLOBAL.name}' AND scope_object_id IS NULL) "
```
regisb: I just noticed this piece of code, though it was not introduced in this PR. This pattern is generally avoided, because it's typical of SQL injection. In this particular case we don't run the risk of SQL injection, because GLOBAL.name is a hardcoded variable. But I'm afraid that this pattern could be replicated elsewhere. Are we OK with keeping this here? I'm not familiar with this pattern so I can't suggest alternatives.

app/lib/permissions/registry.py
```diff
@@ -1,61 +1,38 @@
...
 
 
-class SingletonRegistry:
-    """Base for the permission vocabulary registries.
-
-    Holds one shared instance per subclass and seeds the platform defaults once on
-    construction. Subclasses implement :meth:`reset` to (re)build their storage and
-    seed defaults; the base reuses it for the one-time seed and exposes it publicly.
-    """
+class PermissionsRegistry:
```
regisb, line 18: Do you intend to keep these registries around, remove them here or in a future PR?

app/api/v1/whitelist.py
abdulrafey1: Yup, this will be done in a later PR.

app/lib/permissions/registry.py
abdulrafey1: I intend on making the registries a read only wrapper around the hook. Just like`.create()` methods provide a higher level write access to the hook, the registries will provide a higher level read access to the hook.

This will be done in the next PR.

I still don't believe that fetching permissions or scopes directly from their hooks is a good idea.

WDYS?

app/core/permissions/models.py
abdulrafey1: We could use a string literal here but that would be at risk of drifting over time.
On second thought, maybe this won't really drift over time, it is a small detail, I doubt we will ever change the global scope to something other than `global`.

I will use a string literal here then.

app/cli/roles.py
abdulrafey1: this import alias is because the handler function for the `assign-role` CLI command is also named `assign_role`.
The permission module utility is also named `assign_role`, to avoid the collision we use this alias. We can update the alias to something else or add a comment here mentioning the need for this alias

app/cli/roles.py
abdulrafey1: This wont be the case once the next PR checks in.
We will update the registries to be a read only wrapper around the hooks so we wont have to run two functions to populate them

app/core/permissions/utils.py
```diff
@@ -0,0 +1,173 @@
...
+            RoleAssignment.scope == permission_scope.name,
+            RoleAssignment.scope_object_id == scope_object_id,
+            RoleAssignment.is_deleted == False,
+        )
+        .limit(1)
+    )
+    return (await session.exec(statement)).first()
+
+
+async def can(
```
hamza-56, line 66: Minor, non-blocking: now that `permission` and `permission_scope` are real objects, `role_name` staying a plain string here (and in `assign_role`/`revoke_role`) sticks out a bit. Passing the scope object earns its place via the future `get_parents()` cascade, but roles end up the odd one out. Worth a follow-up to either give roles the same treatment or consciously keep all three as names, just so it's a deliberate choice rather than drift.

abdulrafey1: The roles exist in the DB hence they have to be passed as plain strings, the reason we have classes for `permission` and `permission_scope` is because they are managed by hooks and are meant to be extensible.

app/lib/permissions/registry.py
regisb: OK I agree :)

regisb (approved):
LGTM! Maybe merge this after https://github.com/edly-io/sparkth/pull/476?

PR #477, opened 2026-06-30
refactor(permissions): remove registries in favor of hook lookups

abdulrafey1:
## What
Removes the `PermissionsRegistry` / `PermissionScopesRegistry` wrapper classes and resolves the permission vocabulary directly against the `PERMISSIONS` / `PERMISSION_SCOPES` hooks (the single source of truth) through two get-or-raise helper functions.

## Changes
- feat(core): add `SingleNamedItemHook` and its `.get(name, default=None)` lookup
- refactor(permissions): enforce the "parent must be registered" invariant in `PermissionScope.create()`
- refactor(permissions): re-export the `PERMISSIONS` / `PERMISSION_SCOPES` hooks from a dedicated `app.lib.permissions.hooks` submodule
- refactor(permissions): add `get_permission_or_raise` / `get_permission_scope_or_raise` (raising `PermissionNotFound` / `PermissionScopeNotFound`), exposed via `app.lib.permissions`
- refactor(permissions): delete `PermissionsRegistry` / `PermissionScopesRegistry` and their test; migrate the `assign-role` CLI to `get_permission_scope_or_raise`
- docs(permissions): update the README "Permission Management System" section, the CLAUDE.md public-surface bullet, and the `app/core/permissions` source comments

## How to Test
1. `make test.backend.pytest` — full backend suite passes (1258 passed). New/changed coverage: `tests/lib/test_hooks.py` (hook `.get`), `tests/permissions/test_service.py` (both helpers — patched hooks plus an unpatched assertion against the real `PERMISSIONS`), `tests/permissions/test_scopes.py` (parent-rejection guard).
2. `make mypy && make lint.backend` — both clean.
3. Scope validation still works end-to-end via the CLI: `uv run python -m app.cli assign-role <user> <role> --scope global` succeeds, while `--scope bogus` exits non-zero with "Unknown scope kind" — now resolved through `get_permission_scope_or_raise`, which reads the hook after plugins load.

## Notes
- No DB migration — no schema change.
- Public API: `get_permission_or_raise` / `get_permission_scope_or_raise` and `PermissionNotFound` are now exported from `app.lib.permissions`. `get_permission_or_raise` has no in-tree caller yet — it is added for symmetry with the scope helper and for the role-management API to adopt.
- Supersedes the earlier "read-only registry views" approach on this branch: rather than keep the registries as thin read-only wrappers, they are removed entirely — they held no state of their own and had at most one consumer (the CLI), so plain hook lookups are the right-sized replacement.


abdulrafey1:
@claude-review

app/lib/permissions/registry.py
```diff
@@ -1,117 +1,48 @@
...
+from app.core.permissions import PERMISSIONS, Permission
 from app.core.permissions.exceptions import PermissionNotFound, PermissionScopeNotFound
 from app.core.permissions.scopes import PERMISSION_SCOPES, PermissionScope
-from app.lib.log import get_logger
-
-logger = get_logger(__name__)
+from app.lib.hooks import SingleNamedItemHook
 
 
 class PermissionsRegistry:
```
regisb: What's the point of keeping a class around? In production, this class is always going to use the PERMISSIONS hook. Let's simplify this class and hard-code the hook, which should alway be the same. In unit tests we can mock/patch the hook object.

Same for the PermissionScopesRegistry.

app/lib/hooks.py
```diff
@@ -74,3 +74,7 @@ def iter_values(self) -> Iterator[N]:
 
     def iter_items(self) -> Iterator[tuple[str, N]]:
         yield from self._items.items()
+
+    def get(self, name: str) -> N | None:
```
regisb: nit: In `get` methods it's customary to add a `default=` argument.

README.md
```diff
@@ -249,7 +249,7 @@ Each hook is a `SingleNamedItemHook` keyed by name: declaring two permissions or
...
 
-At startup the app reads both hooks into two singleton registries — `PermissionsRegistry` (a flat list of permission names) and `PermissionScopesRegistry` (a name-indexed tree of scope kinds, where a scope's parent must already be registered). They are the in-memory catalogue of the declared vocabulary, so extending it needs no schema change; the tables above stay the system of record for what is actually *granted* and *assigned*.
+`PermissionsRegistry` and `PermissionScopesRegistry` are read-only views over those hooks — they hold no vocabulary of their own and expose only lookups (`get` returns the registered `Permission` / `PermissionScope` object, raising `PermissionNotFound` / `PermissionScopeNotFound` on a miss; `PermissionsRegistry.all()` returns every `Permission`). There is no separate storage, no singleton, and no startup priming step: the hooks are the catalogue, so extending the vocabulary needs no schema change; the tables above stay the system of record for what is actually *granted* and *assigned*.
```
regisb: I don't think this needs to be part of the README. Actually, the whole section on permissions in the README is really out of place. I mean, it does contain useful information, but its audience is just not the same as the README. I suggest to move this information to docstrings.

zaira-bibi (commented):
agreed with Regis: the the README section has grown to full API-reference depth, which doesn't seem the right place for this level of detail.

README.md
abdulrafey1: I've filed #488 to move the permission-system documentation into docstrings. I'd prefer to keep this PR's README touch minimal and handle the relocation there, so the change stays focused and reviewable.

app/core/permissions/utils.py
```diff
@@ -171,3 +171,19 @@ async def revoke_role(
         # them dirty — no session.add needed.
         assignment.soft_delete()
     await session.flush()
+
+
+def get_permission_or_raise(name: str) -> Permission:
```
regisb: Can we get rid of the "_or_raise" suffix, which is a little awkward? Same for `get_permission_scope_or_raise`. Otherwise Claude is going to create only `*_or_raise` lib functions, as pretty much all our functions can raise exceptions.

regisb (approved):
Just one very minor comment, otherwise we are good to go

tests/permissions/test_service.py
```diff
@@ -305,3 +318,46 @@ async def test_soft_deleted_assignment_does_not_block_reassignment(session: Asyn
...
+
+    with pytest.raises(PermissionScopeNotFound):
+        get_permission_scope_or_raise("course")
+
+
+def test_get_permission_or_raise_reads_the_global_hook() -> None:
+    # Unpatched: resolves against the real PERMISSIONS hook, which carries the core
+    # email-whitelist permissions registered at import. This is the wiring the deleted
+    # test_registry.py used to cover.
+    from app.lib.permissions import EMAIL_WHITELIST_READ
```
hamza-56: This could just live at the top with the other `app.lib.permissions` imports. The local import in `test_facade_exposes_public_surface` is deliberate, but here there is no reason for it.

app/core/permissions/utils.py
```diff
@@ -171,3 +171,19 @@ async def revoke_role(
...
+    """Return the registered permission named ``name``, or raise PermissionNotFound."""
+    permission = PERMISSIONS.get(name)
+    if permission is None:
+        raise PermissionNotFound(name)
+    return permission
+
+
+def get_permission_scope_or_raise(name: str) -> PermissionScope:
```
hamza-56: These two feel a bit out of place here, everything else in this file is an async DB-backed function while these are plain in-memory lookups. They would sit more naturally next to the hook they read: `get_permission_or_raise` in `__init__.py` and `get_permission_scope_or_raise` in `scopes.py`. The latter would also let `PermissionScope.create()` reuse it instead of inlining the same lookup, which it currently only does because `scopes.py` cannot import from `utils.py` without a cycle.

PR #475, opened 2026-06-30
refactor(analytics): self-describing, collision-safe, default-deny registry

zaira-bibi:
## What
Refactors the analytics event schema registry so each schema class carries its own identity (`event_type`, `version`) and emission policy (`server_only`) as `ClassVar`s, enabling single-argument registration and eliminating the risk of the type string drifting from the model. Adds collision detection and default-deny `server_only` as hardening for the plugin extension surface.

## Changes
- `refactor(analytics)`: `AnalyticsEventSchema` gains `event_type`, `version`, `server_only` `ClassVar`s; `server_only` defaults to `True` (default-deny)
- `refactor(analytics)`: `EventRegistry.register(schema)` derives `(event_type, version, server_only)` from the class; idempotent for the same class; raises `DuplicateEventTypeError` on a conflicting class
- `refactor(analytics)`: `resolve` and `is_server_only` return types widened to `type[AnalyticsEventSchema]`; behaviour unchanged
- `feat(analytics)`: `DuplicateEventTypeError` added to `app/core/analytics/exceptions.py` and re-exported from `app/lib/analytics`
- `refactor(analytics)`: `AnalyticsEventSchema` and `DuplicateEventTypeError` added to `app/lib/analytics` public API
- `test(analytics)`: new `test_event_schema_base.py` covering identity `ClassVar`s, default-deny, and Pydantic field exclusion
- `test(analytics)`: `test_registry.py` updated for single-arg registration, idempotency, and collision raise
- `test(analytics)`: `tests/analytics/conftest.py` migrated to the new `register(Class)` API with an `AnalyticsEventSchema` subclass

## How to Test
1. `uv run pytest tests/analytics -v` — all 45 tests pass
2. `make test.backend.pytest` — full backend suite passes (no other callers of the old 4-arg `register` signature)
3. `make mypy` — `Success: no issues found`
4. `make lint.backend && make lint.format.backend check=1` — no lint errors, no files would be reformatted

## Notes
No migrations. No breaking changes to external callers — `ingest_event`, `resolve(event_type, version)`, `is_server_only(event_type, version)`, and the `POST /api/v1/events/` request contract are all unchanged. `server_only` semantics (core events server-only; endpoint returns `403` for them) are preserved. The `server_only = True` default applies only to newly defined schemas; all existing core schemas inherit it and behave identically.

_This PR description was written with the assistance of an LLM (Claude)._

PR #478, opened 2026-07-01
refactor(permissions): move endpoint gating onto Permission.require*

abdulrafey1:
## What

Replace the `RequirePermission` FastAPI dependency class with two methods on the `Permission` object, so a route gates on a permission by calling the permission itself (`EMAIL_WHITELIST_READ.require_in_global_scope()` / `PERM.require("course", "course_id")`).

## Changes

- refactor(auth): relocate `get_current_user` to `app/lib/auth.py` as its single canonical home; every caller (routes, the permission gate, the test harness) imports it from there — no backward-compat re-export from `app.api.v1.auth`
- feat(permissions): add `Permission.require_in_global_scope()` and `Permission.require(scope_name, path_param)`, each returning a FastAPI dependency built by a private closure factory on the class (no dependency class, no separate module)
- refactor(api): gate the whitelist routes via `Permission.require*`
- refactor(permissions): remove the `RequirePermission` class and its now-unused imports
- refactor(permissions): consolidate the RBAC engine (`can`, `has_role`, `assign_role`, `revoke_role`) from `app/core/permissions/utils.py` into `app/core/permissions/__init__.py` and delete `utils.py` — the `require*` gate needs `can` while `utils` imports the `Permission` class at runtime, so keeping them in separate modules cycles; one module removes the cross-module edge. The `get_permission` / `get_permission_scope` lookups keep their `main` placement (beside their respective hooks in `__init__.py` / `scopes.py`); `require()` resolves the scope via `get_permission_scope`
- docs(permissions): rewrite the README endpoint-gating section and add a four-scenario behavior table

## How to Test

1. `make test.backend.pytest` — full backend suite is green (1271 passing). Includes the rewritten `tests/permissions/test_dependency.py`, which covers all four scenarios: authorized, 403 not-granted, `PermissionScopeNotFound` for an unregistered scope name, and 500 for a `scope_param` the route does not provide.
2. `make mypy` and `make lint.backend` — both clean.
3. End-to-end: `uv run pytest tests/api/v1/test_whitelist.py` — the migrated routes still return 403 without the permission and 200/201/204 with it, exercising the new dependency through real FastAPI dependency injection.

## Notes

- **Targets `main`.** This was stacked on `rafey/shifting-require-permission` (#477); now that #477 is merged, it targets `main` directly and has been synced with it (including the analytics move under `app/core/analytics`, #475).
- **No DB migration, no new env var.**
- **Internal breaking changes:**
  - `RequirePermission` is removed — use `Permission.require_in_global_scope()` / `Permission.require(scope_name, path_param)` instead. The README documents the new usage and behavior.
  - `get_current_user` is no longer re-exported from `app.api.v1.auth`; import it from `app.lib.auth`. All in-tree callers are updated in this PR.
  - `app/core/permissions/utils.py` is removed; the RBAC engine (`can`, `has_role`, `assign_role`, `revoke_role`) now lives in `app/core/permissions/__init__.py`. The public surface is unchanged — import everything from `app.lib.permissions` as before.


app/core/permissions/__init__.py
```diff
@@ -16,13 +32,66 @@ def create(cls, name: str) -> "Permission":
...
+
+        Args:
+            permission_scope: A registered scope kind's name (e.g. ``"course"``). Looked up on
+                the PERMISSION_SCOPES hook; an unregistered name raises ``PermissionScopeNotFound``
+                here (at route-definition/import time) so misconfiguration fails fast at startup.
+            scope_param: The name of the route's path parameter carrying the scope object id;
+                resolved from ``request.path_params`` on each request.
+        """
+        scope = PERMISSION_SCOPES.get(permission_scope)
+        if scope is None:
```
regisb: This error handling is already performed by `PermissionScopesRegistry.get`. Should we really bypass the registry here? (or anywhere really)

app/core/permissions/__init__.py
```diff
@@ -16,13 +32,66 @@ def create(cls, name: str) -> "Permission":
...
+            current_user: User = Depends(get_current_user),
+            session: AsyncSession = Depends(get_async_session),
+        ) -> User:
+            if scope_param is not None and scope_param not in request.path_params:
+                logger.error(
+                    "Permission dependency misconfigured: scope_param %r is not among the route's path params %s",
+                    scope_param,
+                    list(request.path_params),
+                )
+                raise HTTPException(
```
regisb: Instead of raising a 500 error, we should just raise a regular exception, which will then be converted to 500 by FastAPI.

app/api/v1/whitelist.py
```diff
@@ -57,7 +55,7 @@ async def add_whitelist_entry(
 @router.delete(
     "/{entry_id}",
     status_code=status.HTTP_204_NO_CONTENT,
-    dependencies=[Depends(RequirePermission(EMAIL_WHITELIST_DELETE, GLOBAL))],
+    dependencies=[Depends(EMAIL_WHITELIST_DELETE.require_in_global_scope())],
```
regisb, line 58: Why is the scope global here? Shouldn't this be scoped to the current whitelist entry?

app/core/permissions/__init__.py
abdulrafey1: Can't use the registry here - it lives in `app/lib` and imports from this module (`app.core.permissions`), so calling it would be a circular import (core can't depend on the `app/lib` layer built on top of it). `PERMISSION_SCOPES` is the source of truth, so we check it directly.

In short the registry is a layer above this thats why we don't use it here.

I can extract the shared `get_or_raise` logic into a helper on the hook if the duplication bugs you.

app/api/v1/whitelist.py
abdulrafey1: Currently there is only one scope, which is the `GLOBAL` scope. Creating a new scope for whitelists is out of the scope of this PR.

I've opened #489 (a `WHITELIST` scope) and #490 (a `ROLE` scope), both as children of `GLOBAL`. Once they land, these endpoints will be re-gated to the appropriate scopes.

app/core/permissions/__init__.py
regisb: Well the issue is more fundamental than that. We are saying that all accesses to PERMISSIONS should happen via the PermissionRegistry. But then the core isn't. Why is that? This issue is telling us that the PermissionRegistry should really be part of the core, in the same module as the Permission class.

app/core/permissions/utils.py
```diff
@@ -3,17 +3,25 @@
...
 from sqlmodel import col, select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
-from app.core.permissions import Permission
 from app.core.permissions.exceptions import RoleNotFound
 from app.core.permissions.models import Role, RoleAssignment, RolePermission
 from app.core.permissions.scopes import PermissionScope
 from app.models.user import User
 
+if TYPE_CHECKING:
```
regisb: Ouch... I'd really like to avoid `TYPE_CHECKING` if possible at all. This points to a circular dependency issue, which must be resolved.

app/core/permissions/__init__.py
abdulrafey1: The registries have been removed in #477

regisb (approved):
lgtm!

app/core/permissions/__init__.py
```diff
@@ -1,4 +1,38 @@
...
+
+These live in one module on purpose: the ``require*`` gate needs ``can`` while the
+engine functions need the ``Permission`` class and the ``PERMISSIONS`` hook, so
+splitting them forces the two modules to import each other at runtime — a cycle.
+
+The public surface is re-exported from ``app.lib.permissions`` — application code
+and plugins import from there, never from this module directly.
+"""
+
+from __future__ import annotations
```
hamza-56: This future import is unnecessary

PR #479, opened 2026-07-01
feat(analytics): plugin-extensible event schemas via ANALYTICS_SCHEMAS hook

### LINKED ISSUE #484: Plugins cannot register analytics event schemas without modifying core
### Summary
Analytics event schemas are defined only in core (`app/core/analytics/schemas/`). A plugin that wants to emit its own event type has no way to contribute a schema without editing core code, and there is nothing enforcing that plugin event types stay namespaced or collision-free.

### Why it matters
- Plugins are meant to be self-contained units (eventually their own repos); requiring a core edit to add a plugin event breaks that boundary and blocks the extensible event vocabulary the analytics epic depends on.
- Without namespace enforcement, two plugins can pick the same event type and collide.
- The permissions system already solves this shape with a registration hook; analytics should mirror it rather than invent a different mechanism.

### Where
- `app/lib/analytics` — no plugin-facing hook for contributing schemas.
- `app/core/analytics/registry.py` — no namespace enforcement for plugin-contributed event types; nothing drains plugin schemas at startup.

### What's needed
- An `ANALYTICS_SCHEMAS` plugin hook (mirroring the permissions pattern) that plugins call from their `__init__`, drained into the registry at app startup, with enforcement that plugin event types are namespaced under the plugin name and do not collide.

zaira-bibi:
## What
Add the `ANALYTICS_SCHEMAS` plugin hook and `initialize_event_registry()` drain so plugins can register their own analytics event schemas without editing core — mirroring the existing `PERMISSIONS` / `PERMISSION_SCOPE` pattern.

## Changes
- feat(analytics): add `EventNamespaceError` for unnamespaced plugin event types
- feat(analytics): convert `app/lib/analytics` from module to package with `hooks.py`
- feat(analytics): add `ANALYTICS_SCHEMAS: PluginCollectionHook` and `initialize_event_registry()` drain
- feat(analytics): wire `initialize_event_registry()` into `assemble_app()` after permissions inits
- test(analytics): add `tests/analytics/test_schema_hook.py` covering hook drain, namespace rejection, and gateway round-trip
- docs: update `CLAUDE.md` and `.claude/docs/architectural_patterns.md` for the new extension surface

## How to Test
1. Run `make test.backend.pytest` — all tests pass (1274/1274).
2. Run `make mypy` — `Success: no issues found`.
3. Run `make lint.backend && make lint.format.backend check=1` — no errors.
4. In a plugin's `__init__`, subclass `AnalyticsEventSchema` with a namespaced `event_type` (e.g. `"myplugin.thing_happened"`), set `version = 1`, register via `ANALYTICS_SCHEMAS.add_item(self, MyEvent)`, and start the app — the event resolves through `EventRegistry` and can be ingested via `ingest_event`.

## Notes
- `app/lib/analytics` is now a package; all prior importers (`from app.lib.analytics import ingest_event`, etc.) are unaffected — the package re-exports the same symbols plus `ANALYTICS_SCHEMAS`, `initialize_event_registry`, `EventRegistry`, `EventNamespaceError`, and `DuplicateEventTypeError`.
- Two startup-fatal guards in the drain (stricter than the permissions drain): namespace enforcement (`EventNamespaceError` if `event_type` is not prefixed with the plugin name) and collision detection (`DuplicateEventTypeError` if two different classes claim the same `(event_type, version)`).
- Plugin events default to `server_only=True` (default-deny); set `server_only = False` to allow client emission via the HTTP gateway.


zaira-bibi:
@claude-review

zaira-bibi:
@claude-review

app/lib/analytics/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+    "DuplicateEventTypeError",
+    "EventNamespaceError",
+    "EventRegistry",
+    "UnknownEventTypeError",
+    "ingest_event",
+    "initialize_event_registry",
+]
+
+
+def initialize_event_registry() -> None:
```
hamza-56: Could this drain live in its own `app/lib/analytics/registry.py` and just be re-exported here? The permissions drain this mirrors sits in `app/lib/permissions/registry.py` with a re-export-only `__init__.py`, so it would keep the two parallel.

app/lib/analytics/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+    Idempotent for the same hook contents: ``EventRegistry.register`` is a no-op
+    for an already-registered class, so calling this more than once (e.g. test
+    setup and a second ``assemble_app``) is safe.
+
+    Plugins must already be instantiated; ``assemble_app`` registers plugin
+    routes before calling this, so the hook is populated.
+    """
+    registry = EventRegistry()
+    for plugin, schema in ANALYTICS_SCHEMAS.iter_items():
+        missing = [attr for attr in ("event_type", "version") if not hasattr(schema, attr)]
```
hamza-56: `EventRegistry.register()` already does this exact check with the same `TypeError` message, so the two copies will drift eventually. Since the point here is just attaching the plugin name to the log, catching the `TypeError` from `register()` (like the `DuplicateEventTypeError` branch below) would avoid the duplication.

.claude/docs/architectural_patterns.md
```diff
@@ -36,6 +37,16 @@ A plugin may also contribute a `CONFIG_ADAPTERS` entry (`app/lib/config/hooks.py
 `LLMConfigAdapter` that pre/post-processes its stored config. `PluginService` resolves it by
 plugin name via `get_plugin_adapter` (`app/lib/config`), so the framework never names a concrete plugin.
 
+A plugin may also contribute analytics event schemas via `ANALYTICS_SCHEMAS`
```
hamza-56: `app/plugins/PLUGIN_GUIDE.md` needs the same update: its list of what a plugin contributes from `__init__` doesn't mention analytics event schemas yet.

zaira-bibi:
@claude-review

app/lib/analytics/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+    "DuplicateEventTypeError",
+    "EventNamespaceError",
+    "EventRegistry",
+    "UnknownEventTypeError",
+    "ingest_event",
+    "register_analytics_event",
+]
+
+
+def register_analytics_event(plugin: SparkthPlugin, schema: type[AnalyticsEventSchema]) -> None:
```
regisb: Shouldn't this be named "register_event_schema"?

app/core/analytics/registry.py
```diff
@@ -37,14 +37,10 @@ def __init__(self) -> None:
         # the existing singleton — guard so re-construction never wipes registrations.
         if not hasattr(self, "_schemas"):
             self._schemas: dict[tuple[str, int], type[AnalyticsEventSchema]] = {}
```
regisb: I thought we had agreed to replace registries by hooks?

app/core/analytics/__init__.py
```diff
@@ -1,6 +1,57 @@
...
+(``app/lib/analytics``), which validates the plugin-name namespace before calling
+``ANALYTICS_EVENTS.add_item``.
 """
+
+from app.core.analytics.exceptions import DuplicateEventTypeError, UnknownEventTypeError
+from app.core.analytics.schemas.base import AnalyticsEventSchema
+from app.core.analytics.schemas.v1 import AssessmentSubmitted, UserLoggedIn
+
+
+class VersionedSchemaHook:
```
abdulrafey1: lets define this hook class in `app/lib/hooks.py` where all the other hook classes are defined.

You can define the `ANALYTICS_EVENTS` hook here or in `app/lib/analytics/hooks.py` where ever you think is better but lets not add this class here.

sparkth/lib/hooks.py
```diff
@@ -78,3 +79,29 @@ def iter_items(self) -> Iterator[tuple[str, N]]:
     def get(self, name: str, default: N | None = None) -> N | None:
         """Return the item registered under ``name``, or ``default`` if none is registered."""
         return self._items.get(name, default)
+
+
+class KeyedItemHook(Generic[K, T]):
```
regisb, line 115: Can I suggest a simpler implementation? A `SingleItemHook(Generic[K, T])` class. Then, `SingleNamedItemHook` would inherit from `SingleItemHook(Generic[str, T])`. And we would have the following implementation of get_event_schema:

```
def get_event_schema(event_type: str, version: int) -> type[AnalyticsEventSchema]:
    """Return the schema for ``(event_type, version)``, or raise UnknownEventTypeError."""
    key = (event_type, version)
    schema = ANALYTICS_EVENTS.get(key)
    if schema is None:
        raise UnknownEventTypeError(event_type, version)
    return schema
```

I feel like this implementation would be more simple and more reusable than the current KeyedItemHook.

app/lib/analytics/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+      name (e.g. plugin ``slack`` → ``"slack.*"``), else ``EventNamespaceError``.
+      This stops a plugin squatting a core or another plugin's event name.
+    - **Collision.** A *different* class claiming an already-registered
+      ``(event_type, version)`` raises ``DuplicateEventTypeError``.
+
+    Idempotent for the same class: ``ANALYTICS_EVENTS.add_item`` is a no-op when the
+    same schema is registered again, so a module re-imported in tests is safe.
+    """
+    missing = [attr for attr in ("event_type", "version") if not hasattr(schema, attr)]
+    if missing:
```
regisb: I don't understand why we need to perform this check here. Type checking should help us catch errors, no?

app/lib/analytics/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+        logger.error(
+            "Plugin '%s' schema %s is missing required identity ClassVars: %s",
+            plugin.name,
+            schema.__qualname__,
+            exc,
+        )
+        raise exc
+    if not schema.event_type.startswith(f"{plugin.name}."):
+        raise EventNamespaceError(plugin.name, schema.event_type)
+    existing = ANALYTICS_EVENTS.get((schema.event_type, schema.version))
```
regisb: Since `add_item` performs the same check, I suggest we write instead:

```
try:
    ANALYTICS_EVENTS.add_item(schema)
except ValueError as e:
    logger.error(...)
    raise DuplicateEventTypeError(schema.event_type, schema.version) from e
```

app/lib/analytics/__init__.py
zaira-bibi: not this one, as `event_type: ClassVar[str]` with no value is an annotation only so mypy treats it as "this attribute has this type" but doesn't require subclasses to assign it.

regisb: Could we simplify the implementation by using `__init_subclass__` to validate the child classes at build time?

sparkth/lib/hooks.py
zaira-bibi: there's one possible issue with a `SingleItemHook(Generic[K, T])` as it would need an `add_item(key, item)`. but at present the `SingleNamedItemHook` just has `add_item(item)` (called by `Permission.create`), so if we inherit it from `SingleItemHook(Generic[K, T])` the child would either need to override with a different signature or hold the key function (not take the key as an `add_item` argument) so the `add_item(item)` stays one-arg on both classes.

there's also one behaviour difference: the `SingleNamedItemHook` raises on any repeated name while the `KeyedItemHook` only raises when a different object already holds the key, so unifying would change this as well.

regisb: > there's one possible issue with a SingleItemHook(Generic[K, T]) as it would need an add_item(key, item)

Good point. So SingleNamedItemHook should not inherit from SingleItemHook.

> the SingleNamedItemHook raises on any repeated name while the KeyedItemHook only raises when a different object already holds the key

I would argue that we should also raise when we attempt to introduce the same object with the same key. It points to an error in the code.

app/lib/analytics/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+    class-definition time, via ``__init_subclass__``):
+
+    - **Namespace.** ``event_type`` must be prefixed with the contributing plugin's
+      name (e.g. plugin ``slack`` → ``"slack.*"``), else ``EventNamespaceError``.
+      This stops a plugin squatting a core or another plugin's event name.
+    - **Collision.** Any class claiming an already-registered ``(event_type, version)``
+      raises ``DuplicateEventTypeError``.
+    """
+    if not schema.event_type.startswith(f"{plugin.name}."):
+        raise EventNamespaceError(plugin.name, schema.event_type)
```
hamza-56: The collision path below logs before raising, but this namespace failure raises silently. A quick `logger.error` here would make a plugin failing startup on namespacing just as easy to diagnose from the logs as a collision is.

README.md
```diff
@@ -460,6 +460,67 @@ the permission identically. Behavior:
...
+```
+
+`register_event_schema` enforces three guards at import time, so a misconfigured plugin fails
+fast at startup instead of at first emit:
+
+- **Namespace** — `event_type` must be prefixed with the plugin's name, else
+  `EventNamespaceError`. This stops a plugin from squatting a core event name or another
+  plugin's namespace.
+- **Collision** — a *different* class claiming an already-registered `(event_type, version)`
+  raises `DuplicateEventTypeError` (registering the *same* class again is a no-op).
```
hamza-56: This parenthetical doesn't match the code anymore: `KeyedItemHook.add_item` raises on any duplicate key, same class or not, and `register_event_schema` turns that into `DuplicateEventTypeError`. There's even a test asserting re-registration is fatal (`test_register_event_schema_rejects_re_registration`). Either restore same-class idempotency or drop this claim. The `DuplicateEventTypeError` message ("A different schema is already registered") also reads wrong for the same-class case.

tests/analytics/test_plugin_registration.py
```diff
@@ -0,0 +1,122 @@
...
+    get_event_schema,
+    ingest_event,
+    register_event_schema,
+)
+from app.lib.analytics.hooks import ANALYTICS_EVENTS
+from app.lib.plugins import SparkthPlugin
+
+
+class _PluginEvent(AnalyticsEventSchema):
+    event_type = "fake-analytics-plugin.thing_happened"
```
hamza-56: `fake-analytics-plugin` / `fake-analytics-plugin.thing_happened` appear a dozen-plus times in this file. Module-level constants (`PLUGIN_NAME`, `EVENT_TYPE = f"{PLUGIN_NAME}.thing_happened"`) would keep the cleanup fixture from silently desyncing if one occurrence ever gets a typo.

tests/analytics/test_plugin_registration.py
```diff
@@ -0,0 +1,122 @@
...
+@pytest.fixture(autouse=True)
+def _cleanup_plugin_registrations() -> Generator[None, None, None]:
+    """Remove the plugin event key from the hook after each test.
+
+    register_event_schema writes straight into the process-wide ANALYTICS_EVENTS
+    hook; without this a later test registering a different class under the same key
+    would hit a spurious DuplicateEventTypeError from leftover state.
+    """
+    yield
+    ANALYTICS_EVENTS._items.pop(("fake-analytics-plugin.thing_happened", 1), None)
```
hamza-56: This reaches into the hook's private `_items` (same in `test_events.py`). Since `KeyedItemHook` is new in this PR, maybe give it a small public removal method for cleanup, like the `clear()` the permissions registries expose, so tests don't depend on the internal dict.

abdulrafey1 (approved):
LGTM

hamza-56 (approved):

Issue #435, opened 2026-06-12
[EPIC] No audit trail: AI actions and security events are unrecorded


regisb:
# Summary

Sparkth records no audit trail of any kind. There is no persistent, queryable record of who did what, when, and with what effect — neither for AI-driven actions (MCP tool invocations, LLM interactions) nor for security-relevant events (authentication, credential changes, plugin configuration).

As Sparkth moves toward supporting residential learning and competency-based education at scale for higher-ed and governmental organizations, this becomes a blocking gap: these customers are subject to regulatory regimes (financial-aid audits, government security baselines, AI-transparency rules) that require exactly this kind of evidence, and procurement processes ask for it explicitly.

# What is not recorded today

**AI actions — the core of the product — are completely opaque:**

- MCP tool invocations are not persisted anywhere. All plugin tools are registered and served through a single path (`app/mcp/main.py`); nothing there records who called which tool, when, with which inputs, or what the outcome was. This includes every state-changing operation against downstream LMSes (`canvas_create_*`, `openedx_create_*`, …). (#417 reworks the registration mechanism but does not change this: tool calls remain unrecorded.)
- Tool calls made by the LLM inside a chat turn are not stored. The chat plugin persists `Conversation` and `Message` rows (`app/core_plugins/chat/models.py`) but not the tool executions, model identity, or model version involved in producing a response.
- RAG ingestion and deletion (`app/rag/`) happen silently.

**Security-relevant events are unlogged:**

- Authentication flows (login, registration, OAuth, email verification — `app/api/v1/auth.py`) leave no audit record: no failed-login history, no record of when or from where an account authenticated.
- Credential lifecycle: stored LLM API keys and plugin credentials can be created, read (decrypted), updated, and deleted with no access trail (`app/api/v1/llm.py`, `app/api/v1/user_plugins.py`).
- There is no request-level audit middleware; mutating REST endpoints across `app/api/v1/` are unrecorded.

The only timestamps in the system are `created_at`/`updated_at` on `TimestampedModel` — current state, not history: previous values are overwritten and there is no record of the actor.

# Why it matters

[...]

Issue #491, opened 2026-07-02
Python library API has no generated reference documentation


regisb:
## Summary

The Python library surface has no generated reference documentation. The
codebase carries ~250 docstrings (Google style, with Sphinx cross-references
such as `:class:` / `:func:` — see `app/lib/db.py`), but nothing renders them
into a browsable API reference. Reference material is currently maintained by
hand in `README.md` and the markdown guides, which duplicates information and
drifts from the code.

## Where it matters

- `app/lib/` is the curated, stable public API that application code **and
  plugins** import from (see #379). Plugin authors need a reference for this
  surface, plus the plugin authoring classes (`SparkthPlugin`, `PluginConfig`,
  …).
- The **REST API** is already covered — FastAPI serves Swagger at `/docs`,
  ReDoc at `/redoc`, and `/openapi.json`. This gap is specifically the
  **Python API reference**, which those do not provide.

## Why it matters

- Docstrings are the source of truth but are invisible to consumers.
- Hand-maintained API listings in markdown violate the project's
  documentation-hygiene rule against duplicating information, and go stale.

## What's needed

A tool that generates a Python API reference from docstrings, scoped to the
public surface (`app/lib/` and the plugin authoring surface), wired into a
`make docs` target and linked from the README instead of pasting API listings
inline.

_This description was written with the assistance of an LLM (Claude)._

regisb:
## Options for generating the reference

All read the existing Google-style docstrings; the choice is mostly about
footprint vs. how much narrative documentation we want alongside the reference.

**pdoc** (leanest) — zero-config, single command:
`uv run --with pdoc pdoc sparkth.lib -o site/api`. Produces a pure API reference site, handles
Google style and reST cross-references out of the box. Best if we only want
reference and nothing else.

**mkdocs + mkdocstrings[python]** — a full docs site combining our existing
markdown guides with generated reference via `::: app.lib.rag`-style autodoc
directives (griffe engine). More setup (`mkdocs.yml`), but gives one published
home for guides + API reference. Best if we want to consolidate all docs.

**Sphinx + autodoc + napalm/napoleon** — the most powerful/customisable, but
RST-oriented and heavier than this project needs. Likely overkill.

### Recommendation

Start with **pdoc** scoped to `app/lib/` (+ plugin authoring surface): smallest
footprint, matches the "avoid bloat" preference, and immediately closes the gap.
Add a `make docs` target and link it from the README. If we later want the
narrative guides and reference under one published site, migrate to
[...]

PR #492, opened 2026-07-03
feat(audit): add append-only audit trail core with login capture

### LINKED ISSUE #500: No core audit event store exists and login attempts leave no audit record
# Summary

Part of epic #435.

Sparkth has no core facility for recording audit events: no append-only system of record, no capture API that application code can call, and no request-scoped origin context (request id, client IP, user agent) to attribute events to an actor. As a first, concrete consequence, authentication attempts leave no persistent record: there is no failed-login history and no record of when or from where an account authenticated.

# Where it happens

- Login and the rest of the auth flows live in `app/api/v1/auth.py`; nothing there persists an audit record on success, failure, or denial.
- There is no audit table in the application database and no audit module under `app/core/` or `app/lib/`.
- Request IP and User-Agent are captured nowhere in the request path; actor identity exists only transiently at the HTTP boundary (`get_current_user`).

# Why it matters

Every other capture task in epic #435 (AI tool executions, credential lifecycle, RAG operations) needs this foundation: a store, a capture API with defined failure semantics, and actor/origin context. NIST SP 800-53 AU-3 requires audit records that answer what, when, where, who, and outcome, and login events are the minimum security-event baseline every compliance regime in the epic's scoping expects. Best-effort analytics (`app/lib/analytics.py`) is explicitly not an audit trail: failed inserts are swallowed by design.

# How to observe

1. `make backend.up.dev`, then log in via the REST API (or fail a login with a wrong password).
2. `make db-shell`: no table records the event. Application logs are stdout-only and unstructured.

hamza-56:
## What

Adds the core of the audit trail (part of #435): an append-only system of record for who did what, when, and with what effect, with login as the first captured event. Tamper-evidence (hash chaining/anchoring) is intentionally not part of this PR; the model already stores the pinned canonical bytes it will need.

Closes #500

## Changes

- feat(audit): add append-only `audit_events` table (NIST AU-3 six fields, mutation diffs, AI provenance columns, reserved `purpose` field, `canonical_bytes`) and Alembic migration
- feat(audit): add `app/lib/audit.py` public facade over `app/core/audit/`: `record_event(session, event)` / `record_event_now(event)` take self-describing event objects (`BaseAuditEvent` subclasses carrying `event_type` / `fail_open` as ClassVars, registered on the module-level `AUDIT_EVENTS` hook), with the envelope grouped into value objects (`AuditTarget`, `AuditChange`, `AuditToolCall`, `AuditModelInfo`)
- feat(audit): model actors as a closed tagged union: `UserActor` (mandatory immutable user id), `SystemActor` (platform-initiated, no id), `AnonymousActor` (`label` is a claimed, untrusted identity), with `AuditActor` as the union alias
- feat(audit): add per-origin audit context classes seeded into a contextvar: `AuditRequestContext` (mandatory request id, client IP (X-Forwarded-For aware), user agent, seeded by pure-ASGI middleware) and `AuditSystemContext` (CLI / platform jobs, request fields structurally absent)
- feat(audit): add secret redaction (auth subtrees, credential key names) applied before persistence
- feat(audit): add pinned canonical JSON serialization stored per event for future integrity sealing
- feat(auth): record `auth.login` success/failure/denied events, fail-closed (no audit record, no token)
- docs: register the new lib module in CLAUDE.md's public library list

## How to Test

1. `make test.backend.pytest` (new suites: `tests/audit/`, `tests/api/v1/test_auth_audit.py`)
2. `make mypy` and `make lint.backend`
3. `make migrations`, then log in via the API and check the row: `make db-shell` then `SELECT category, action, outcome, actor_label, request_ip FROM audit_events;`
4. Try a wrong password and confirm a `failure` row with the attempted username is recorded despite the 401.

## Notes

- New Alembic migration `3fafda1c666e` (app lineage) creates `audit_events`; run `make migrations`.
- Audit writes are deliberately fail-closed for login: if the audit insert fails, the login fails. This is the intended contrast with the best-effort analytics pipeline.
- The API was reshaped during review (event classes + hook instead of a registry and flat kwargs; actor and context unions); the canonical byte encoding and table columns are unchanged throughout.
- The review suggestion to let plugins contribute middlewares via a hook is tracked in #498.
- No new environment variables.


hamza-56:
@claude-review

app/core/audit/context.py
```diff
@@ -0,0 +1,73 @@
...
+from contextvars import ContextVar
+from dataclasses import dataclass
+from typing import Iterator
+
+from app.core.audit.enums import AuditSource
+
+
+@dataclass(frozen=True, slots=True)
+class AuditActor:
+    """Who performed the action: ``type`` is user/system/anonymous."""
```
regisb: What's the difference between these 3?

app/core/audit/context.py
```diff
@@ -0,0 +1,73 @@
...
+
+    type: str
```
regisb: If "type" can be just one of three, why not use an Enum? Or even better: create child classes. For instance, it does not make sense to have an "id" for the "system" actor, right?

app/core/audit/context.py
```diff
@@ -0,0 +1,73 @@
...
+
+@dataclass(slots=True)
+class AuditContext:
+    """Request-scoped origin metadata merged into every recorded event."""
+
+    request_id: str | None = None
+    request_ip: str | None = None
+    user_agent: str | None = None
+    source: AuditSource = AuditSource.SYSTEM
+    actor: AuditActor | None = None
```
regisb: Why are all these fields optional? I would expect them to be mandatory.

app/core/audit/context.py
```diff
@@ -0,0 +1,73 @@
...
+class AuditActor:
+    """Who performed the action: ``type`` is user/system/anonymous."""
+
+    type: str
+    id: str | None = None
+    label: str | None = None
+
+
+@dataclass(slots=True)
+class AuditContext:
```
regisb: The name is odd. Shouldn't it mention the request somehow? E.g: "AuditRequestContext"

app/api/v1/auth.py
```diff
@@ -173,22 +174,44 @@ async def login_for_access_token(
     result = await session.exec(select(User).where(User.username == form_data.username))
     user = result.one_or_none()
 
+    # Audit writes are fail-closed and committed independently of this request
+    # (record_event_now), so a failed attempt stays on record despite the 401.
+    attempted_actor = AuditActor(type="anonymous", label=form_data.username)
```
regisb: How's this an anonymous actor but with a username? And do we really want to tie audit actor to a username, which is a mutable field? Instead we should use the user ID, no?

app/core/audit/recorder.py
```diff
@@ -0,0 +1,160 @@
...
+    old_values: Mapping[str, Any] | None = None,
+    new_values: Mapping[str, Any] | None = None,
+    tool_name: str | None = None,
+    tool_args: Mapping[str, Any] | None = None,
+    model_provider: str | None = None,
+    model_name: str | None = None,
+    model_version: str | None = None,
+    purpose: str | None = None,
+    occurred_at: datetime | None = None,
+) -> AuditEvent:
```
regisb: This function has 16 arguments. I haven't looked at the code but this is a huge red flag for me.

sparkth/core/audit/recorder.py
```diff
@@ -0,0 +1,160 @@
...
+        target_id=target_id,
+        old_values=redacted_old,
+        new_values=redacted_new,
+        tool_name=tool_name,
+        tool_args=redacted_args,
+        model_provider=model_provider,
+        model_name=model_name,
+        model_version=model_version,
+        purpose=purpose,
+        canonical_bytes=canonical_bytes,
```
regisb: 23 arguments. Plus about 15 other objects discretely sent via `canonical_bytes`. I think we need to seriously re-think our API...

sparkth/main.py
```diff
@@ -109,6 +111,9 @@ def assemble_app(lifespan: Lifespan[FastAPI] | None = None) -> FastAPI:
             "/api/v1/auth",
         ],
     )
+    # Added after PluginAccessMiddleware so it wraps it (outermost): the audit
+    # context must exist before any other middleware or handler runs.
+    application.add_middleware(AuditContextMiddleware)
```
regisb, line 117: What this tells us is that we need to create a new hook to which the app and plugins can add their own middlewares, which will then be added at runtime in `assemble_app`.

app/lib/audit.py
```diff
@@ -0,0 +1,36 @@
...
+and with what effect. It is deliberately *not* the analytics pipeline:
+analytics (:mod:`app.lib.analytics`) is best-effort and droppable, audit is
+fail-closed, meaning an event that cannot be written fails the mutating or AI
+action it records.
+"""
+
+from app.core.audit.context import AuditActor, AuditContext, audit_context, bind_audit_actor, current_audit_context
+from app.core.audit.enums import AuditOutcome, AuditSource
+from app.core.audit.exceptions import DuplicateAuditEventTypeError, UnknownAuditEventTypeError
+from app.core.audit.middleware import AuditContextMiddleware
```
regisb: I don't think we need to add this middleware to the library, right?

app/core/audit/registry.py
```diff
@@ -0,0 +1,87 @@
...
+    @property
+    def category(self) -> str:
+        return self.event_type.partition(".")[0]
+
+    @property
+    def action(self) -> str:
+        return self.event_type.partition(".")[2]
+
+
+class AuditEventRegistry:
```
regisb: Oh come on, you created yet another "registry" class which is nothing but an empty shell with a singleton and that would work better as a simple hook?

sparkth/core/audit/models.py
```diff
@@ -0,0 +1,82 @@
...
+
+from app.models.base import TimestampedModel, utc_now
+
+
+def _json_column() -> Column[Any]:
+    """JSONB on Postgres, plain JSON on the SQLite test database."""
+    return Column(JSON().with_variant(JSONB(), "postgresql"), nullable=True)
+
+
+class AuditEvent(TimestampedModel, SQLModel, table=True):
```
regisb, line 31: Here we have a huge audit event table in which many fields are going to be None. Should we stick with a single table or go with multiple, specialized tables? (Honest question, I don't have the answer) For instance, we could have an LLMToolAuditEvent table. One drawback of having one single, huge table is that whenever we add a new field we need to add it to all row.

app/core/audit/context.py
hamza-56: **user:** an authenticated, known identity. The action is attributed to a real account, so id is set to the user's database ID and label to their username. 

**anonymous:** someone acted, but we don't know (or can't trust) who they are. The canonical case is a failed login: the caller typed a username, but since they failed authentication we can't attribute the event to that account's identity. So id stays empty.

**system:** no human behind the action: the platform itself did it (scheduled jobs, background tasks, CLI maintenance commands, automated cleanup).

app/core/audit/context.py
hamza-56: Went with child classes, plus a small `AuditActorType` enum for the stored value: e79146e.

You're right that an enum alone wouldn't cut it, since the invariants differ per type. So now `UserActor` requires the immutable user `id`, while `SystemActor` and `AnonymousActor` don't even have the field. With mypy strict, building an invalid actor fails at type-check time instead of persisting silently. The recorder and DB columns are unchanged, `AuditActor` lives on as the union alias.

The failed-login site is now `AnonymousActor(label=form_data.username)`, which also answers your other question, `label` there is the *claimed* username, kept as evidence but never used for attribution.

app/core/audit/context.py
hamza-56: I'd keep the name, the request is only one producer of this context. CLI commands, background tasks, and the upcoming MCP/chat seams all carry the same object with their own `source` and no `request_*` fields, so `AuditRequestContext` would make those call sites look like misuse.

The real problem was the docstring saying "Request-scoped", fixed in d00d11e. If the name still bugs you, `AuditOrigin` would be my pick, happy to rename.

app/core/audit/context.py
hamza-56: Because each of them is genuinely unknowable for some unit of work. The middleware builds the context before authentication runs, so `actor` can't exist yet, it's bound later via `bind_audit_actor` and stays `None` for unauthenticated calls (the recorder then records `anonymous`). The `request_*` fields only exist for HTTP: CLI and background tasks have no IP or user agent, and even a real request can lack a `User-Agent` header. `source` is defaulted rather than optional, `system` makes the fallback context correct with no setup and the middleware overrides it with `rest`.

Making them mandatory would force placeholder values into audit rows, and for evidence I'd rather store an honest `None` than a fake. The reworded docstring (d00d11e) now spells out the multi-surface story.

app/api/v1/auth.py
hamza-56: Both fixed in e79146e. Attribution now hangs only off the immutable user ID: `UserActor.id` is mandatory on the success path, and the username rides along purely as a display `label`. For failed logins there's no verified user to point at (the typed username may not even exist), so it's `AnonymousActor(label=<claimed username>)`: the label is evidence of what was attempted, never an attribution, and the class structurally can't carry an `id`.

app/core/audit/recorder.py
hamza-56: Fair, it was the envelope flattened into a signature. Reworked in bb62fd4: events are now self-describing classes (same pattern as the analytics schema registry from #475), so it's `record_event(session, event)`, two arguments. Related fields moved into value objects: `AuditTarget`, `AuditChange`, `AuditToolCall`, `AuditModelInfo`. A login call site is now `record_event_now(LoginAuditEvent(outcome=..., actor=...))`.

sparkth/core/audit/recorder.py
hamza-56: Same fix as above (bb62fd4): the public API now takes one event object. The wide construction that remains is the SQLModel row itself, one internal flatten point in the recorder mapping the event onto the flat table columns; the canonical bytes and columns are unchanged so no migration.

app/core/audit/registry.py
hamza-56: Yeah, guilty. bb62fd4 deletes it: event classes carry their own identity (`event_type`, `fail_open` ClassVars) and register on a module-level `AUDIT_EVENTS` hook, same shape as the permissions hooks. No singleton, no definition objects; `@AUDIT_EVENTS.register` on the class.

app/lib/audit.py
hamza-56: Right, removed in 14063e2. The middleware is wired exactly once in `assemble_app`, and `main.py` already imports core internals for its own assembly (`EventRegistry`, `PLUGIN_ROUTERS`), so it now comes from `app.core.audit.middleware` and the lib surface stays consumer-only.

app/core/audit/context.py
regisb: Maybe add this information as comment?

app/core/audit/context.py
hamza-56: Update: went further in 532a944 and adopted your name. The context is now split per origin, same pattern as the actors: `AuditRequestContext` (with `request_id` now mandatory, seeded at the edge) and `AuditSystemContext` for CLI/platform jobs, with `AuditContext` kept as the union alias. Best of both: the request-shaped class says "request", and non-request call sites don't look like misuse.

app/core/audit/context.py
hamza-56: Follow-up: 532a944 makes this structural instead of conventional. `request_id` is now mandatory on `AuditRequestContext`, and `AuditSystemContext` doesn't have the `request_*` fields at all. What stays optional is only what's genuinely unknowable: `actor` (bound after authentication) and `request_ip`/`user_agent` (a request can lack them).

sparkth/main.py
hamza-56: Agreed. Filed #498 so it doesn't get lost in this PR, with a proposed design in the comments (the interesting part is ordering: the audit context middleware must stay outermost, so hook registrations need an explicit priority rather than list order). Until then the two middlewares stay hand-wired in `assemble_app`.

sparkth/core/audit/models.py
hamza-56: Good question, here's what I can see pulling each way:

**The seal wants one sequence.** The upcoming tamper-evidence chain hashes rows in insert order, so verification needs a single total order. Specialized tables would mean per-table chains, or a shared sequence spine, which starts to look like a single table again.

**Audit queries are cross-category timelines.** "Everything user X did last month", "all denied actions": these cut across categories, and with specialized tables each becomes a UNION that grows with every new table.

**The NULLs are cheaper than they look.** Postgres NULLs cost a bitmap bit, the variable payloads already live in JSONB, and adding a nullable column is metadata-only (no rewrite). Append-only means we'd never backfill old rows anyway.

The per-category shape concern is also partly handled a layer up since bb62fd4: each event class carries only its category's fields, and the wide table is the storage flattening. A middle ground if the AI columns keep growing: a 1:1 satellite table keyed by event id, keeping the spine table for ordering and sealing. Curious what you think, happy to go whichever way.

app/core/audit/context.py
hamza-56: Done in c3125f0: added it as the `AuditActorType` docstring, so the three definitions live where the taxonomy is declared.

app/core/audit/context.py
```diff
@@ -0,0 +1,149 @@
...
+
+from contextlib import contextmanager
+from contextvars import ContextVar
+from dataclasses import dataclass
+from typing import ClassVar, Iterator
+
+from app.core.audit.enums import AuditActorType, AuditSource
+
+
+@dataclass(frozen=True, slots=True)
```
abdulrafey1: Lets move all the dataclasses ot a file named `types.py` to stay consistent with rest of the codebase

If you want to put them in something called `context` then lets make context a module rather than a file, the dataclasses can go to `context/types.py` and the utility functions can go to `context/__init__.py`, this would not break the import structure.

app/core/audit/context.py
```diff
@@ -0,0 +1,149 @@
...
+    an attribution, which is why the class has no ``id``.
+    """
+
+    label: str | None = None
+    id: ClassVar[None] = None
+    type: ClassVar[AuditActorType] = AuditActorType.ANONYMOUS
+
+
+AuditActor = UserActor | SystemActor | AnonymousActor
+"""Who performed the action.
```
abdulrafey1: Wont this be better a multi-line comment above line 59 rather than a doc-string in the middle of the file.

We don't follow this documentation convention anywhere else.

app/core/audit/context.py
```diff
@@ -0,0 +1,149 @@
...
+
+    source: AuditSource = AuditSource.SYSTEM
+    actor: AuditActor | None = None
+    request_id: ClassVar[None] = None
+    request_ip: ClassVar[None] = None
+    user_agent: ClassVar[None] = None
+
+
+AuditContext = AuditRequestContext | AuditSystemContext
+"""Origin metadata for the current unit of work, merged into every recorded
```
abdulrafey1: Same comment as above

app/core/audit/events.py
```diff
@@ -0,0 +1,156 @@
...
+from datetime import datetime
+from typing import Any, ClassVar, TypeVar
+
+from app.core.audit.context import AuditActor
+from app.core.audit.enums import AuditOutcome
+from app.core.audit.exceptions import DuplicateAuditEventTypeError, UnknownAuditEventTypeError
+
+
+@dataclass(frozen=True, slots=True)
+class AuditTarget:
```
abdulrafey1: Same comment as above, lets move the dataclasses to a file named `types.py`

app/core/audit/events.py
```diff
@@ -0,0 +1,156 @@
...
+
+    @property
+    def action(self) -> str:
+        return self.event_type.partition(".")[2]
+
+
+E = TypeVar("E", bound=BaseAuditEvent)
+
+
+class AuditEventTypeHook:
```
abdulrafey1: lets move this hook class to `app/lib/hooks.py` since all hook classes exist there.

app/core/audit/events.py
```diff
@@ -0,0 +1,156 @@
...
+    """Flat hook mapping event types to their event classes.
+
+    Like :class:`app.lib.hooks.SingleNamedItemHook` but keyed by
```
abdulrafey1: Lets not refer to other classes in any doc-strings.
If those classes change, this information will get outdated and we are at risk of a drift in the codebase

app/core/audit/events.py
```diff
@@ -0,0 +1,156 @@
...
+
+        Raises:
+            UnknownAuditEventTypeError: The class (or its event type) is not
+                registered; recording it is a programming error.
+        """
+        if self._types.get(event_cls.event_type) is not event_cls:
+            raise UnknownAuditEventTypeError(event_cls.event_type)
+
+
+AUDIT_EVENTS = AuditEventTypeHook()
```
abdulrafey1: lets type hint this hook to stay consistent with hook definitions throughout the codebase

sparkth/core/audit/recorder.py
```diff
@@ -0,0 +1,112 @@
...
+        model_version=model.version if model is not None else None,
+        purpose=event.purpose,
+        canonical_bytes=canonical_bytes,
+    )
+    session.add(row)
+    await session.flush()
+    return row
+
+
+async def record_event_now(event: BaseAuditEvent) -> AuditEvent:
```
abdulrafey1, line 43: this function could be named better, if the only difference between this function and `record_event` is committing, then a better name would be `record_event_and_commit`.

app/core/audit/recorder.py
```diff
@@ -0,0 +1,112 @@
...
+
+from app.core.audit.canonical import canonicalize
+from app.core.audit.context import AnonymousActor, current_audit_context
+from app.core.audit.events import AUDIT_EVENTS, BaseAuditEvent
+from app.core.audit.redaction import redact
+from app.lib.db import session_scope
+from app.models.audit import AuditEvent
+from app.models.base import utc_now
+
+_ANONYMOUS = AnonymousActor()
```
abdulrafey1: lets put this in a file named `constants.py`, it would make things much easier to navigate.

If this is a private constant to this module, we can put it in `constants.py` import it here just not re-export it via the public library so its never imported outside of this module.

app/core/audit/redaction.py
```diff
@@ -0,0 +1,70 @@
...
+with an API token; OpenEdx passes username/password and OAuth tokens), so every
+payload column (``tool_args``, ``old_values``, ``new_values``) is redacted here
+*before* canonicalization and persistence, so unredacted secrets never touch
+the database or the canonical bytes.
+"""
+
+from collections.abc import Mapping
+from typing import Any
+
+REDACTED = "[REDACTED]"
```
abdulrafey1: The constants in this file should also be moved to `constants.py` for better organization

app/lib/audit.py
```diff
@@ -0,0 +1,65 @@
...
+from app.core.audit.events import (
+    AUDIT_EVENTS,
+    AuditChange,
+    AuditModelInfo,
+    AuditTarget,
+    AuditToolCall,
+    BaseAuditEvent,
+    LoginAuditEvent,
+)
+from app.core.audit.exceptions import DuplicateAuditEventTypeError, UnknownAuditEventTypeError
```
abdulrafey1: lets make audit a module instead of a file, in that module, we can have these re-exports under `app/lib/audit/exceptions.py` this would make the import path more verbose.

app/lib/audit.py
```diff
@@ -0,0 +1,65 @@
...
+    AuditSystemContext,
+    SystemActor,
+    UserActor,
+    audit_context,
+    bind_audit_actor,
+    current_audit_context,
+)
+from app.core.audit.enums import AuditActorType, AuditOutcome, AuditSource
+from app.core.audit.events import (
+    AUDIT_EVENTS,
```
abdulrafey1: This hook should be re-exported under `app.lib.audit.hooks`, rest should be under `app.lib.audit.events` and `app.lib.audit.context` etc

app/lib/audit.py
```diff
@@ -0,0 +1,65 @@
...
+    AuditChange,
+    AuditModelInfo,
+    AuditTarget,
+    AuditToolCall,
+    BaseAuditEvent,
+    LoginAuditEvent,
+)
+from app.core.audit.exceptions import DuplicateAuditEventTypeError, UnknownAuditEventTypeError
+from app.core.audit.recorder import record_event, record_event_now
```
abdulrafey1: these can be re-exported under `app/lib/audit/__init__.py`

sparkth/core/audit/models.py
```diff
@@ -0,0 +1,82 @@
+"""The append-only audit event table.
```
abdulrafey1, line 1: this file should really be at `app/core/audit/models.py`

sparkth/main.py
abdulrafey1: +1

sparkth/core/audit/models.py
```diff
@@ -0,0 +1,82 @@
...
+    ``canonical_bytes`` holds the pinned canonical serialization of the event
+    (see :mod:`app.core.audit.canonical`) so the tamper-evidence sealer can
+    later hash exactly what was written, byte for byte.
+    """
+
+    __tablename__ = "audit_events"
+
+    # Column names starting with "model_" are intentional (AI provenance);
+    # disable pydantic's protected namespace check for them.
+    model_config = ConfigDict(protected_namespaces=())  # type: ignore[assignment]
```
abdulrafey1, line 43: lets avoid adding more `# type: ignore` comments to the codebase, we should add a type hint here so mypy doesn't flag this.

sparkth/core/audit/models.py
```diff
@@ -0,0 +1,82 @@
...
+
+    id: int | None = Field(default=None, primary_key=True)
+
+    # What / when / where
+    occurred_at: datetime = Field(
+        sa_type=DateTime(timezone=True),  # type: ignore
```
abdulrafey1, line 49: same comment as above

abdulrafey1 (commented):
Some comments below, one more thing I'd like to add, even though the core audit module is well-documented, lets add a small section in the README that describes the audit modules overall in just 1-2 lines. WDYS?

app/core/audit/context.py
```diff
@@ -0,0 +1,149 @@
...
+class AnonymousActor:
+    """An unauthenticated caller.
+
+    ``label`` carries a *claimed*, untrusted identity (e.g. the username
+    typed into a failed login); it is evidence of what was attempted, never
+    an attribution, which is why the class has no ``id``.
+    """
+
+    label: str | None = None
+    id: ClassVar[None] = None
```
regisb: This contradicts the documentation, that the class has no `id`. Unless I'm mistaken, this class is only ever used with a "label", and this attribute should be absent.

app/core/audit/context.py
```diff
@@ -0,0 +1,149 @@
...
+@dataclass(frozen=True, slots=True)
+class AnonymousActor:
+    """An unauthenticated caller.
+
+    ``label`` carries a *claimed*, untrusted identity (e.g. the username
+    typed into a failed login); it is evidence of what was attempted, never
+    an attribution, which is why the class has no ``id``.
+    """
+
+    label: str | None = None
```
regisb: I don't think we need a "label". In most cases, anonymous events will not have such a label. Think: anonymous users who casually browse the courses will trigger many events, none of which will have a "label". There will be just a few events, such as a failed login, during which there will be a label. In these cases, the event should carry the username, not the actor.

app/core/audit/events.py
regisb: Actually, this class is not really a hook, it just has the "hook" name suffix. It's really just a registry in disguise. I don't understand the benefit, as opposed to a SingleItemHook.

sparkth/core/audit/events.py
```diff
@@ -0,0 +1,156 @@
...
+
+    outcome: AuditOutcome
+    actor: AuditActor | None = None
+    error_detail: str | None = None
+    occurred_at: datetime | None = None
+    target: AuditTarget | None = None
+    change: AuditChange | None = None
+    tool: AuditToolCall | None = None
+    model: AuditModelInfo | None = None
+    purpose: str | None = None
```
regisb: Why so many optional fields?

app/core/audit/context.py
regisb: `request_*` and user_agent fields should be available (and not None) only on AuditRequestContext objects. If `actor` is available only after authentication, then we should build the audit context after authentication, no?

sparkth/core/audit/models.py
```diff
@@ -0,0 +1,82 @@
...
+        default_factory=utc_now,
```
regisb, line 50: Shouldn't we default to the application timezone? Not sure if it's configurable.

sparkth/core/audit/recorder.py
regisb: Still, the AuditEvent class constructor is way too complex. We can't have a class that takes 38 arguments. Why isn't this code all part of the class constructor? Some of the class attributes could then be converted to properties to make sure that the values are always up-to-date and consistent with the attributes.

app/core/audit/events.py
```diff
@@ -0,0 +1,156 @@
...
+    type: str
+    id: str | None = None
+
+
+@dataclass(frozen=True, slots=True)
+class AuditChange:
+    """Before/after snapshots of a mutation; redacted before persistence."""
+
+    old: Mapping[str, Any] | None = None
+    new: Mapping[str, Any] | None = None
```
regisb: 1. Why are these two attributes optional? 
2. Why do we record mutations? Is this a feature that we want to support?

sparkth/core/audit/models.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # Outcome / target
+    outcome: str = Field(max_length=20, nullable=False)
+    error_detail: str | None = Field(default=None, sa_column=Column(Text, nullable=True))
+    target_type: str | None = Field(default=None, max_length=100)
+    target_id: str | None = Field(default=None, max_length=255)
+
+    # Mutation diffs
+    old_values: dict[str, Any] | None = Field(default=None, sa_column=_json_column())
+    new_values: dict[str, Any] | None = Field(default=None, sa_column=_json_column())
```
regisb, line 77: These two names are not very explicit. What's their purpose? And why are they optional?

sparkth/core/audit/models.py
regisb: tl;dr: I think I agree that it's better to go with a single audit event table, but we need to simplify its structure. 

> Specialized tables would mean per-table chains

Yes, I understand. Whatever happens, we should not go with "shared sequence spine", which would be way  too complex.

> "Everything user X did last month", "all denied actions": these cut across categories, and with specialized tables each becomes a UNION that grows with every new table.

Just to get this straight, the different contexts that we need are the following:

1. Request: user made an **authenticated or unauthenticated** HTTP request.
2. CLI: some **unauthenticated** user ran a shell command
3. Async tasks: asynchronous tasks that run in an **unauthenticated** context.

My point is that the use case "Everything user X did last month" will be covered by querying just one audit log 1 (HTTP requests). The other two will not have user ID information.

Do I get this right?

> The NULLs are cheaper than they look.

I'm not concerned about performance, I'm concerned about the data consistency at the code level. With just one AuditEvent table, we will have to check at the code level whether request/user agent attributes are None.

Still, I think it's OK to go with a single table. But at the moment the data structure is too messy: there are too many fields, too many of which are optional. Basically, all fields are nullable. This makes the AuditEvent class a kitchen sink. Any downstream application will have to go through lots of `if/elif/else` to perform any sort of analysis. We really need to rethink the data structure.

sparkth/core/audit/models.py
hamza-56: They're the before/after snapshots of a mutation, the NIST AU-3 "what effect" field, flattened from `AuditChange` after redaction: this is what preserves the previous state that `updated_at`-style columns overwrite, and #503 (credential lifecycle and plugin-config changes) is the first producer that will fill them. They're optional because the pair encodes the mutation kind: a create has no `old`, a delete has no `new`, an update has both, and non-mutation events like logins have neither, so NOT NULL would force placeholder objects into evidence rows.

On the names: `old_values`/`new_values` is the common audit-table convention (Laravel's auditing schema uses exactly these), and they're already pinned as keys in `canonical_bytes`, so I kept them and made the column comment carry this explanation instead: 4e0dc0c.

app/core/audit/events.py
hamza-56: 1. Same encoding as the columns (more detail on the model thread): the pair encodes the mutation kind, a create has no `old`, a delete has no `new`, an update has both, so making either field mandatory would break two of the three shapes. What was genuinely missing is the fourth combination: `AuditChange()` with neither side was constructible and meaningless. 4e0dc0c makes it a ValueError at construction instead of two silent NULLs on an evidence row, and documents the encoding on both the dataclass and the columns.

2. Yes, deliberately: recording mutations is the epic's core "with what effect" requirement, and #503 (credential lifecycle, account flows, plugin configuration changes) is the sub-issue that will produce them. The value object and columns land in this PR so that producer needs no schema change; until it lands, nothing writes them.

sparkth/core/audit/models.py
hamza-56: Deliberately UTC. For audit rows specifically: NIST AU-8 wants audit timestamps mappable to UTC, local time is ambiguous during DST fallback (one local time, two instants), and `occurred_at.isoformat()` is pinned into `canonical_bytes`, so a configurable timezone would make the same event hash differently per deployment and break byte-for-byte verification. Where local time does belong is display: the audit read UI/export (#508) should render in the viewer's timezone.

app/core/audit/events.py
hamza-56: These are Sphinx cross-references, and the repo already uses them outside audit (`core/db`, `core/permissions`, `core/analytics`, `lib/db`, `lib/log`, `lib/settings`), so dropping them here would be the inconsistency. They also age better than prose: once #491 generates the API reference, a stale `:class:` target can be caught by the docs build instead of drifting silently.

app/core/audit/context.py
hamza-56: Done in e6b7820: the actor/context dataclasses and the event value objects now live in `app/core/audit/types.py`, and `context.py` keeps only the contextvar plumbing.

app/core/audit/context.py
hamza-56: Converted to comments above the aliases in e6b7820.

app/core/audit/context.py
hamza-56: Done in e6b7820 (moved to `types.py`).

app/core/audit/events.py
hamza-56: Done in e6b7820: the value objects moved to `app/core/audit/types.py`.

app/core/audit/events.py
hamza-56: Done in 9f25633: the generic mechanics live in `app/lib/hooks.py` as `KeyedClassHook`; `AuditEventTypeHook` stays here as a thin subclass (the format check and audit exceptions) because a full move would have lib importing `app.core.audit`.

app/core/audit/events.py
hamza-56: Annotated in e6b7820.

sparkth/core/audit/recorder.py
hamza-56: One correction on the premise: committing isn't the only difference, it also opens its own session, so the caller's transaction can still roll back while this record survives (the failed-login case). `record_event_and_commit` would read as committing the caller's session; if the current name reads badly I'd take `record_event_isolated`.

app/core/audit/recorder.py
hamza-56: Done in e6b7820: now `ANONYMOUS_ACTOR` in `app/core/audit/constants.py`, kept out of the lib surface.

app/core/audit/redaction.py
hamza-56: Done in e6b7820: `REDACTED`/`MAX_DEPTH`/`SECRET_KEYS` moved to `app/core/audit/constants.py`.

app/lib/audit.py
hamza-56: Done in 8b4d893: `app/lib/audit` is now a package; the exceptions import from `app.lib.audit.exceptions`.

app/lib/audit.py
hamza-56: Done in 8b4d893: `AUDIT_EVENTS` re-exports from `app.lib.audit.hooks`, the event classes and value objects from `app.lib.audit.events`, and the actors/contexts/helpers from `app.lib.audit.context`.

app/lib/audit.py
hamza-56: Done in 8b4d893: `record_event` / `record_event_now` are the `app.lib.audit` re-exports.

sparkth/core/audit/models.py
hamza-56: Done in 9a59e56: the model lives at `app/core/audit/models.py`, with the migrations env importing it for autogenerate the same way it does the permission tables.

sparkth/core/audit/models.py
hamza-56: This one I'd leave: the identical ignore exists in 8 other models including `models/base.py` (`sa_type=DateTime(...)` with SQLModel has no annotation that satisfies mypy --strict), so it's the codebase's current pattern rather than a new debt. Removing it repo-wide deserves its own issue; happy to file it.

sparkth/core/audit/models.py
hamza-56: Covered by the reply on the `model_config` thread above.

hamza-56:
@abdulrafey1 all your layout comments are applied now:

- README section on the audit trail: 0715602
- Hook mechanics moved to `app/lib/hooks.py` as `KeyedClassHook`: 9f25633
- Model co-located at `app/core/audit/models.py`: 9a59e56
- Dataclasses split into `app/core/audit/types.py`, constants into `constants.py`, alias docstrings converted to comments, `AUDIT_EVENTS` annotated: e6b7820
- `app/lib/audit` expanded into a package (`events` / `context` / `hooks` / `exceptions` submodules, write path in `__init__`), consumers and CLAUDE.md updated: 8b4d893

The only two threads where I pushed back instead: the docstring cross-references (the repo already uses them in six-plus modules and #491 consumes them) and the two `type: ignore` comments (same pattern as 8 other models including `models/base.py`; removing it repo-wide deserves its own issue).

app/core/audit/context.py
hamza-56: Docstring wording, fixed in e6b7820: it now says "no per-instance `id`; the ClassVar pins it to `None` for the uniform read surface". The attribute stays because the recorder reads every actor through one shape (`type`/`id`/`label`), and as a ClassVar it's unsettable: `AnonymousActor(id="x")` is a `TypeError`.

app/core/audit/context.py
hamza-56: The structural part is already in (532a944, then e6b7820): `AuditSystemContext` has no `request_*` fields at all, and `request_id` is mandatory on `AuditRequestContext`. `request_ip`/`user_agent` stay nullable on purpose: ASGI allows `client` to be absent and `User-Agent` is an optional header, so I'd rather record an honest `None` than a placeholder.

On building the context after authentication: the events we care most about are the ones where authentication *fails*. A failed login still needs its request id/IP/user agent recorded, and at that point there's no actor and never will be. Hence two-phase: the middleware seeds what the edge knows, and the auth dependency binds the actor via `bind_audit_actor` the moment it exists.

sparkth/core/audit/models.py
abdulrafey1: sure, lets create an issue about it, but still we should avoid adding thing that we already intend on removing.
The similar comments that exist in the codebase are already on death row.

sparkth/core/audit/recorder.py
hamza-56: Agreed, fixed in c971eab, with one adjustment to your suggestion. The assembly now lives on the class as a named constructor, `AuditEvent.from_event(event, context)`: it resolves the actor and timestamp, redacts, pins the canonical bytes, and flattens the value objects onto columns, so a row can't be built with mutually inconsistent fields and nobody spells out 23 kwargs anymore (the recorder is down to registration check + add/flush).

Why not literally the `__init__`: SQLAlchemy bypasses `__init__` when loading rows from the DB, and SQLModel generates the field-based constructor, so overriding it fights the framework. And properties can't replace columns like `category`/`action`: they need to be real stored columns to be indexable, so the factory guarantees consistency at build time and append-only keeps it that way. (Considered a builder too, but it would just be the same 23 fields as setters with mutable intermediate state.)

app/core/audit/context.py
hamza-56: You're right, and 9b9bc51 goes your way: `AnonymousActor` is now a pure marker with no fields at all. The claimed username moved onto the event, as `target=AuditTarget(type="username", id=<typed username>)`, which is also how the Windows Security Log models failed logons (the attempted account is the *target*; the subject stays anonymous). Anonymous browsing events just carry the bare marker. This reverses what I argued earlier in the actor thread; the target framing is simply better.

sparkth/core/audit/events.py
hamza-56: Fixed in 9b9bc51: the base now carries only what every category has (`outcome`, `actor`, `error_detail`, `occurred_at`, `target`), and the rest moved to tiers: `MutationAuditEvent` adds `change`, `AIActionAuditEvent` adds `tool`/`model`/`purpose`. Building a login event with `change=...` is now a `TypeError`. What's left optional on the base is only what's resolved later (`actor`, `occurred_at`) or genuinely absent for some outcomes (`target`, `error_detail`). Storage is unchanged: the tiers carry the shape, the single table stays flat.

hamza-56:
Filed it: #514, with a couple of fix options in a comment.

sparkth/core/audit/models.py
hamza-56: On the kitchen sink: the structure got tightened since this comment. Since 9b9bc51 the event classes are tiered (`BaseAuditEvent` → `MutationAuditEvent` adds `change` → `AIActionAuditEvent` adds `tool`/`model`/`purpose`), so which columns a row populates is fully determined by its `category`, its origin, and its outcome; downstream filters by category and reads that tier's columns, no if/elif guessing.

regisb (approved):
Alright! Let's go with this and we'll figure out later if we need to adjust the API. Feel free to rebase without re-review AFAIC.

abdulrafey1 (approved):

PR #496, opened 2026-07-06
feat(permissions): add whitelist container permission scope

### LINKED ISSUE #489: [Core API] Whitelist permissions can only be granted globally, not scoped to a single entry
## Problem

The email-whitelist permissions (`email.whitelist.read`, `email.whitelist.create`,
`email.whitelist.delete`) can only be granted at the `GLOBAL` scope — currently the only scope the
platform ships (`app/core/permissions/scopes.py`). As a result, the whitelist endpoints gate with
`require_in_global_scope()`, which authorizes a holder for the **entire** whitelist. There is no way
to authorize a user for a single whitelist entry.

## Where

- `app/api/v1/whitelist.py` — `GET /`, `POST /`, and `DELETE /{entry_id}` all use
  `EMAIL_WHITELIST_*.require_in_global_scope()`.
- `DELETE /{entry_id}` receives an `entry_id` path parameter, but it plays no part in
  authorization — the permission check is global.

## Why it matters

Global-only grants over-authorize: to let someone manage even one entry you must grant them
authority over the whole whitelist. A per-entry boundary would let a delegated admin manage specific
entries without full whitelist control.

abdulrafey1:
## What
Adds a `whitelist` **container** permission scope so a "Whitelist Manager" role can manage the whole registration whitelist without being a global admin — solving #489's over-authorization by modeling the whitelist as an objectless singleton scope under `global`, with existing global admins preserved via a new parent→child scope cascade.

## Changes
- feat(permissions): add `objectless` scope property and the `whitelist` scope (objectless singleton container under `GLOBAL`)
- refactor(permissions): enforce the `(scope, object_id)` pairing in `assign_role` (raising `InvalidScopeObjectId`) and drop the name-based DB CHECK — the database stays vocabulary-agnostic
- feat(permissions): cascade grants from objectless ancestor scopes (`can`/`has_role`); a `global` grant now satisfies a `whitelist` check (parent→child only; object-bearing multi-level cascade still deferred, #420 Phase 2)
- feat(permissions): make `require()`'s `scope_param` optional, with a startup guard (objectless ⇒ no param, object-bearing ⇒ param)
- feat(permissions): gate all three whitelist endpoints at `require("whitelist")`; drop `require_in_global_scope()` from those routes
- fix(permissions): let the CLI assign objectless non-global scopes (`roles assign-role --scope whitelist`)
- docs: document the whitelist scope, `objectless` scopes, and the cascade in the README

## How to Test
1. `make test.backend.pytest` — full suite passes (new: engine cascade + pairing tests in `tests/permissions/`, dependency guard tests, CLI whitelist tests, and endpoint integration tests in `tests/api/v1/test_whitelist.py`).
2. Create a role granting `email.whitelist.read/create/delete` and assign it at the whitelist scope (no global grant): `make cli -- roles assign-role <user> <role> --scope whitelist`.
3. As that user, `GET/POST/DELETE /api/v1/whitelist/*` all succeed (200/201/204) — a Whitelist Manager manages the whole whitelist without global admin.
4. An existing **global** admin still manages the whitelist (authorized via the `global → whitelist` cascade); a user with neither grant gets **403**.

## Notes
- **Stacked PR:** targets `rafey/feat/permissions-role-api` (#473), not `main`. Review/merge #473 first; this diff is the 8 commits on top of it.
- **Migration required:** new Alembic migration drops the `ck_role_assignment_scope` CHECK (Postgres). Tests build the schema from `SQLModel.metadata`, so the migration is not exercised by the suite; it was verified live against Postgres (upgrade → downgrade → upgrade). Its `downgrade()` recreates the `global`-only CHECK, so any objectless non-global (`whitelist`) rows must be cleared before downgrading.
- **Behavior change (backward compatible):** `can`/`has_role` now cascade from objectless ancestor scopes. Verified no existing behavior regresses (parent→child only; writes never cascade); the retained global-admin whitelist tests are the regression guard.
- No new env vars or dependencies.

### DIFF
--- a/app/core/permissions/__init__.py
+++ b/app/core/permissions/__init__.py
@@ -24,7 +24,7 @@ from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.core.permissions.exceptions import InvalidScopeObjectId, PermissionNotFound, RoleNotFound
 from app.core.permissions.models import Role, RoleAssignment, RolePermission
-from app.core.permissions.scopes import GLOBAL, PermissionScope, get_permission_scope
+from app.core.permissions.scopes import GLOBAL, PermissionScope
 from app.lib.auth import get_current_user
 from app.lib.db import get_async_session
 from app.lib.hooks import SingleNamedItemHook
@@ -92,28 +92,32 @@ class Permission:
         """
         return self._require_permission(GLOBAL, None)
 
-    def require(self, permission_scope: str, scope_param: str | None = None) -> Callable[..., Awaitable[User]]:
-        """Return a FastAPI dependency enforcing this permission at a named scope.
+    def require(
+        self, permission_scope: PermissionScope, scope_param: str | None = None
+    ) -> Callable[..., Awaitable[User]]:
+        """Return a FastAPI dependency enforcing this permission at a scope.
 
         Args:
-            permission_scope: A registered scope kind's name (e.g. ``"course"``, ``"whitelist"``).
-                Resolved via ``get_permission_scope``; an unregistered name raises
-                ``PermissionScopeNotFound`` here (at route-definition/import time).
+            permission_scope: The scope to check at, as a declared ``PermissionScope`` object
+                (e.g. ``GLOBAL``, ``WHITELIST``, your ``COURSE``). Pass the registered instance,
+                not its name — this keeps scope references object-based, like the RBAC engine
+                functions, so a typo is an import error rather than a silent lookup miss.
             scope_param: The route path parameter carrying the scope object id, resolved from
                 ``request.path_params`` per request. Omit for an objectless scope (it names no
                 object). Supplying it for an objectless scope, or omitting it for an
                 object-bearing one, is a wiring error and raises ``ValueError`` at definition time.
         """
-        scope = get_permission_scope(permission_scope)
-        if scope.objectless and scope_param is not None:
+        if permission_scope.objectless and scope_param is not None:
             raise ValueError(
-                f"Scope {scope.name!r} is objectless and names no object; drop the scope_param {scope_param!r}"
+                f"Scope {permission_scope.name!r} is objectless and names no object; "
+                f"drop the scope_param {scope_param!r}"
             )
-        if not scope.objectless and scope_param is None:
+        if not permission_scope.objectless and scope_param is None:
             raise ValueError(
-                f"Scope {scope.name!r} is object-bearing; require() needs a scope_param naming its path parameter"
+                f"Scope {permission_scope.name!r} is object-bearing; "
+                "require() needs a scope_param naming its path parameter"
             )
-        return self._require_permission(scope, scope_param)
+        return self._require_permission(permission_scope, scope_param)
 
 
 # Every permission the platform knows; Permission.create() registers each one here.
--- a/app/core/permissions/__init__.py
+++ b/app/core/permissions/__init__.py
@@ -126,6 +126,16 @@ EMAIL_WHITELIST_READ = Permission.create("email.whitelist.read")
 EMAIL_WHITELIST_CREATE = Permission.create("email.whitelist.create")
 EMAIL_WHITELIST_DELETE = Permission.create("email.whitelist.delete")
 
+# The role-management permissions gate the role-management API (app/api/v1/permissions).
+ROLE_CREATE = Permission.create("role.create")
+ROLE_READ = Permission.create("role.read")
+ROLE_UPDATE = Permission.create("role.update")
+ROLE_DELETE = Permission.create("role.delete")
+
+# Reading the permission vocabulary itself (the assignable-permissions listing), distinct from
+# reading roles.
+PERMISSION_READ = Permission.create("permission.read")
+
 
 def _active_assignment_at_scope(
     user_id: int | None,
--- a/app/core/permissions/__init__.py
+++ b/app/core/permissions/__init__.py
@@ -92,17 +92,27 @@ class Permission:
         """
         return self._require_permission(GLOBAL, None)
 
-    def require(self, permission_scope: str, scope_param: str) -> Callable[..., Awaitable[User]]:
+    def require(self, permission_scope: str, scope_param: str | None = None) -> Callable[..., Awaitable[User]]:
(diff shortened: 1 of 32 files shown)


abdulrafey1:
@claude-review

app/core/permissions/scopes.py
```diff
@@ -14,18 +14,26 @@ class PermissionScope:
     The hierarchy lets a role granted at a parent scope cascade to its descendants.
     """
 
-    def __init__(self, name: str, parent: "PermissionScope | None" = None) -> None:
+    def __init__(self, name: str, parent: "PermissionScope | None" = None, *, objectless: bool = False) -> None:
```
regisb: Boolean flags are a red flag for me. It's often better to replace them by a different function, or in this case, a different child class. The boolean flag will cause some different behaviour in class methods, which means that child class should override their parent. 

There's another issue with the `objectless` flag here, which is that it's leaking into the rest of the code base. We can't make this flag private because it needs to be accessed by some code outside of the PermissionScope class. Which means that we should probably move this code inside the class, such that the objectless attribute is not exposed to the outside world.

app/core/permissions/__init__.py
```diff
@@ -142,24 +156,36 @@ def _active_assignment_at_scope(
...
+    ``course`` under an ``org``) would need a materialized path to resolve its id, so it is not
+    walked yet (deferred; see issue #420 Phase 2).
+
+    Shared by the read-side checks (``can`` and ``has_role``). The write operations
+    (``assign_role``, ``revoke_role``) target the exact assignment row and must NOT use this —
+    they never cascade to ancestor scopes.
     """
+    scope_pairs: list[tuple[str, str | None]] = [(permission_scope.name, scope_object_id)]
+    for ancestor in permission_scope.get_parents():
+        if ancestor.objectless:
```
regisb: Wait, I need to understand something. Consider the following example:

1. user does not have admin permission.
2. user has permission to write to course ID=X.
3. user does not have explicit permission to write to learning activity Y, which belongs to course X.

In this example, should user have write access to learning activity Y? (I think yes) What does your implementation do?

abdulrafey1: **Short answer:** yes — the user *should* get write access to activity Y, but the current
implementation **denies it (403)**. Course→activity cascade isn't built yet; today only cascade
from *objectless* ancestor scopes (`global`) works.

### What the code does

The write endpoint for Y would run `can(user, ACTIVITY_WRITE, ACTIVITY, "Y")`. Internally,
`_active_assignment_at_scope` builds the set of `(scope, object_id)` pairs a grant may satisfy the
check from:

```python
scope_pairs = [("activity", "Y")]          # the target scope itself
for ancestor in ACTIVITY.get_parents():    # [COURSE, GLOBAL]
    if ancestor.objectless:                # ← only objectless ancestors are walked
        scope_pairs.append((ancestor.name, None))
```

- `COURSE` is **object-bearing** (many courses exist) → `objectless` is `False` → **skipped**.
- `GLOBAL` is objectless → added as `("global", None)`.

So the chain is `[("activity", "Y"), ("global", None)]`. The user's grant sits at `("course", "X")`,
which matches **neither** pair → `can()` returns `False` → **403**.

### Why it's skipped (and why that's intentional for now)

The engine only knows the scope-**kind** hierarchy (`activity → course → global` via
`get_parents()`), not the **instance** relationship *"activity Y belongs to course X."* To fold
`("course", "X")` into the chain it would have to resolve *which* course Y lives under — and there's
no resource model / `scope_chain` / materialized path to answer that yet. For objectless ancestors
the object id is trivially `NULL`, so they cascade for free; object-bearing ones can't, so they're
deferred. This is called out in the comment on `_active_assignment_at_scope` and tracked as **Phase 2
(scope hierarchy + materialized path)** in #420.

| Question | Answer |
|---|---|
| Should Y be writable? | **Yes** — an ancestor (course) grant should cascade to the descendant activity. |
| Does it today? | **No** — object-bearing ancestors are skipped; only `global` cascades. |
| Why | No `scope_chain` / materialized path to know that Y ∈ X. |
| Bug or deferred? | **Deferred** (Phase 2, #420) — this PR scoped cascade to objectless ancestors only. |

app/core/permissions/scopes.py
abdulrafey1: - **No more boolean flag** — `PermissionScope` is now an abstract base; `ObjectlessScope` (`global`, `whitelist`) and `ObjectScope` (`course`, …) subclasses encapsulate the divergent behavior.
- **No more leaking `objectless`** — the three former `if scope.objectless:` call sites now call methods (`validate_object_id`, `validate_scope_param`, `scope_chain`); the attribute is gone entirely.
- **Bonus:** the deferred `course → activity` cascade now has a clear home (`ObjectScope._cascade_pairs`), with a test documenting that it doesn't cascade yet.

app/core/permissions/scopes.py
```diff
@@ -48,11 +54,85 @@ def get_parents(self) -> list["PermissionScope"]:
...
+        """
 
+    @abstractmethod
+    def _cascade_pairs(self) -> list[tuple[str, str | None]]:
+        """This scope's ``(name, object id)`` contribution to a descendant's :meth:`scope_chain`
+        when it appears as an ancestor."""
 
-# Every scope kind the platform knows; PermissionScope.create() registers each one here.
+
+class ObjectlessScope(PermissionScope):
```
regisb: I'm wondering if this is the right name for this class. Shouldn't we merge it in PermissionScope, instead of having an abstract class? Then ObjectScope would be renamed to ObjectPermissionScope, or something similar?

regisb (approved):
Just a naming issue, but otherwise good to go! Feel free to ignore my comment if you disagree.

app/core/permissions/scopes.py
abdulrafey1: As the codebase grows `ObjectScope` would get more common, so, if anything, `PermissionScope` should represent the common occurrence , and the other class should represent the rare one.

We can have the scope with objects called `PermissionScope` and the ones without object references can be called `ObjectlessPermissionScope`

WDYT?

regisb: Agreed!

PR #499, opened 2026-07-06
feat(permissions): add per-role permission scope

### LINKED ISSUE #490: [Core API] Role management can only be authorized globally, not scoped to a single role
## Problem

Role assignment and management are authorized only at the `GLOBAL` scope — currently the only scope
the platform ships (`app/core/permissions/scopes.py`). There is no way to grant a user authority over
a **specific** role (e.g. permission to assign/revoke only the `grader` role); a grant is all-or-nothing
across every role.

## Where

- `app/core/permissions/utils.py` — `assign_role` / `revoke_role` operate with no per-role
  authorization boundary.
- `app/cli/roles.py` — role assignment via the CLI; any future role-management endpoint would have
  to gate globally.

## Why it matters

Delegating management of one role shouldn't require granting authority over all roles. A per-role
boundary is needed before role administration can be safely delegated.

abdulrafey1:
## What
Adds an object-bearing `role` permission scope so role management can be delegated to a single role — the role-management endpoints addressing one role by id now authorize at `require("role", "role_id")` instead of globally, solving #490's all-or-nothing authority. Existing global admins are preserved for free via the `global → role` scope cascade built in #496.

## Changes
- feat(permissions): add the object-bearing `role` scope (child of the objectless `global` root; re-exported from `app.lib.permissions.scopes`)
- feat(permissions): gate the five `{role_id}` role-management endpoints (get/update/delete + the role's permission grants) at the `role` scope; listing all roles, creating a role, and reading the permission vocabulary stay global
- docs: document the `role` scope, per-role delegation, the endpoint scope table, the CLI delegation command, and the accepted escalation limitation
- test(permissions): scope registration, per-role endpoint enforcement (incl. cross-role 403, global-admin cascade, per-role grant not authorizing the global endpoints, 403-before-404), and an end-to-end CLI object-bearing assignment

## How to Test
1. `make test.backend.pytest` — full suite passes (new: `role` scope registration in `tests/permissions/test_scopes.py`, per-role endpoint tests in `tests/api/v1/test_permissions.py`, and a CLI success test in `tests/cli/test_roles.py`). `make mypy`, `make lint.backend`, and `make lint.format.backend check=1` are all clean.
2. Create a role to be managed and note its id (e.g. `5`). Create a "role-manager" role granting `role.update`, and delegate it to just that role: `make cli -- roles assign-role alice role-manager --scope role --scope-object-id 5`.
3. As `alice`, `PATCH /api/v1/permissions/roles/5` and `POST /api/v1/permissions/roles/5/permissions` succeed (200); the same calls against a **different** role id return **403**. `GET /api/v1/permissions/roles` (list all), `POST /api/v1/permissions/roles` (create), and `GET /api/v1/permissions` return **403** for her (they stay global).
4. An existing **global** admin still manages every role (authorized via the `global → role` cascade); a user with neither grant gets **403**.

## Notes
- **Stacked PR:** targets `rafey/add-whitelist-permission-scope` (#496), not `main` — this PR builds on #496's `objectless` scope property, the parent→child cascade, and `require()`'s optional `scope_param`. Review/merge #496 first; this diff is the 3 commits on top of it.
- **No migration, no CLI code, no frontend change.** The scope vocabulary lives in code (#496 already dropped the name-based CHECK), the cascade authorizes existing global admins for free, the CLI already accepts `--scope-object-id`, and swapping FastAPI dependencies leaves the OpenAPI schema (and `frontend/lib/api/generated.ts`) unchanged. No new env vars or dependencies.
- **Behavior change (backward compatible):** the five per-role endpoints move off `require_in_global_scope()`. Global admins are unaffected (verified by the retained global-admin tests); the change only *adds* a narrower delegation path.
- **Accepted limitation:** a per-role manager holding `role.update` can grant its role **any** registered permission — the same capability a global `role.update` holder already has. Per-role delegation narrows *who* holds it, not *what* it can grant; restricting the grantable set is out of scope.
- **Dangling assignments on delete:** `DELETE /roles/{role_id}` is now a reachable per-role endpoint, and `role` is the first object-bearing scope reachable via the API. Deleting a role can leave `role_assignment` rows whose `scope_object_id` pointed at it (that column is deliberately polymorphic, not a foreign key). This is inherent to the scoped-RBAC design and applies to any future object-bearing scope; cleanup is out of scope here.


abdulrafey1:
@claude-review

app/api/v1/permissions/routes/roles.py
```diff
@@ -73,7 +78,7 @@ async def create_role(payload: RoleCreate, session: AsyncSession = Depends(get_a
 @router.get(
     "/{role_id}",
     response_model=RoleResponse,
-    dependencies=[Depends(ROLE_READ.require_in_global_scope())],
+    dependencies=[Depends(ROLE_READ.require(ROLE, "role_id"))],
```
regisb, line 81: What's interesting with this change is that we can see that a role is always going to be tied to the same `ROLE` scope type. For instance, we are never going to write `ROLE_READ.require(COURSE, "course_id")`. Does this mean that ROLE_READ should be tied in some way to the `ROLE` scope at construction time?

regisb (commented):
just a question, otherwise looks good.

app/api/v1/permissions/routes/roles.py
abdulrafey1: Most permissions do have a "natural home scope" (`email.whitelist.*` → `whitelist`, `role.*` → `role`). But I'd argue against binding the scope onto the permission at construction, for a few reasons.

**1. It isn't actually 1:1 — even in this file.** `ROLE_READ` is already gated at *two* different scopes:

- `GET /roles` (list all) → `ROLE_READ.require_in_global_scope()` — `role.read` at **`global`**
- `GET /roles/{role_id}` (one role) → `ROLE_READ.require(ROLE, "role_id")` — `role.read` at **`role`**

And these aren't the same check reached via cascade. Cascade only flows **parent → child**, so a holder of `role.read` at `(role, 5)` can read role 5 but is *not* authorized to list all roles, while a `global` `role.read` holder can do both. Same permission, genuinely different boundaries — that's exactly the delegation split this PR is built on. If `ROLE_READ` were pinned to `ROLE` at construction, the list endpoint could no longer express "read across *all* roles."

So `ROLE_READ.require(COURSE, "course_id")` is indeed nonsense — but `require_in_global_scope()` vs `require(ROLE, "role_id")` is not, and both are needed.

**2. "What" and "where" are orthogonal by design.** A `Permission` answers *what* action; a `PermissionScope` answers *where* it applies. `can(user, permission, scope, object_id)` takes them as independent arguments, and the two vocabularies live in independent registries (`PERMISSIONS`, `PERMISSION_SCOPES`). Pinning a scope onto a permission collapses that separation — and it's precisely that separation that lets one permission be enforced at different granularities on different routes and cascade across the hierarchy.

**3. The route is the right place to state the boundary.** `require(ROLE, "role_id")` on the decorator puts the endpoint's full authorization contract on one line: *this permission, at this scope, keyed by this path param.* Moving the scope onto the permission hides half of that from the route, and it introduces a declaration-order coupling — the scope object would have to be constructed before the permission, whereas today the two vocabularies are declared independently.

**4. If the worry is mis-gating** (someone writing `ROLE_READ.require(COURSE, ...)`), a lighter guard fits better than construction-time binding — e.g. a test asserting the role router only ever gates on `ROLE`/`GLOBAL`. But it's one extra token per route and a wrong scope surfaces immediately in tests, so that's probably YAGNI.

**tl;dr** — the pairing is real and worth noting as a convention, but I'd keep the scope at the gate rather than on the permission: the one-token repetition buys orthogonality, multi-scope reuse (which `role.read` already relies on today), and routes that document their own authorization.

regisb: OK I understand and I agree!

regisb (approved):
lgtm!

PR #509, opened 2026-07-07
refactor: restructure package into core / lib / plugins tiers

### LINKED ISSUE #497: Package layout has no structural core boundary and ambiguous naming
## Summary

The backend has no structural boundary that marks what is **"core"**, and the
top-level package naming is ambiguous. Plugins are meant to depend only on the
`app.lib` façade, but nothing in the layout enforces that — the boundary exists
by convention alone, and in practice plugins reach past the façade directly into
internal packages. Combined with generic/inverted directory names, this makes the
codebase hard to navigate and blocks the "plugins as standalone units" goal.

## Where it happens

**1. Generic top-level package name**

- `app/` — the backend package is named `app`. Nothing about the name identifies
  it as Sparkth, and it collides conceptually with the FastAPI `app` object.

**2. Inverted plugin naming**

- `app/plugins/` — the plugin **framework** (`base.py`, `loader.py`,
  `middleware.py`, `config_base.py`).
- `app/core_plugins/` — the **built-in plugins** (`canvas/`, `chat/`,
  `googledrive/`, `openedx/`, `slack/`).

  The names are inverted: the framework is the part that is *core*, while the
  built-ins are "just plugins". A newcomer reasonably expects `plugins/` to hold
  the plugins.

**3. Plugins bypass the `app.lib` façade**

`app/lib/` is documented as the curated public API that plugins import from, yet
plugins import internals directly (`grep -rE "from app\.(models|main|core|services)" app/core_plugins/`):

- `from app.models.user import User` ×16, plus `app.models.llm`, `app.models.base`
- `from app.main import app` ×5, `from app.main import assemble_app` ×1
- `from app.core.config import get_settings` ×2
- `from app.core.encryption import get_encryption_service` ×2
- `from app.services.plugin import PluginService` ×2

Because there is no structural core, every internal symbol a plugin imports
becomes implicit public API and blocks refactoring.

**4. Shared test harness sits outside the façade**

- `app/testing.py` — the shared pytest fixtures/environment. Plugin test suites
  import it as `app.testing` (`app/core_plugins/slack/tests/conftest.py`,
  `app/core_plugins/googledrive/tests/conftest.py`), i.e. through a top-level
  module rather than the façade plugins are supposed to consume.

## Why it matters

- **No enforceable core boundary.** With core scattered across `models/`,
  `services/`, `main.py`, and `core/`, there is no single place that answers
  "what may a plugin depend on?" Every accidental import becomes frozen API
  (directly related to #379 — *"Plugins are missing a well-defined core API"*).
- **Ambiguous naming raises onboarding cost** and steers contributors toward the
  wrong imports.
- **Plugins are expected to become standalone units** (eventually separate repos —
  see the test-layout rationale in `CLAUDE.md`). Without a crisp
  core / façade / plugins split, that separation is painful.

## How to observe

Run at the repo root:

```bash
# Façade bypass — plugins importing internals directly
grep -rE "from app\.(models|main|core|services)" app/core_plugins/ --include="*.py"

# Test harness imported outside the façade
grep -rl "app.testing" app/core_plugins/
```

abdulrafey1:
## What
Restructures the backend package into three clear tiers — `sparkth/core/` (shared internals plugins depend on), `sparkth/lib/` (the façade, the only surface plugins import from), and `sparkth/plugins/` (built-in plugins) — replacing the old `app/` name and the ambiguous `plugins/` (framework) vs `core_plugins/` (built-ins) split, so "core" is a real boundary and plugins stop reaching into internals (ties into #379).

## Changes
- refactor: rename the top-level `app/` package to `sparkth/` (imports, string module-paths, `pyproject`, `lefthook`, `Makefile`, `Dockerfile`, both `alembic*.ini`, playwright CI, `conftest` `pytest_plugins`, docs)
- refactor(plugins): move built-in plugins `core_plugins/` → `sparkth/plugins/` and the plugin framework `plugins/` → `sparkth/core/plugins/`
- refactor(core): move the whole `models/` package → `sparkth/core/models/` and `PluginService` → `sparkth/core/plugins/service.py`; add `sparkth/lib/models.py` and `sparkth/lib/encryption.py` façades and re-export the plugin-service surface from `sparkth/lib/plugins.py`
- refactor(testing): move the test harness `testing.py` → `sparkth/lib/testing.py`
- fix(log): `get_logger` no longer double-prefixes (`sparkth.sparkth.*` → `sparkth.*`) after the rename
- fix(e2e): point the Playwright e2e backend wrapper at `sparkth.main:app`
- docs: rewrite the CLAUDE.md directory tree to the three-tier layout and fix stale path references

## How to Test
1. `make test.backend` — full backend gate is green: `ruff check`, `ruff format --check`, `mypy --strict sparkth/ tests/ scripts/` (343 files), and `pytest` (**1354 passed**).
2. `make test.frontend` — green: oxlint, react-doctor, **OpenAPI drift check clean** (`generated.ts` unchanged — the restructure doesn't alter the API surface), tsc typecheck, vitest (158 passed), oxfmt.
3. App boots and plugins load via the new module strings:
   `uv run python -c "from sparkth.main import app; from sparkth.lib.plugins import get_plugin_loader; print(len(app.routes), sorted(n for n,_ in get_plugin_loader().get_loaded_plugins()))"`
   → `72 ['canvas', 'chat', 'google-drive', 'open-edx', 'slack']`.
4. CLI entrypoint works: `make cli -- users --help` (`python -m sparkth.cli.main`). Both Alembic lineages resolve: `uv run alembic heads` and `uv run alembic -c alembic_analytics.ini heads`.
5. Plugins import only through the façade: `grep -rnE "sparkth\.(core|models|services)\b" sparkth/plugins/` returns nothing (they use `sparkth.lib.*`).

## Notes
- **Now targets `main`.** Originally opened stacked on #499; #496 and #499 have since merged, so this branch was rebased onto `main` (which also picked up #512's CI-on-all-target-branches change). Rebasing preserved main's evolved permissions code — the diverged files were reconstructed as *main's content with the restructure applied*, verified file-by-file.
- **No migration, no new env vars, no new dependencies.** Pure structural refactor: no schema change, and the two Alembic `env.py` files only had import paths updated.
- **No API-surface change.** Routes, operation IDs, and Pydantic schema names are unchanged (module renames don't affect them), so `frontend/lib/api/generated.ts` is unmodified (verified by the frontend OpenAPI drift check).
- **Plugin identity unchanged.** Runtime plugin names derive from the class name, not the module path, so no `plugin` DB rows or frontend plugin keys change — only `PLUGINS` module-strings in `sparkth/core/config.py`.
- **Behavior-preserving**, with one intended, backward-compatible change: logger names go from `sparkth.app.*` to `sparkth.*` (previously would have been `sparkth.sparkth.*`; see the `fix(log)` commit). `PluginService` is re-exported from `sparkth.lib.plugins` **lazily** to avoid an import cycle, mirroring the existing `PluginAccessMiddleware` pattern.


hamza-56 (approved):

regisb (changes requested):
How come CI did not run on this PR? We shouldn't merge until we have a greenlight from mypy, unit tests, etc.

EDIT: that's because this PR is not running on top of "main". Should we run CI for all PRs instead of just those that target the "main" branch?

abdulrafey1:
> How come CI did not run on this PR? We shouldn't merge until we have a greenlight from mypy, unit tests, etc.
> 
> EDIT: that's because this PR is not running on top of "main". Should we run CI for all PRs instead of just those that target the "main" branch?

I think yes, since stacked PRs are easier to review while working on multiple tickets/multi-stage tickets.

regisb (approved):
I am prepared to blindly accept whatever is in this PR

PR #511, opened 2026-07-08
feat(analytics): add permission-gated login-activity read api

zaira-bibi:
## What
Add a permission-gated read API for daily login activity — the first read-side slice of the analytics subsystem, built on the `user.logged_in` event.

## Changes
- feat(migrations): add `login_activity_daily` TimescaleDB continuous aggregate (Postgres-only; no-op on SQLite)
- feat(core): declare and export the `analytics.read` permission
- feat(analytics): add dialect-aware `get_login_activity` read function, exposed via `app.lib.analytics`
- feat(api): add `GET /api/v1/analytics/login-activity`, gated by `analytics.read`
- feat(migrations): grant `analytics.read` to the seeded `admin` role
- docs: document `analytics.read` (README) and the read API (CLAUDE.md)

## How to Test
1. `make test.backend.pytest` — all green (1316 passed), including `tests/analytics/test_login_activity_read.py` and the migration no-op guard in `tests/analytics/test_analytics_migrations.py`.
2. `make mypy` — clean; `make lint.backend` and `make lint.format.backend check=1` — clean.
3. Single migration heads: `uv run alembic heads` and `uv run alembic -c alembic_analytics.ini heads` each show one head.
4. On real Postgres/TimescaleDB (`make services.up`, `make migrations`): 
- `make db-shell-analytics` → `\d+ login_activity_daily` and `SELECT * FROM timescaledb_information.continuous_aggregates;` show the aggregate + refresh policy.
   - `make db-shell` → confirm the `admin` role has a `analytics.read` grant.
   - Log in as an admin and call `GET /api/v1/analytics/login-activity?days=30`; a non-admin gets `403`.

## Notes
- **Migrations required (two lineages):** app grant `c5b6e911f9d6`; analytics continuous aggregate `287f281e6558`. Run `make migrations`.
- The continuous aggregate is TimescaleDB-only — a no-op on SQLite (tests/e2e), where the read path aggregates `raw_events` directly. The Postgres cagg and the admin grant can only be verified on real Postgres (step 4).
- No breaking changes, no new env vars, no dependency bumps.
- Frontend dashboard (the `/dashboard/analytics` page + typed client) is intentionally out of scope here and will follow in a separate PR.

### DIFF
--- a/sparkth/core/analytics/reads.py
+++ b/sparkth/core/analytics/reads.py
@@ -1,0 +1,78 @@
+"""Read-side queries over analytics data.
+
+The read path never touches the application database — only analytics tables and
+rollups. Exposed to the API layer via ``sparkth.lib.analytics``.
+
+``get_login_activity`` is dialect-aware because the login-activity rollup only
+exists as a TimescaleDB continuous aggregate on PostgreSQL. On SQLite (the test /
+e2e database) the aggregate is not created (see the analytics migration), so we
+compute the same daily buckets directly from ``raw_events``. The two query variants
+below must stay semantically identical — same filter (``user.logged_in``), same
+day bucketing, same columns, and same date-floor window (last ``days`` calendar
+days).
+
+Day boundaries are UTC on both paths, and this is the premise that makes them
+identical: PG's ``time_bucket('1 day', occurred_at)`` buckets the timestamptz in
+UTC, while SQLite's ``date(occurred_at)`` buckets on the stored string's offset —
+they agree only because the ingest path stores/interprets events as UTC. If the PG
+side is ever parameterized with a bucketing timezone, the SQLite side must move in
+lockstep or the two will silently diverge.
+"""
+
+from pydantic import BaseModel
+from sqlalchemy import text
+from sqlmodel.ext.asyncio.session import AsyncSession
+
+
+class LoginActivityPoint(BaseModel):
+    """One day's login count. ``day`` is an ISO ``YYYY-MM-DD`` string."""
+
+    day: str
+    login_count: int
+
+
+# Both queries are raw SQL rather than SQLAlchemy Core selects, for two reasons:
+#   1. The PostgreSQL query reads ``login_activity_daily`` — the TimescaleDB continuous
+#      aggregate. That relation is created by raw migration DDL and is deliberately NOT
+#      registered as a ``sa.Table`` on ``analytics_metadata`` (doing so would make Alembic
+#      autogenerate try to create/drop it, and it doesn't exist on SQLite). With no table
+#      object to select against, a Core query would need a throwaway Table declaration plus
+#      Postgres-only funcs (``to_char``) — i.e. the same SQL with more ceremony.
+#   2. The two dialects need genuinely different queries (different source relation, different
+#      day-bucketing), so Core could not unify them anyway. Writing them as two side-by-side
+#      SQL strings makes the "must stay semantically identical" invariant (see module docstring)
+#      easy to eyeball. ``raw_events`` IS a ``sa.Table``, so the SQLite branch could be Core, but
+#      mixing one Core branch with one raw branch would obscure that pairing.
+
+# PostgreSQL: read the pre-rolled continuous aggregate.
+_PG_SQL = text(
+    "SELECT to_char(day AT TIME ZONE 'UTC', 'YYYY-MM-DD') AS day, login_count "
+    "FROM login_activity_daily "
+    "WHERE (day AT TIME ZONE 'UTC')::date >= (now() AT TIME ZONE 'UTC')::date - :days "
+    "ORDER BY day DESC "
+    "LIMIT :days"
+)
+
+# SQLite (tests/e2e): aggregate raw_events directly — same semantics as the cagg.
+_SQLITE_SQL = text(
+    "SELECT date(occurred_at) AS day, count(*) AS login_count "
+    "FROM raw_events "
+    "WHERE event_type = 'user.logged_in' "
+    "AND date(occurred_at) >= date('now', '-' || :days || ' days') "
+    "GROUP BY date(occurred_at) "
+    "ORDER BY day DESC "
+    "LIMIT :days"
+)
+
+
+async def get_login_activity(session: AsyncSession, days: int = 30) -> list[LoginActivityPoint]:
+    """Return daily login counts, newest first, for the last ``days`` calendar days.
+
+    The window is bounded by a date floor (``now - days``), so gap days never let
+    the series reach back beyond the requested window. Days with no logins are
+    omitted entirely — the series is not zero-filled; callers must tolerate gaps.
+    """
+    dialect = session.bind.dialect.name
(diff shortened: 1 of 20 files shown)


zaira-bibi:
@claude-review

sparkth/core/analytics/reads.py
```diff
@@ -0,0 +1,47 @@
...
+
+class LoginActivityPoint(BaseModel):
+    """One day's login count. ``day`` is an ISO ``YYYY-MM-DD`` string."""
+
+    day: str
+    login_count: int
+
+
+# PostgreSQL: read the pre-rolled continuous aggregate.
+_PG_SQL = text(
```
abdulrafey1, line 48: Lets avoid adding raw SQL in code

sparkth/migrations/analytics/versions/287f281e6558_create_login_activity_daily_cagg.py
```diff
@@ -0,0 +1,57 @@
...
+    # Continuous aggregates are a TimescaleDB (Postgres) feature. On SQLite (tests,
+    # e2e) this is a no-op — the S2 read path aggregates raw_events directly there.
+    if bind.dialect.name != "postgresql":
+        return
+    # WITH NO DATA so creation does not backfill inside Alembic's transaction; the
+    # continuous-aggregate policy below keeps it fresh going forward, and a one-off
+    # backfill (refresh_continuous_aggregate) is run out-of-band if history matters.
+    op.execute(
+        """
+        CREATE MATERIALIZED VIEW login_activity_daily
```
abdulrafey1, line 40: Could we check if we can do this via the ORM

sparkth/migrations/app/versions/c5b6e911f9d6_grant_analytics_read_to_admin_role.py
```diff
@@ -0,0 +1,48 @@
...
+        ),
+        {"permission": _PERMISSION, "role": _ROLE},
+    )
+
+
+def downgrade() -> None:
+    conn = op.get_bind()
+    conn.execute(
+        sa.text(
+            "DELETE FROM role_permission WHERE permission = :permission "
```
abdulrafey1, line 44: Same here

sparkth/migrations/analytics/versions/287f281e6558_create_login_activity_daily_cagg.py
```diff
@@ -0,0 +1,63 @@
...
+        WHERE event_type = 'user.logged_in'
+        GROUP BY day
+        WITH NO DATA
+        """
+    )
+    op.execute(
+        """
+        SELECT add_continuous_aggregate_policy(
+            'login_activity_daily',
+            start_offset      => INTERVAL '3 days',
```
hamza-56, line 54: This `start_offset` combined with `WITH NO DATA` means anything older than 3 days at migration time never gets materialized: once the first policy run advances the watermark, real-time aggregation stops covering those old buckets and they disappear from the view entirely. Since `user.logged_in` has been emitting for a while, a deployed instance silently loses its pre-migration history here (the SQLite path keeps full history, so the two dialects drift). The comment above mentions an out-of-band backfill, but nothing in the repo runs one. Creating the view `WITH DATA`, or a one-off `refresh_continuous_aggregate(login_activity_daily, NULL, NULL)`, would fix it.

sparkth/core/analytics/reads.py
```diff
@@ -0,0 +1,60 @@
...
+#      Postgres-only funcs (``to_char``) — i.e. the same SQL with more ceremony.
+#   2. The two dialects need genuinely different queries (different source relation, different
+#      day-bucketing), so Core could not unify them anyway. Writing them as two side-by-side
+#      SQL strings makes the "must stay semantically identical" invariant (see module docstring)
+#      easy to eyeball. ``raw_events`` IS a ``sa.Table``, so the SQLite branch could be Core, but
+#      mixing one Core branch with one raw branch would obscure that pairing.
+
+# PostgreSQL: read the pre-rolled continuous aggregate.
+_PG_SQL = text(
+    "SELECT to_char(day, 'YYYY-MM-DD') AS day, login_count FROM login_activity_daily ORDER BY day DESC LIMIT :days"
```
hamza-56: `time_bucket` buckets on UTC midnight, but `to_char` renders a `timestamptz` in the session `TimeZone`, and we don't pin one on the analytics engine. On any server west of UTC every label shifts one day back, so today's logins show up under yesterday. The SQLite path is effectively UTC, so tests will never catch this. `to_char(day AT TIME ZONE 'UTC', 'YYYY-MM-DD')` makes it deterministic.

sparkth/core/analytics/reads.py
```diff
@@ -0,0 +1,60 @@
...
+)
+
+# SQLite (tests/e2e): aggregate raw_events directly — same semantics as the cagg.
+_SQLITE_SQL = text(
+    "SELECT date(occurred_at) AS day, count(*) AS login_count "
+    "FROM raw_events "
+    "WHERE event_type = 'user.logged_in' "
+    "GROUP BY date(occurred_at) "
+    "ORDER BY day DESC "
+    "LIMIT :days"
```
hamza-56, line 64: `LIMIT :days` returns the N most recent days that had logins, not "the last N days" as the endpoint doc says, so a quiet stretch pushes the window arbitrarily far back. It also leaves this SQLite branch scanning all of `raw_events` on every call. A `WHERE occurred_at >= ...` bound on both queries would match the documented contract and cap the scan.

sparkth/core/analytics/reads.py
zaira-bibi: went with raw SQL here on purpose as the postgres query reads the `login_activity_daily` continuous aggregate, which isn't (and shouldn't be) registered on `analytics_metadata`, reason being that the autogenerate will try to create/drop it and it doesn't exist on sqlite either. also worth noting that the two can't share a query (because of different source relation and day bucketing), so it's two selects either way. 

the sqlite query could be written in sqlAlchemy core, but then the readers couldn't diff them by eye.

sparkth/migrations/analytics/versions/287f281e6558_create_login_activity_daily_cagg.py
zaira-bibi: unfortunately sqlAlchemy does not have DDL construct for materialized views at all.

sparkth/migrations/analytics/versions/287f281e6558_create_login_activity_daily_cagg.py
zaira-bibi: we'd need a `refresh_continuous_aggregate(xyz_activity, NULL, NULL)` for every cagg that's created, which is the wrong shape. i can either do (a) `start_offset => NULL` in the policy itself to refresh from the beginning of time or (b) one generic backfill command to discover all caggs and full-refresh each one. i lean towards (b) as (a) will cause each refresh to re-scan full history, which will grow expensive as data accumulates. what do you say?

hamza-56 (approved):

PR #513, opened 2026-07-08
feat(api): add centralized exception-to-http mapping mechanism

### LINKED ISSUE #493: Routes need a centralized, reusable way to map domain exceptions to HTTP responses
## Problem

Across the REST API, every route that can surface a domain error repeats the same
pattern: wrap the call in `try/except`, catch each domain exception, and re-raise it
as an `HTTPException` with a hand-picked status code and detail. The mapping from a
domain exception to its HTTP response is duplicated at every call site — there is no
shared place that says "exception X renders as HTTP status Y".

Where it shows up today:

- `app/api/v1/whitelist.py` — each handler catches `ValueError` / `IntegrityError`
  inline and maps them to `409` / `422` / `404`.
- The in-flight role-management routes (#473, `app/api/v1/permissions/routes/roles.py`)
  carry ~10 near-identical `try/except` blocks doing the same thing across endpoints
  (`RoleNotFound → 404`, `RoleAlreadyExists → 409`, `RoleInUse → 409`,
  `PermissionNotFound → 422`).

## Why it matters

- **Repetition and drift.** The same exception is translated in many places. A route
  that forgets a `try/except` leaks a `500` for what should be a clean `4xx`, and two
  routes can map the same exception to different statuses by accident.
- **No extension point.** Core and plugins have no common way to declare how their
  own exceptions should be rendered. A plugin that raises a domain exception must
  hand-map it in every route that can trigger it.
- **Coupling.** HTTP status decisions are scattered through business-logic routes
  instead of living in one API-layer mapping.

## What this issue asks for

A robust, centralized mechanism for exceptions raised in routes: one place where
**core and plugins can register how each exception type is rendered as an HTTP
response**, wired onto the app once at startup — so a route can raise a domain
exception and let the framework translate it, instead of repeating
`try/except → HTTPException` everywhere. Core exceptions must stay HTTP-agnostic;
the mapping lives in the API layer.

## Scope

- **In scope:** the mechanism itself — the registration surface, its wiring into the
  app at startup, and tests — so any route or plugin can use it.
- **Out of scope:** migrating existing endpoints. The per-module migrations
  (permissions routes, whitelist, and any others) will each be tracked in their own
  separate issues, not here. This issue delivers only the capability.

abdulrafey1:
## What
Add a centralized, reusable mechanism where core and plugins register how each domain exception type renders as an HTTP response, wired onto the app once at startup — so routes can raise HTTP-agnostic domain exceptions instead of repeating `try/except → HTTPException` at every call site.

## Changes
- feat(api): add the exception→HTTP registry in `sparkth/core/exceptions/handlers.py` (`EXCEPTION_HANDLERS` + `register_exception_handler(exc_type, status_code)`, rendering `{"detail": str(exc)}`) with a public facade at `sparkth/lib/exceptions/handlers.py`
- refactor(core): the registry reuses the existing `KeyedItemHook` primitive rather than adding a new type-keyed hook (per review); it stores `(exc_type, handler)` entries keyed on `exc_type` (duplicate → `ValueError`), and adds `KeyedItemHook.iter_values()` so `assemble_app` can iterate them
- feat(api): wire the registry into `assemble_app()` via `_register_exception_handlers` in `sparkth/main.py`, inheriting Starlette's MRO-correct dispatch through `add_exception_handler`
- refactor: `git mv` `sparkth/lib/exceptions.py` → `sparkth/lib/exceptions/__init__.py`, preserving the existing `HttpError`/`AuthenticationError`/`LMSRequestError` surface, and flag that legacy `status_code`-carrying family as pre-standard so it isn't copied as the current pattern
- test(api): cover registration, duplicate rejection, MRO subclass dispatch, the rendered body, the ships-empty invariant, and that `assemble_app()` performs the wiring
- docs: document the public module in `CLAUDE.md` and the plugin authoring surface in `sparkth/core/plugins/PLUGIN_GUIDE.md`

## How to Test
1. `make mypy` — passes (`mypy --strict`).
2. `make lint.backend && make lint.format.backend check=1` — both clean.
3. `make test.backend.pytest` — full suite passes. New coverage lives in `tests/core/exceptions/test_handlers.py` and `tests/lib/test_hooks.py`.
4. Behaviour check: register a mapping (e.g. `register_exception_handler(SomeError, 409)`), raise `SomeError` from a route, and confirm the response is `409` with body `{"detail": "<message>"}`; a handler registered on a base exception also catches its subclasses.

## Notes
No migration, no breaking change, no new env var, no dependency bump. **The registry ships empty** — this PR delivers only the capability. Migrating existing endpoints (e.g. `whitelist`, role-management routes) to the mechanism is intentionally out of scope and tracked in separate per-module issues. Fully unifying the remaining keyed-registry hook primitives (`KeyedClassHook` + `KeyedItemHook`) is tracked in #522.


abdulrafey1:
@claude-review

regisb:
_reviewed with experimental sparkth-code-reviewer_

## Review: exception-to-http mapping mechanism

**0 must-fix, 3 suggestions.** No checklist (M1–M21) violations found. Core mechanism, tests, and docs are consistent with each other and with existing conventions (`sparkth/lib` façade, hook pattern, TDD coverage, LLM-assistance notices, `git mv` for the `exceptions.py` → `exceptions/__init__.py` rename).

### Suggestions

1. **`sparkth/lib/hooks.py:117-129`** (new `TypeKeyedHook`) — duplicates a shape already generalized once in this file. `SingleNamedItemHook` (dict + "duplicate key raises `ValueError`" + `iter_items`) was itself generalized into `KeyedItemHook`. `TypeKeyedHook` re-implements that pattern a third time rather than extending `KeyedItemHook` to accept an explicit (non-derived) key. *Guideline: "Extract duplicated logic into a shared helper" (abstraction.md).* Judgment call — `KeyedItemHook`'s key is derived from the stored item via a callback, and the handler closure here doesn't carry its own `exc_type`, so reuse isn't a drop-in. Worth a comment either justifying the new class or attempting the `KeyedItemHook` route.

2. **`sparkth/core/exceptions/handlers.py:26-32`** — `register_exception_handler`'s docstring says `assemble_app wires the registry onto the app at startup`, describing behavior owned by a different module (`sparkth/main.py`). *Guideline: "Scope a docstring to only the class/module it documents" (docs.md).* If `_register_exception_handlers` in `main.py` is refactored later, this comment can go stale. Consider trimming the `assemble_app` implementation detail.

3. **`sparkth/lib/exceptions/__init__.py`** (renamed, unchanged content) now sits in the same package as the new `handlers.py`, but encodes the opposite design principle the PR documents: `HttpError`/`AuthenticationError`/`LMSRequestError` carry `status_code` on the exception, whereas the new CLAUDE.md section states "a domain exception never carries an HTTP status." Two competing exception→HTTP philosophies now live side-by-side with no note reconciling them. Migrating existing call sites is reasonably out of scope, but a one-line comment noting the legacy pattern predates this standard would stop a future contributor copying `HttpError`-with-`status_code` as the "current" pattern. *Guideline: "Keep a single source of truth" (abstraction.md).*

### Outside the corpus — PR description drift

The "Changes" section describes `sparkth/core/exception_handlers.py` with `EXCEPTION_HANDLERS`, `register_exception_handler`, `status_handler`, `register_status`, and a facade at `sparkth/lib/exception_handlers.py`. The shipped diff (after `d5b095b` and `a564011`) actually uses `sparkth/core/exceptions/handlers.py` / `sparkth/lib/exceptions/handlers.py` and only `register_exception_handler` — `status_handler`/`register_status` were dropped. Please update the description to match what's being merged.

abdulrafey1:
Thanks @regisb — addressed all three suggestions and the description drift.

**1. `TypeKeyedHook` duplicated `KeyedItemHook`'s shape** (`9bf338f`) — took the `KeyedItemHook` route: dropped `TypeKeyedHook` entirely and folded the exception registry onto the existing `KeyedItemHook`. As you noted, the handler closure doesn't carry its own `exc_type`, so the registry now stores `(exc_type, handler)` entries keyed on `exc_type` (`key=lambda entry: entry[0]`), plus a new `KeyedItemHook.iter_values()` for `assemble_app` to iterate. Fully unifying the two remaining keyed-registry hooks (`KeyedClassHook` + `KeyedItemHook`) into one caller-passes-key primitive is a larger cross-consumer change (audit + analytics + exceptions), so I tracked it separately in #522 rather than expand this PR.

**2. Docstring described another module's behavior** (`9bf338f`) — trimmed the `assemble_app` wiring detail from `register_exception_handler`'s docstring (and from the module comment), scoping both to what `handlers.py` owns.

**3. Competing exception→HTTP philosophies side-by-side** (`504514b`) — added a module docstring to `sparkth/lib/exceptions/__init__.py` marking the legacy `status_code`-carrying `HttpError`/`AuthenticationError`/`LMSRequestError` family as pre-standard, and pointing new code to `register_exception_handler` instead.

**PR description drift** — rewrote the description to match the shipped diff: `sparkth/core/exceptions/handlers.py` / `sparkth/lib/exceptions/handlers.py`, only `register_exception_handler` (the `status_handler`/`register_status` shorthand was never shipped).

All green: `mypy --strict`, `ruff` lint + format, full suite (1420 passed).

sparkth/core/exceptions/handlers.py
```diff
@@ -0,0 +1,41 @@
...
+# The single registry of exception type -> HTTP handler. Each entry is an
+# ``(exc_type, handler)`` pair keyed on ``exc_type`` (KeyedItemHook derives the key from
+# the entry), so a second registration for the same type raises ``ValueError``. Ships
+# empty; core and plugins populate it at import/plugin-load time.
+EXCEPTION_HANDLERS: KeyedItemHook[type[Exception], tuple[type[Exception], ExceptionHandler]] = KeyedItemHook(
+    key=lambda entry: entry[0]
+)
+
+
+def register_exception_handler(exc_type: type[Exception], status_code: int) -> None:
```
hamza-56, line 33: Nothing validates `status_code`, so a typo like `20` or `200` registers fine and only shows up later as a confusing runtime response. A quick `if not 400 <= status_code <= 599: raise ValueError(...)` at registration time would catch that early, and it doubles as documentation that this registry is meant for error statuses only.

sparkth/core/exceptions/handlers.py
```diff
@@ -0,0 +1,41 @@
...
+    """Map ``exc_type`` (and, via MRO, its subclasses) to HTTP ``status_code``.
+
+    Registers a handler on the global ``EXCEPTION_HANDLERS`` registry that renders the
+    exception as ``{"detail": str(exc)}``. A second registration for the same type raises
+    ``ValueError``.
+    """
+
+    async def handler(request: Request, exc: Exception) -> JSONResponse:
+        return JSONResponse(status_code=status_code, content={"detail": str(exc)})
```
hamza-56, line 54: One thing that worries me here: the handled exception is never logged. Starlette's `ExceptionMiddleware` doesn't log exceptions that a handler renders, so once a type is mapped, its occurrences leave no trace in the logs. A spike of 409s in production would be hard to debug.

sparkth/core/exceptions/handlers.py
```diff
@@ -0,0 +1,41 @@
...
+EXCEPTION_HANDLERS: KeyedItemHook[type[Exception], tuple[type[Exception], ExceptionHandler]] = KeyedItemHook(
+    key=lambda entry: entry[0]
+)
+
+
+def register_exception_handler(exc_type: type[Exception], status_code: int) -> None:
+    """Map ``exc_type`` (and, via MRO, its subclasses) to HTTP ``status_code``.
+
+    Registers a handler on the global ``EXCEPTION_HANDLERS`` registry that renders the
+    exception as ``{"detail": str(exc)}``. A second registration for the same type raises
```
hamza-56, line 37: Since `str(exc)` goes to the client verbatim, whatever the service layer puts in the message becomes API-visible the moment the type is registered. Could we call out in the docstring that registered exception types must carry client-safe messages (no internal IDs, paths, or SQL)?

sparkth/core/exceptions/handlers.py
```diff
@@ -0,0 +1,41 @@
...
+    ``ValueError``.
+    """
+
+    async def handler(request: Request, exc: Exception) -> JSONResponse:
+        return JSONResponse(status_code=status_code, content={"detail": str(exc)})
+
+    EXCEPTION_HANDLERS.add_item((exc_type, handler))
```
hamza-56, line 56: The duplicate check guards the registry against itself, but not against FastAPI's built-ins: registering `HTTPException` or `RequestValidationError` here would silently override FastAPI's default handlers when `assemble_app` wires the registry, losing `HTTPException.headers` support and the standard 422 body shape. Probably worth rejecting `starlette.exceptions.HTTPException` subclasses in this function.

hamza-56 (commented):
_These review comments were written with the assistance of an LLM (Claude)._

sparkth/core/exceptions/handlers.py
hamza-56, line 54: The registry stores opaque handler closures, so the status code is not introspectable after registration. That means tests can only assert that a type is registered, never the actual type to status contract (e.g. that `WhitelistEntryAlreadyExists` maps to 409), the duplicate-registration `ValueError` cannot report which status it conflicts with, and the `{"detail": str(exc)}` body shape gets baked in at registration time instead of at the wiring layer.

Have you considered storing data instead? Keep the registry as exception type to status code (plain `int`) and build the handler once inside `_register_exception_handlers` in `sparkth/main.py`. Registration stays one line for callers, the rendering policy lives in exactly one place, and tests can assert `registered[SomeError] == 409`.

sparkth/core/exceptions/handlers.py
zaira-bibi: Leaving this for @abdulrafey1 to review and consider.

sparkth/core/exceptions/handlers.py
zaira-bibi: Leaving this for @abdulrafey1 to review and consider.

hamza-56 (approved):

sparkth/core/exceptions/handlers.py
abdulrafey1: This is documented in the issue #493 

> Store type → handler, not type → int. Storing a handler … lets plugins return richer bodies, headers, or error codes when they need to, without changing the mechanism.

We decided on not leaking HTTP status codes into the the exception management system.

PR #517, opened 2026-07-08
refactor(whitelist): map domain errors via central exception registry

### LINKED ISSUE #494: Whitelist routes map domain errors with inline per-route try/except
## Problem

`app/api/v1/whitelist.py` translates errors into HTTP responses with inline
`try/except` in each handler — catching `ValueError` and `IntegrityError` and
re-raising them as `HTTPException(409)` / `HTTPException(422)` / `HTTPException(404)`.
The mapping is repeated per endpoint and lives inside the business-logic routes rather
than in one shared place.

This is an instance of the broader problem in #493 (routes need a centralized,
reusable way to map domain exceptions to HTTP responses), scoped to the whitelist
module.

## Why it matters

- **Repetition:** the same error→status mapping is duplicated across the whitelist handlers.
- **Consistency:** a missed `except` would leak a `500` for what should be a `4xx`.
- **Coupling:** HTTP status decisions sit in the routes instead of a shared API-layer mapping.

## Scope

Migrate `app/api/v1/whitelist.py` to the centralized exception-handling mechanism from
#493 and drop the inline `try/except` from the handlers.

## Depends on

- #493 — the centralized mechanism must land first.

abdulrafey1:
## What
Migrate `sparkth/api/v1/whitelist` off inline `try/except → HTTPException` onto the centralized exception→HTTP registry from #493 (PR #513), so HTTP-status decisions live in one API-layer mapping instead of being duplicated (and message-string-matched) inside each route.

## Changes
- `refactor(whitelist)`: split the two single-file whitelist modules into packages — `sparkth/api/v1/whitelist/` (`routes.py` + `__init__.py`) and `sparkth/services/whitelist/` (`service.py`, `exceptions.py`, `__init__.py`) — with back-compat re-exports so existing import paths are unchanged.
- `refactor(whitelist)`: add HTTP-agnostic domain exceptions (`WhitelistError` base + `InvalidWhitelistValue`, `WhitelistEntryAlreadyExists`, `WhitelistEntryNotFound`); the service raises them instead of `ValueError`.
- `refactor(whitelist)`: move the `IntegrityError` unique-index race backstop into the service (rollback + log-with-context + re-raise as `WhitelistEntryAlreadyExists`), fixing the old inline catch that logged nothing.
- `refactor(api)`: register the three concrete exception→status mappings (`409`/`404`/`422`) at import in the whitelist API package via `register_exception_handler`; drop all `try/except` / `HTTPException` from the route handlers.
- `test(whitelist)`: switch service tests to the new exception types, add an `IntegrityError` race-path test (asserts rollback + the conflict log), assert exact API error bodies, and assert the mappings are registered.

## How to Test
1. `make test.backend.pytest` — full suite passes (1364 passed). New/updated coverage in `tests/services/test_whitelist.py` and `tests/api/v1/test_whitelist.py`.
2. `make mypy` — clean (`--strict`).
3. `make lint.backend` and `make lint.format.backend check=1` — both clean.
4. Behaviour check against a running backend (as a user with the whitelist permissions):
   - `POST /api/v1/whitelist/` with a value already present → `409` `{"detail": "Entry already exists: <value>"}`.
   - `POST /api/v1/whitelist/` with an invalid value (e.g. `not-valid`) → `422` `{"detail": "Invalid email format: not-valid"}`.
   - `DELETE /api/v1/whitelist/{missing_id}` → `404` `{"detail": "Whitelist entry not found: <id>"}`.

## Notes
- **Stacked PR:** based on `rafey/exception-to-http-code-mapping` (PR #513), not `main`. Merge #513 first.
- **Minor behaviour change (intentional):** the `404` `detail` body changes from the old hard-coded `"Whitelist entry not found"` to `"Whitelist entry not found: <id>"`. The old delete route discarded the service message; the new path renders `str(exc)`, which is consistent with the `409`/`422` bodies (they already echo the offending value). Non-breaking — status is still `404` and no consumer reads this body.
- **Test removed from #513's suite:** `tests/core/exceptions/test_handlers.py::test_global_registry_ships_empty` is deleted here. Its premise ("the registry ships empty — this issue delivers only the mechanism") is deliberately ended by this PR: whitelist is the registry's first consumer, registering mappings at import. The `_restore_registry` fixture docstring was updated to match.
- No migration, no new env var, no dependency bump.


abdulrafey1:
@claude-review

sparkth/services/whitelist/service.py
```diff
@@ -30,26 +39,31 @@ async def add_entry(
...
             entry_type = "domain"
         else:
             try:
                 _EmailValidator(email=normalized)
             except ValidationError as exc:
-                raise ValueError(f"Invalid email format: {value}") from exc
+                raise InvalidWhitelistValue(f"Invalid email format: {value}") from exc
             entry_type = "email"
 
         result = await session.exec(select(WhitelistedEmail).where(WhitelistedEmail.value == normalized))
```
hamza-56: Now that `value` has `unique=True` and the commit below translates `IntegrityError` into `WhitelistEntryAlreadyExists`, this pre-check SELECT is a redundant round trip.

sparkth/services/whitelist/service.py
```diff
@@ -30,26 +39,31 @@ async def add_entry(
...
         entry = WhitelistedEmail(
             value=normalized,
             entry_type=entry_type,
             added_by_id=added_by_id,
         )
         session.add(entry)
-        await session.commit()
+        try:
+            await session.commit()
+        except IntegrityError as exc:
```
hamza-56, line 62: This blanket translation assumes every `IntegrityError` is a duplicate `value`, but the insert can also violate the FK on `added_by_id` (e.g. the user was deleted mid-request). On Postgres that failure would surface as a 409 "Entry already exists" and get logged as an insert conflict, which is misleading.

sparkth/services/whitelist/service.py
```diff
@@ -58,7 +72,7 @@ async def remove_entry(session: AsyncSession, *, entry_id: int) -> None:
         result = await session.exec(select(WhitelistedEmail).where(WhitelistedEmail.id == entry_id))
```
hamza-56: nit: `select(...).where(WhitelistedEmail.id == entry_id)` plus `one_or_none()` is the long way to spell `await session.get(WhitelistedEmail, entry_id)`, which is the idiomatic primary-key lookup and also uses the identity map.

zaira-bibi (approved):

zaira-bibi:
Failing playwright test is preexisting, issue and fix reported in #524.

hamza-56 (approved):

PR #518, opened 2026-07-09
refactor(permissions): map role domain errors via central registry

### LINKED ISSUE #495: Role-management routes map domain exceptions with per-route try/except
## Problem

The role-management routes (added in #473, `app/api/v1/permissions/routes/roles.py`)
translate domain exceptions into HTTP responses with ~10 near-identical `try/except`
blocks across the endpoints: `RoleNotFound → 404`, `RoleAlreadyExists → 409`,
`RoleInUse → 409`, `PermissionNotFound → 422`. The same mapping is repeated per endpoint.

This is an instance of the broader problem in #493 (routes need a centralized, reusable
way to map domain exceptions to HTTP responses), scoped to the permissions module.

## Why it matters

- **Repetition:** the same domain-exception→status mapping is duplicated across the role handlers.
- **Consistency:** a missed `except` would leak a `500` for what should be a `4xx`.
- **Coupling:** HTTP status decisions sit in the routes instead of a shared API-layer mapping.

## Scope

Migrate the role-management routes to the centralized mechanism from #493 and delete the
inline `try/except` from the handlers. The permission domain exceptions are already
HTTP-agnostic, so they map cleanly onto the hook.

## Depends on

- #493 — the centralized mechanism must land first.
- #473 — introduces the role routes being migrated.

abdulrafey1:
## What
Move the role-management routes' HTTP status decisions out of ~7 repeated per-route `try/except → HTTPException` blocks and into a single API-layer mapping registered on the central exception registry (issue #495), so a missed `except` can no longer leak a `500` for what should be a `4xx`.

## Changes
- refactor(permissions): register role domain exceptions centrally
- refactor(permissions): map role domain errors via central registry
- test(permissions): pin exact error bodies for the migrated routes

The permission domain exceptions were already HTTP-agnostic, so no new exception
classes were needed — the migration registers the existing ones and deletes the
inline handlers:

- `RoleNotFound → 404`
- `RoleAlreadyExists → 409`
- `RoleInUse → 409`
- `PermissionNotFound → 422`

## How to Test
1. Check out the branch and run the affected tests:
   `uv run pytest tests/api/v1/test_permissions.py -v` — all pass, including the new
   `TestPermissionExceptionRegistration` and the strengthened `detail`-body assertions.
2. Confirm the mapping is wired end-to-end (not caught inline): the role routes in
   `sparkth/api/v1/permissions/routes/roles.py` contain no `try/except`/`HTTPException`,
   and `sparkth/api/v1/permissions/__init__.py` registers the four mappings via
   `register_exception_handler`.
3. Run the full backend suite + checks: `make test.backend.pytest` (1365 passed),
   `make mypy` (clean), `make lint.backend` and `make lint.format.backend check=1` (clean).

## Notes
- **Stacked on #517** (`rafey/migrate-whitelist-routes-to-central-exception-management`,
  issue #494). This PR targets that branch, not `main`; review/merge #517 first. Both
  are per-module migrations onto the central exception mechanism from #493 (PR #513).
- No migration, no new env var, no dependency change, no breaking change — response
  status codes and `{"detail": ...}` bodies are byte-for-byte identical to the previous
  inline handlers.
- The registration is app-wide by design (Starlette dispatches by MRO). Verified the four
  exception types are reached only through the migrated role routes, the CLI (which catches
  `RoleNotFound` locally), and tests — no other endpoint's behavior shifts.


abdulrafey1:
@claude-code

zaira-bibi (approved):

zaira-bibi (approved):

PR #523, opened 2026-07-12
feat(audit): record ai tool executions on every execution path

### LINKED ISSUE #501: AI tool executions are not audited on any execution path
# Summary

Part of epic #435.

AI tool executions, the core of the product, are completely opaque: no persistent record exists of who invoked which tool, when, with which inputs, from which model, or with what outcome. This includes every state-changing operation against downstream LMSes (`canvas_create_*`, `openedx_create_*`, and the rest).

# Where it happens

Tool *registration* is unified via the `MCP_TOOLS` hook, but *execution* fans out across several independent paths, none of which records anything:

- The FastMCP server mounted at `/ai/mcp` (`app/mcp/main.py`).
- The chat LangChain loop's tool execution (`app/llm/providers.py`), the dominant path. The chat plugin persists `Conversation` and `Message` rows (`app/core_plugins/chat/models.py`) but not the tool executions, model identity, or model version involved in producing a response.
- The RAG LangGraph agent, which builds LangChain `StructuredTool`s directly (`app/rag/mcp/agent_tools.py`) and bypasses `Tool.handler`, serving both chat and Slack.
- Auxiliary LLM calls (scope classifier / intent router, title generation, Slack synthesis).

A design that instruments only one or two of these paths silently misses the others.

# Why it matters

The EU AI Act (Art. 12, Annex III point 3) makes event logging a design obligation for AI systems that evaluate learning outcomes. Operationally, incident-response questions like "which courses did this compromised account modify through the Canvas tools last week?" currently have no answer in the system. No incumbent LMS audits AI actions; this is the epic's central differentiator.

# How to observe

1. `make backend.up.dev`, then invoke any Canvas or Open edX tool through chat or the MCP endpoint.
2. `make db-shell`: no record of the invocation, its arguments, the model involved, or its outcome exists anywhere.

hamza-56:
## What
Every AI tool execution, on every execution path (the FastMCP server on `/ai/mcp`, the chat LangChain loop, and the RAG LangGraph agent), now leaves an append-only audit record with actor, origin, arguments, model identity, and outcome, closing the epic #435 gap for AI actions.

## Changes
- feat(core): `audited_tool_handler` wrapper records a fail-closed `tool.invoked` event before the handler runs (a failed write refuses the call, per ADR-0002 / NIST AU-5) plus a `tool.completed` or `tool.failed` outcome event; both share a generated invocation id on `target` so a crash mid-execution is detectable
- feat(core): new `tool.invoked` / `tool.completed` / `tool.failed` event types on the `AUDIT_EVENTS` hook, and `ai_audit_context` to stamp the surface (`chat` / `rag`) and driving model onto tool executions
- feat(core): the `Tool` hook dataclass audit-wraps its handler at construction, covering the FastMCP server and the chat tool registry with one seam; wrapping preserves name, docstring, and generated input schema (regression-tested)
- feat(core): the chat provider loop records model provenance (provider, configured model, provider-reported model version) with source `chat`
- feat(rag): the RAG agent tools (which bypass `Tool.handler`) wrap explicitly; the agent runner stamps source `rag` with best-effort model identity
- feat(mcp): `ToolCallAuditMiddleware` backstop records protocol-level failures that never reach a handler (unknown tool, input validation) without double-recording handler failures; the audit context middleware stamps source `mcp` for the `/ai` mount
- test(core): completeness test (ADR-0002 §6) asserts every registered server tool and RAG tool is audited and pins the set of modules allowed to construct executable tools, so new execution paths cannot appear unaudited
- docs: plugin guide section on audited tool executions; CLAUDE.md audit directory entry

## How to Test
2. Call any tool on the MCP endpoint, e.g. with a FastMCP client against `http://localhost:7727/ai/mcp`: `call_tool("get_course_generation_prompt_tool", {"course_params": {"course_name": "Algebra", "course_description": "Linear equations"}})`
3. `make db-shell`, then `SELECT category, action, source, outcome, tool_name, target_id, tool_args FROM audit_events ORDER BY id;` shows the `tool.invoked` / `tool.completed` pair with `source=mcp`, a shared `target_id`, and captured request IP / request id
4. Call an unknown tool name: a single `tool.failed` row appears (protocol-level backstop)
5. Call `canvas_authenticate` with a bogus URL and token: the `auth` payload is stored as `"[REDACTED]"` and the failed execution records `tool.invoked` + `tool.failed`
6. Tool calls made through chat carry `source=chat` plus `model_provider` / `model_name` / `model_version` (covered by `tests/llm/test_provider_tool_audit.py`)
7. `make test.backend.pytest && make mypy && make lint.backend` all pass (1439 tests)

## Notes
No migration: the `audit_events` AI-provenance columns shipped with #492. Fail-closed semantics mean an audit-database outage refuses AI tool calls instead of running them unrecorded; this is the deliberate ADR-0002 trade-off. Argument-validation refusals inside the chat tool conversion layer (before a handler is reached) are not yet recorded; auxiliary LLM calls without tools (title generation, scope classifier, Slack synthesis) are generation-level provenance and remain for a follow-up issue.

_This pull request (code and description) was written with the assistance of an LLM (Claude)._

### DIFF
--- a/sparkth/core/audit/execution.py
+++ b/sparkth/core/audit/execution.py
@@ -1,0 +1,146 @@
+"""Audited execution of AI tool handlers.
+
+The single wrapper every tool-execution path goes through (ADR-0002 seam
+table): plugin tools are wrapped at :class:`sparkth.lib.mcp.hooks.Tool`
+construction, the RAG agent tools and the FastMCP server's directly-registered
+tools wrap explicitly. Failure semantics are fail-closed (NIST AU-5): a
+``tool.invoked`` event is committed *before* the handler runs, so a write
+failure refuses the call and a crash mid-execution still leaves the
+invocation on record; a second ``tool.completed`` / ``tool.failed`` event
+records the outcome. Corrections are new events, never updates.
+"""
+
+import functools
+import inspect
+from collections.abc import Awaitable, Callable, Iterator, Mapping
+from contextlib import contextmanager
+from contextvars import ContextVar
+from typing import Any
+from uuid import uuid4
+
+from pydantic import BaseModel
+
+from sparkth.core.audit.constants import MAX_DEPTH, REDACTED, TOOL_INVOCATION_TARGET_TYPE
+from sparkth.core.audit.context import current_model_info
+from sparkth.core.audit.enums import AuditOutcome
+from sparkth.core.audit.events import ToolCompletedAuditEvent, ToolFailedAuditEvent, ToolInvokedAuditEvent
+from sparkth.core.audit.recorder import record_event_now
+from sparkth.core.audit.types import AuditTarget, AuditToolCall
+
+AsyncToolHandler = Callable[..., Awaitable[Any]]
+
+# Whether an audited handler has recorded the tool call currently being served.
+# Scoped per protocol-level call by tool_execution_recording(); the FastMCP
+# middleware uses it to record failures that never reached a wrapped handler
+# without ever double-recording ones that did.
+_execution_recorded: ContextVar[bool] = ContextVar("audit_tool_execution_recorded", default=False)
+
+
+@contextmanager
+def tool_execution_recording() -> Iterator[Callable[[], bool]]:
+    """Track whether an audited handler recorded the enclosed tool call.
+
+    Yields a callable that returns True once a wrapped handler has written its
+    ``tool.invoked`` event inside the block. Protocol-layer backstops (the
+    FastMCP middleware) use it to record exactly the failures the handler
+    wrapper could not see.
+    """
+    token = _execution_recorded.set(False)
+    try:
+        yield _execution_recorded.get
+    finally:
+        _execution_recorded.reset(token)
+
+
+def audited_tool_handler(handler: AsyncToolHandler) -> AsyncToolHandler:
+    """Wrap an async tool handler so every execution is audited.
+
+    Wrapping preserves the handler's name, docstring, signature, and type
+    hints, so generated input schemas are byte-identical post-wrap (the
+    ADR-0002 schema-identity requirement). Model identity is read from the
+    ambient :func:`sparkth.core.audit.context.ai_audit_context`; origin and
+    actor come from the audit context as usual. Already-wrapped handlers are
+    returned unchanged, so seams can wrap defensively.
+
+    Raises (from the returned wrapper):
+        sqlalchemy.exc.SQLAlchemyError: The ``tool.invoked`` write failed; the
+            handler was **not** executed (fail-closed refusal).
+    """
+    if getattr(handler, "__audit_wrapped__", False):
+        return handler
+
+    @functools.wraps(handler)
+    async def audited(*args: Any, **kwargs: Any) -> Any:
+        call = AuditToolCall(name=handler.__name__, args=_named_jsonable_args(handler, args, kwargs))
+        model = current_model_info()
+        target = AuditTarget(type=TOOL_INVOCATION_TARGET_TYPE, id=uuid4().hex)
+
+        await record_event_now(
+            ToolInvokedAuditEvent(outcome=AuditOutcome.SUCCESS, tool=call, model=model, target=target)
+        )
+        _execution_recorded.set(True)
+        try:
+            result = await handler(*args, **kwargs)
+        # NOTE: Kept broad intentionally: handlers are arbitrary plugin code
+        # that can raise anything, the failure is recorded as the audit event
+        # this except exists for, and the exception is always re-raised.
+        except Exception as exc:
+            await record_event_now(
+                ToolFailedAuditEvent(
+                    outcome=AuditOutcome.FAILURE,
+                    tool=call,
+                    model=model,
+                    target=target,
+                    error_detail=f"{type(exc).__name__}: {exc}",
+                )
+            )
+            raise
+        await record_event_now(
+            ToolCompletedAuditEvent(outcome=AuditOutcome.SUCCESS, tool=call, model=model, target=target)
+        )
+        return result
+
+    # setattr keeps mypy strict happy: functions have no declared attributes.
+    setattr(audited, "__audit_wrapped__", True)
+    return audited
+
+
+def _named_jsonable_args(handler: AsyncToolHandler, args: tuple[Any, ...], kwargs: dict[str, Any]) -> dict[str, Any]:
+    """Map the call's arguments to parameter names and JSON-native values.
+
+    Positional arguments are named via the handler's signature so the recorded
+    payload is stable however the seam invoked the handler; if binding fails
+    (the call would TypeError anyway) positional values are kept under
+    ``__args__`` rather than dropped.
+    """
+    if args:
+        try:
+            named = dict(inspect.signature(handler).bind_partial(*args, **kwargs).arguments)
+        except TypeError:
+            named = {"__args__": list(args), **kwargs}
+    else:
+        named = dict(kwargs)
+    return {key: _jsonable(value, MAX_DEPTH) for key, value in named.items()}
+
+
+def _jsonable(value: Any, depth: int) -> Any:
+    """Reduce ``value`` to JSON-native types for redaction and canonicalization.
+
+    Pydantic models are dumped (so nested credential fields stay visible to the
+    redactor as mappings), containers recurse with the same depth bound the
+    redactor uses, and anything else non-native falls back to ``str`` (an
+    audit record of an odd value beats a refused tool call).
+    """
+    if isinstance(value, BaseModel):
+        return _jsonable(value.model_dump(mode="json"), depth)
+    if isinstance(value, Mapping):
+        if depth <= 0:
+            return REDACTED
+        return {str(key): _jsonable(inner, depth - 1) for key, inner in value.items()}
+    if isinstance(value, (list, tuple, set, frozenset)):
+        if depth <= 0:
+            return REDACTED
+        return [_jsonable(item, depth - 1) for item in value]
+    if value is None or isinstance(value, (str, int, float, bool)):
+        return value
+    return str(value)
--- a/sparkth/core/audit/execution.py
+++ b/sparkth/core/audit/execution.py
@@ -63,7 +63,7 @@ def tool_execution_recording() -> Iterator[Callable[[], bool]]:
         _execution_recorded.reset(token)
 
 
-def audited_tool_handler(handler: AsyncToolHandler) -> AsyncToolHandler:
+def audited_tool(handler: AsyncToolHandler) -> AsyncToolHandler:
     """Wrap an async tool handler so every execution is audited.
 
     Wrapping preserves the handler's name, docstring, signature, and type
@@ -91,7 +91,7 @@ def audited_tool_handler(handler: AsyncToolHandler) -> AsyncToolHandler:
         return handler
     if not inspect.iscoroutinefunction(handler):
         raise TypeError(
-            f"audited_tool_handler requires an async handler; "
+            f"audited_tool requires an async handler; "
             f"'{getattr(handler, '__name__', handler)}' is not a coroutine function"
         )
 
--- a/sparkth/core/audit/execution.py
+++ b/sparkth/core/audit/execution.py
@@ -5,9 +5,14 @@ table): plugin tools are wrapped at :class:`sparkth.lib.mcp.hooks.Tool`
 construction, the RAG agent tools and the FastMCP server's directly-registered
 tools wrap explicitly. Failure semantics are fail-closed (NIST AU-5): a
 ``tool.invoked`` event is committed *before* the handler runs, so a write
-failure refuses the call and a crash mid-execution still leaves the
+failure refuses the call (surfaced as the distinct
+:class:`~sparkth.core.audit.exceptions.AuditCaptureError` so execution seams
+never mistake it for a tool error) and a crash mid-execution still leaves the
 invocation on record; a second ``tool.completed`` / ``tool.failed`` event
-records the outcome. Corrections are new events, never updates.
+records the outcome. A ``tool.failed`` write that itself fails is logged and
+suppressed: the invocation is already on record, so the handler's own
+exception surfaces and the missing outcome event is the abnormal-termination
+signal. Corrections are new events, never updates.
 """
 
 import functools
@@ -19,13 +24,19 @@ from typing import Any
 from uuid import uuid4
 
 from pydantic import BaseModel
+from sqlalchemy.exc import SQLAlchemyError
 
 from sparkth.core.audit.constants import MAX_DEPTH, REDACTED, TOOL_INVOCATION_TARGET_TYPE
 from sparkth.core.audit.context import current_model_info
 from sparkth.core.audit.enums import AuditOutcome
 from sparkth.core.audit.events import ToolCompletedAuditEvent, ToolFailedAuditEvent, ToolInvokedAuditEvent
+from sparkth.core.audit.exceptions import AuditCaptureError
 from sparkth.core.audit.recorder import record_event_now
+from sparkth.core.audit.redaction import scrub_error_detail
 from sparkth.core.audit.types import AuditTarget, AuditToolCall
+from sparkth.lib.log import get_logger
+
+logger = get_logger(__name__)
 
 AsyncToolHandler = Callable[..., Awaitable[Any]]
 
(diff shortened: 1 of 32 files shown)


hamza-56:
@claude-review

sparkth/mcp/audit.py
```diff
@@ -0,0 +1,52 @@
...
+        with tool_execution_recording() as handler_recorded:
+            try:
+                return await call_next(context)
+            # NOTE: Kept broad intentionally: any protocol-layer error
+            # (unknown tool, validation, handler crash) must be considered
+            # for the backstop record, and is always re-raised.
+            except Exception as exc:
+                if not handler_recorded():
+                    await self._record_protocol_failure(context.message, exc)
+                raise
```
zaira-bibi, line 45: When `call_next` fails because the wrapper's `tool.invoked` write raised, `handler_recorded()` is `False`, so this issues another `record_event_now` against the same (down) audit DB. That write also raises, replacing the original exception and leaving no record, which is the opposite of the module docstring's `"every tools/call … leaves a record either way."` Consider catching the backstop write's own failure (log it, preserve/chain the original exc) so an audit-store outage doesn't swap the surfaced error and lose the trail.

sparkth/llm/providers.py
```diff
@@ -205,7 +226,8 @@ async def _send_message_with_tools(
                         continue
 
                     logger.debug(f"Executing tool: {tool_name} with args: {tool_args}")
-                    tool_result = await self._execute_tool(tool_name, tool_args, tools)
+                    with ai_audit_context(source=AuditSource.CHAT, model=self._audit_model_info(response)):
+                        tool_result = await self._execute_tool(tool_name, tool_args, tools)
```
zaira-bibi, line 235: Caught by Claude:

This PR installs `ai_audit_context` around tool execution to make capture fail-closed, but the wrapper's refusal exception is absorbed before it can fail the request:

- Chat: `_execute_tool's` broad `except Exception` at `providers.py:324-326` catches the wrapper's `SQLAlchemyError` and returns it as a string; `tool_func` at `plugins/chat/tools.py:186` only catches (`ValidationError`, `ValueError`, `TypeError`, `RuntimeError`), so the DB error flows straight to that broad handler.
- RAG: `create_agent(...)` at `agent.py:65` runs tools through LangGraph's ToolNode (default `handle_tool_errors=True`), which converts the exception into an error `ToolMessage`; `run_agentic_rag_retrieval` only re-raises `ValidationError`/`GraphRecursionError`/`LangChainException`.

Net: when the `tool.invoked` write fails, the call is refused (side effect prevented) but there's no audit record and no surfaced failure — it looks like an ordinary tool error and chat/RAG continue. That defeats "record on every execution path" for exactly these two seams. Suggest letting audit-infra exceptions propagate as a distinct hard failure past both seams, or documenting chat/RAG capture as best-effort.

sparkth/core/audit/execution.py
```diff
@@ -0,0 +1,146 @@
...
+        )
+        _execution_recorded.set(True)
+        try:
+            result = await handler(*args, **kwargs)
+        # NOTE: Kept broad intentionally: handlers are arbitrary plugin code
+        # that can raise anything, the failure is recorded as the audit event
+        # this except exists for, and the exception is always re-raised.
+        except Exception as exc:
+            await record_event_now(
+                ToolFailedAuditEvent(
```
zaira-bibi: As this write is before the `raise`, so if it itself raises (e.g. if the DB is down), then the callers/logs will see an audit-infra error instead of the actual failure. 

Perhaps guard the audit write as well:

```
except Exception as exc:
    try:
        await record_event_now(ToolFailedAuditEvent(..., error_detail=f"{type(exc).__name__}: {exc}"))
    except SQLAlchemyError:
        logger.exception("audit tool.failed write failed for %s", handler.__name__)
    raise  
```

sparkth/core/audit/execution.py
```diff
@@ -0,0 +1,146 @@
...
+        call = AuditToolCall(name=handler.__name__, args=_named_jsonable_args(handler, args, kwargs))
+        model = current_model_info()
+        target = AuditTarget(type=TOOL_INVOCATION_TARGET_TYPE, id=uuid4().hex)
+
+        await record_event_now(
+            ToolInvokedAuditEvent(outcome=AuditOutcome.SUCCESS, tool=call, model=model, target=target)
+        )
+        _execution_recorded.set(True)
+        try:
+            result = await handler(*args, **kwargs)
```
zaira-bibi, line 113: `await handler(*args, **kwargs)` assumes a coroutine, but `Tool.__post_init__` (`lib/mcp/hooks.py:34-35`) wraps every `Tool.handler` and both FastMCP and chat's `sync_tool_func` (`plugins/chat/tools.py:190`) support sync handlers. A sync handler would write `tool.invoked`, then hit `TypeError: object … can't be used in 'await'`, recorded as `tool.failed` on every call. Latent today (all canvas/openedx handlers are async), but the wrapper narrows the accepted contract without a guard — consider `inspect.iscoroutinefunction` (fail loudly at wrap time) or supporting sync handlers.

_Caught by Claude._

zaira-bibi (changes requested):

sparkth/core/audit/execution.py
```diff
@@ -0,0 +1,146 @@
...
+    Raises (from the returned wrapper):
+        sqlalchemy.exc.SQLAlchemyError: The ``tool.invoked`` write failed; the
+            handler was **not** executed (fail-closed refusal).
+    """
+    if getattr(handler, "__audit_wrapped__", False):
+        return handler
+
+    @functools.wraps(handler)
+    async def audited(*args: Any, **kwargs: Any) -> Any:
+        call = AuditToolCall(name=handler.__name__, args=_named_jsonable_args(handler, args, kwargs))
```
regisb, line 100: What if two different handlers have the same `__name__ `, but in different modules? Will they be tracked under the same name as well?

sparkth/mcp/audit.py
```diff
@@ -0,0 +1,52 @@
+"""Protocol-level audit backstop for the FastMCP server.
+
+Successful tool calls (and handler-raised failures) are recorded by the
+audited handler wrapper (:func:`sparkth.lib.audit.audited_tool_handler`); this
+middleware covers the calls that never reach a handler (unknown tool names
+and input-validation rejections), so every ``tools/call`` on ``/ai/mcp``
+leaves a record either way (ADR-0002 seam table row three).
```
regisb: I don't understand the point of tracking calls to tools that do not exist. This adds extra complexity without any visible gain. The way I see it, this is causing us to record things that did not happen. But I might be missing some context?

And if we do want to keep track of non-existing tool calls, then we should keep this middleware but get rid of the audited_tool_handler decorator. The decorator forces developers to wrap all tools, while the middleware automatically captures all calls. Isn't that a duplicated feature?

sparkth/core/audit/execution.py
```diff
@@ -0,0 +1,146 @@
...
+from sparkth.core.audit.recorder import record_event_now
+from sparkth.core.audit.types import AuditTarget, AuditToolCall
+
+AsyncToolHandler = Callable[..., Awaitable[Any]]
+
+# Whether an audited handler has recorded the tool call currently being served.
+# Scoped per protocol-level call by tool_execution_recording(); the FastMCP
+# middleware uses it to record failures that never reached a wrapped handler
+# without ever double-recording ones that did.
+_execution_recorded: ContextVar[bool] = ContextVar("audit_tool_execution_recorded", default=False)
```
regisb, line 47: This global context variable is a red flag to me. I believe that we need it because of the ToolCallAuditMiddleware, which tracks calls to tools that don't exist. If we get rid of ToolCallAuditMiddleware then we can get rid of this variable and the tool_execution_recording context manager, right?

sparkth/core/audit/middleware.py
```diff
@@ -36,11 +36,19 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+            source=self._source(scope),
         )
         with audit_context(context):
             await self.app(scope, receive, send)
 
+    @staticmethod
+    def _source(scope: Scope) -> AuditSource:
+        """MCP for the FastMCP mount (mounted at ``/ai`` in main.py), REST otherwise."""
+        path: str = scope.get("path", "")
+        if path == "/ai" or path.startswith("/ai/"):
```
regisb: This "/ai" is a hard-coded string that should be replaced by a variable. Unless each router configures its own AuditSource?

sparkth/core/audit/execution.py
```diff
@@ -0,0 +1,146 @@
...
+    wrapper could not see.
+    """
+    token = _execution_recorded.set(False)
+    try:
+        yield _execution_recorded.get
+    finally:
+        _execution_recorded.reset(token)
+
+
+def audited_tool_handler(handler: AsyncToolHandler) -> AsyncToolHandler:
```
regisb: I have a problem with the name of this decorator. If we decide to keep it, and not the middleware, then we should probably rename to something like "audited_tool".

In general "handler"/"manager"/etc. suffixes don't add meaningful to a name and should be either removed or replaced by something more specific. (There are exceptions though)

sparkth/mcp/audit.py
hamza-56: Good catch, fixed in 634b06f4. The backstop write is now wrapped in `try/except SQLAlchemyError`: we log it and re-raise the original error untouched. Also toned down the docstring, since when the audit DB itself is down there is genuinely nothing we can save. Test: `test_backstop_write_failure_does_not_replace_the_original_error`.

sparkth/llm/providers.py
hamza-56: Fixed the chat half in 634b06f4. The wrapper now raises a dedicated `AuditCaptureError` when the invoked/completed write fails, and `_execute_tool` re-raises it before its broad except, so an audit outage fails the request instead of becoming an error string for the model. Test: `test_audit_capture_outage_is_a_hard_failure_not_a_tool_error_string`.

The RAG half turns out not to apply on our pinned versions: langgraph 1.0.7's default handler only converts `ToolInvocationError` into a `ToolMessage` and re-raises everything else (`handle_tool_errors=True` was the old `create_react_agent` default). So the refusal already propagates there; documented it in the docstring.

sparkth/core/audit/execution.py
hamza-56: Fixed in 634b06f4, pretty much as you suggested: the `tool.failed` write is guarded, logged, and the handler's own error re-raised. Since the invocation is already committed at that point, an invoked event with no outcome is exactly the crash signal, so we lose nothing. Test: `test_failed_write_failure_surfaces_the_handler_error`.

sparkth/core/audit/execution.py
hamza-56: Fixed in 634b06f4, went with the fail-loudly option: `TypeError` at wrap time if the handler isn't a coroutine function. Supporting sync handlers would just move the problem (chat's `sync_tool_func` assumes a coroutine too). The plugins guide now says handlers must be async. Test: `test_sync_handler_is_rejected_at_wrap_time`.

sparkth/core/audit/execution.py
```diff
@@ -0,0 +1,146 @@
...
+        # that can raise anything, the failure is recorded as the audit event
+        # this except exists for, and the exception is always re-raised.
+        except Exception as exc:
+            await record_event_now(
+                ToolFailedAuditEvent(
+                    outcome=AuditOutcome.FAILURE,
+                    tool=call,
+                    model=model,
+                    target=target,
+                    error_detail=f"{type(exc).__name__}: {exc}",
```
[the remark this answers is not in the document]
hamza-56: Fixed in 634b06f4. Added `scrub_error_detail` in redaction.py, used by both the wrapper and the backstop: `ValidationError`s are re-rendered without their inputs, secret-keyed assignments (`token=...`, `'api_key': ...`) are masked with the same `SECRET_KEYS` set, and the result is capped at 1000 chars. Tests in test_redaction.py plus `test_handler_error_detail_is_scrubbed` end to end.

tests/mcp/test_server_audit.py
```diff
@@ -0,0 +1,51 @@
...
+    backstop must stay silent: one invocation, one failure record."""
+    async with Client(mcp) as client:
+        with pytest.raises(ToolError):
+            # Invalid input shape passes the protocol layer but fails handler-side
+            # validation before the wrapped handler runs, so this is protocol-level.
+            await client.call_tool("get_course_generation_prompt_tool", {"course_params": "not-a-mapping"})
+
+    rows = await audit_events()
+    failures = [row for row in rows if row.action == "failed"]
+    assert len(failures) == 1
```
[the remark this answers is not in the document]
hamza-56: Added in 634b06f4: `test_runtime_handler_failure_is_recorded_exactly_once` spins up a throwaway FastMCP server with the middleware, calls a tool that raises at runtime, and asserts exactly one `tool.invoked` + one `tool.failed`, i.e. the backstop stayed silent.

sparkth/core/audit/execution.py
hamza-56: Yes, they'd show up under the same `tool_name`; the row only records `handler.__name__`. In practice tool names are already the identity everywhere (the server refuses duplicate registrations and the LLM calls tools by name), so a second same-named tool can't actually be exposed. If you'd like provenance anyway, adding `handler.__module__` as a column is cheap; happy to do it as a follow-up.

sparkth/mcp/audit.py
hamza-56: A rejected `tools/call` is still a request that happened; we record it with `outcome=failure`, not as something that succeeded. Failed attempts (unknown tool, bad input, someone probing tool names) are exactly what an audit trail should have on record, same reason we record failed logins.

They also aren't duplicates, because they cover different ground: chat and RAG run tools in-process and never touch the middleware, so the wrapper is the only capture there; and calls that die before dispatch never reach a wrapped handler, so the middleware is the only capture for those. Developers don't wrap anything by hand either, `Tool.__post_init__` does it automatically.

If we decide pre-dispatch failures aren't worth recording, the middleware (and the contextvar with it) can go, but that's a scope decision on the trail. Happy to discuss.

sparkth/core/audit/execution.py
hamza-56: Right: it only exists so the wrapper and the middleware don't double-record on the MCP surface, so it lives and dies with the middleware. If the other thread lands on dropping the middleware, this and `tool_execution_recording` go with it. While we keep it, a contextvar is the async-safe way to share that one bit without threading state through FastMCP.

sparkth/core/audit/middleware.py
hamza-56: Fixed in 2982a4e1: extracted `MCP_MOUNT_PATH` into `core/config.py`, used by both the mount in main.py and the middleware. Kept it a plain constant since we only have one non-REST mount today; easy to promote to per-router config if that changes.

sparkth/core/audit/execution.py
hamza-56: Agreed, `audited_tool` is better. Holding the rename until the middleware question (other thread) settles, then I'll do it in one pass since it touches the lib surface and the docs.

zaira-bibi (approved):

sparkth/lib/audit/__init__.py
```diff
@@ -15,9 +20,13 @@
...
 """
 
+from sparkth.core.audit.execution import audited_tool_handler
 from sparkth.core.audit.recorder import record_event, record_event_now
+from sparkth.core.audit.redaction import scrub_error_detail
 
 __all__ = [
+    "audited_tool_handler",
     "record_event",
     "record_event_now",
```
abdulrafey1, line 30: Did we not decide to rename this function to a more accurate name?

sparkth/core/audit/redaction.py
```diff
@@ -40,3 +45,30 @@ def _walk(value: Any, depth: int) -> Any:
...
+    """Format ``exc`` for the ``error_detail`` column without leaking secrets.
+
+    The key-based payload redaction cannot see into free text, and exception
+    messages routinely echo the values it was designed to keep out: Pydantic
+    ``ValidationError`` renders the failing ``input_value``, HTTP client
+    errors carry tokenized URLs. Validation errors are re-rendered without
+    their inputs, secret-keyed assignments in any message are masked, and the
+    result is bounded to :data:`MAX_ERROR_DETAIL_LENGTH`.
+    """
+    if isinstance(exc, ValidationError):
```
abdulrafey1, line 66: Why are we handling `ValidationError` differently? Any particular reason?

sparkth/core/audit/execution.py
abdulrafey1: Instead of adding `handler.__module__` as a column could we enforce unique tool names in the MCP hook? WDYS

sparkth/core/audit/execution.py
hamza-56: Renamed to `audited_tool` in 0a40f5c3. The name is right whether or not the middleware stays, so no reason to keep holding it.

sparkth/lib/audit/__init__.py
hamza-56: Yep, done now in 0a40f5c3. I was holding it for the middleware discussion, but the name is right either way.

sparkth/core/audit/redaction.py
hamza-56: pydantic's `str()` prints `input_value=...` for every error, so a validation failure on a credentials payload would echo the secret straight into the message. `errors(include_input=False)` drops the inputs at the source; the regex masking can't reliably catch them in free text.

sparkth/core/audit/execution.py
hamza-56: Names are already unique per surface: the MCP server skips duplicate registrations and the chat registry keys tools by name. The RAG agent tools don't go through the hook at all, so a hook-level check wouldn't cover the one place a `__name__` clash could actually happen. And each row records `source` next to `tool_name`, so reads aren't ambiguous today. If we want duplicate registrations to fail loudly at startup, I'd file that as a separate plugin-loader issue. WDYT?

abdulrafey1 (approved):

PR #526, opened 2026-07-20
test(analytics): add timescaledb lane for continuous-aggregate reads

### LINKED ISSUE #521: TimescaleDB continuous-aggregate read queries are not exercised in CI
## Summary

The analytics read path forks by database dialect, and the **PostgreSQL branch — the one that runs in production — is never executed by any test**. Our test and e2e suites run exclusively on SQLite (`DATABASE_URL=sqlite+aiosqlite`, per `CLAUDE.md` and CI), so every assertion about analytics reads validates a hand-written SQLite imitation, not the code that actually serves production traffic.

## Where

`get_login_activity` in `sparkth/core/analytics/reads.py` switches on the dialect and runs one of two queries:

- **`_PG_SQL`** reads the TimescaleDB continuous aggregate `login_activity_daily`.
- **`_SQLITE_SQL`** aggregates `raw_events` directly, because the continuous aggregate does not exist on SQLite (see `sparkth/migrations/analytics/versions/287f281e6558_create_login_activity_daily_cagg.py`, which is a no-op on non-PostgreSQL binds).

A module-level comment declares that the two variants "must stay semantically identical," but nothing executable enforces it.

## Why it matters

- `_PG_SQL` can ship with a typo'd column, a wrong `to_char` format, or an incorrect aggregation and CI stays green — the query is never run against a real Postgres/Timescale instance.
- The continuous aggregate is created by raw migration DDL (`time_bucket`, `add_continuous_aggregate_policy`, `CREATE MATERIALIZED VIEW ... WITH (timescaledb.continuous)`). If a future migration breaks that DDL, no test catches it: the migration is a no-op on SQLite, so the suite never applies it.
- The "keep both variants identical" invariant is enforced only by code review and eyeballing — it will drift.
- **This multiplies.** Every future rollup (weekly actives, per-course activity, …) adds another untested PG query plus a parallel SQLite imitation, widening the gap with each one.

## How to observe

1. Introduce a deliberate error into `_PG_SQL` in `reads.py` (e.g. rename `login_count` to `login_counts`).
2. Run the full backend suite: `make test.backend.pytest`.
3. Everything passes — the broken production query is never exercised, because tests hit the SQLite branch only.

zaira-bibi:
## What
Adds an opt-in TimescaleDB test lane so the analytics PostgreSQL read path is executed in CI instead of only imitated on SQLite — and fixes a production bug the lane immediately surfaced.

## Changes
- test(analytics): add `@pytest.mark.pg` lane (`tests/analytics/pg/`) that runs the analytics migrations and exercises the real `login_activity_daily` aggregate, incl. a PG↔SQLite parity test
- fix(analytics): cast `:days` to integer in `_PG_SQL` so the login-activity query runs under asyncpg
- ci: add path-gated `analytics-timescale` job with a `timescaledb-ha:pg18` service
- chore: register the `pg` pytest marker; add `make test.backend.pg`
- docs: document the lane in CLAUDE.md (Test Layout)

## How to Test
1. `make services.up` (starts the Timescale container).
2. Create the test DB once: `docker exec -e PGPASSWORD=sparkth_password sparkth-db-1 psql -U sparkth -d sparkth -c "CREATE DATABASE sparkth_analytics_test;"`
3. `make test.backend.pg` → 3 tests pass against real TimescaleDB.
4. `make test.backend.pytest` → full suite passes on SQLite with the 3 pg tests **skipped** (no infra needed).
5. Regression check for the fix: against `main`'s `_PG_SQL`, step 3 fails with `operator does not exist: date >= integer`; with this PR it passes.

## Notes
- **Production bug fix included:** the login-activity endpoint currently 500s on PostgreSQL (`date - ?` can't resolve). No migration or schema change.
- **No new runtime env var.** `ANALYTICS_TEST_PG_URL` is test-only; when unset the lane is skipped.
- CI cost is bounded: the Timescale job is gated (`dorny/paths-filter`) to analytics code, analytics migrations, and the lane itself — deliberately not "a pg test was added," so a migration that breaks the aggregate is still caught.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,5 +66,56 @@ jobs:
 
       # Single high-level target so a CI failure reproduces locally with
       # exactly `make test.backend` (lint + mypy + pytest + format check).
+      # The default suite runs on SQLite; TimescaleDB-only tests (the `pg` marker)
+      # are skipped here and covered by the analytics-timescale job below.
       - name: Run backend tests
         run: make test.backend
+
+  # Always run the TimescaleDB lane.
+  analytics-timescale:
+    name: Analytics TimescaleDB Lane
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    services:
+      timescaledb:
+        image: timescale/timescaledb-ha:pg18
+        env:
+          POSTGRES_USER: sparkth
+          POSTGRES_PASSWORD: sparkth_password
+          POSTGRES_DB: sparkth
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U sparkth -d sparkth"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
+
+    env:
+      # Only the analytics DB needs a real instance; the rest of the app uses the
+      # in-memory SQLite defaults set by sparkth/lib/testing.py. The pg fixtures run the
+      # analytics migrations against this URL (so the continuous aggregate exists).
+      ANALYTICS_TEST_PG_URL: postgresql://sparkth:sparkth_password@localhost:5432/sparkth_analytics_test
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.1.0
+        with:
+          enable-cache: true
+
+      - name: Install the project
+        run: uv sync --locked --dev
+
+      - name: Create analytics test database
+        env:
+          PGPASSWORD: sparkth_password
+        run: psql -h localhost -U sparkth -d sparkth -c "CREATE DATABASE sparkth_analytics_test;"
+
+      - name: Run TimescaleDB test lane
+        run: uv run pytest -m pg tests/analytics/pg -v
--- a/Makefile
+++ b/Makefile
@@ -162,6 +162,11 @@ test.backend.pytest: ## Run backend unit tests (make test.backend.pytest [path] 
 test.backend.format: ## Run backend formatting tests
 	$(MAKE) lint.format.backend check=1
 
+.PHONY: test.backend.analytics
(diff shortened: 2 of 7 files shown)


zaira-bibi:
@claude-review

Makefile
```diff
@@ -162,6 +162,11 @@ test.backend.pytest: ## Run backend unit tests (make test.backend.pytest [path]
 test.backend.format: ## Run backend formatting tests
 	$(MAKE) lint.format.backend check=1
 
+.PHONY: test.backend.pg
+test.backend.pg: ## Run the TimescaleDB test lane (needs ANALYTICS_TEST_PG_URL; see tests/analytics/pg/). Falls back to make services.up's Postgres.
```
abdulrafey1: would `test.backend.analytics` be a better name for this?

.github/workflows/ci.yml
```diff
@@ -66,5 +66,85 @@ jobs:
...
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          # Run when the analytics read/maintenance code, the analytics migrations, or the
+          # pg lane itself change — NOT merely when a pg test is added — so a migration that
+          # breaks the continuous aggregate is caught even in a migration-only PR.
+          filters: |
```
regisb: I think we should just run tests all the time: they don't cost much in terms of time spent and the `changed` list is fragile.

Makefile
```diff
@@ -162,6 +162,11 @@ test.backend.pytest: ## Run backend unit tests (make test.backend.pytest [path]
 test.backend.format: ## Run backend formatting tests
 	$(MAKE) lint.format.backend check=1
 
+.PHONY: test.backend.analytics
+test.backend.analytics: ## Run the TimescaleDB test lane (needs ANALYTICS_TEST_PG_URL; see tests/analytics/pg/). Falls back to make services.up's Postgres.
```
regisb: This should depend on the "services" target.

abdulrafey1 (approved):

PR #528, opened 2026-07-20
refactor(audit): record agent tool runs via the global callback

hamza-56:
## What
Switches agent tool-execution auditing from per-tool decorators to the process-global LangChain callback handler, removing the developer-dependence where forgetting `@audited_tool_handler` on a new agent tool silently skipped auditing.

## Changes
- refactor(core): activate the audit callback handler by default and import it explicitly at app assembly
- refactor(rag): drop the five per-tool `@audited_tool_handler` decorators; the callback now records those runs, and any future agent tool is audited with zero wiring
- refactor(plugins): tag chat's registry tools with `AUDIT_AT_HANDLER_TAG`, since they delegate to hook-wrapped handlers that already record, preventing double events
- test(core): completeness test now asserts the callback is active by default and that RAG tools carry no wrapping (wrapping returning would mean double-recording)
- docs(core): update the seam docstrings in `execution.py`, `agent_tools.py`, and the chat tool registry

The handler-level wrapper and the FastMCP middleware backstop are unchanged: they cover the MCP server path, which LangChain never sees.

## How to Test
1. `uv run pytest tests/audit tests/rag sparkth/plugins/chat/tests` covers the switch-over: exactly one `tool.invoked`/`tool.completed` pair per RAG tool run, no double events for chat tools, callback active by default.
2. `uv run pytest tests/ sparkth/plugins`, `make mypy`, `make lint.backend` all pass.
3. Live check: `make services.up` + `make backend.up.dev`, run a chat message that calls a tool and a RAG retrieval query, then confirm one event pair per call with `source=chat` / `source=rag` in the audit table.

## Notes
No migration, no env vars, no new dependencies. Depends on the previous PR in this stack (the callback handler).

### DIFF
--- a/sparkth/core/audit/callbacks.py
+++ b/sparkth/core/audit/callbacks.py
@@ -6,8 +6,8 @@ individual tool callables, this handler hooks LangChain's callback system (the
 same extension point tracing uses) so every tool run the framework performs is
 recorded automatically, with no per-tool or per-agent wiring. Importing this
 module registers the :data:`audit_tool_callbacks` hook for the whole process
-via ``register_configure_hook``; the handler is active whenever that
-contextvar holds an instance.
+via ``register_configure_hook``, and the contextvar's default instance makes
+the handler active everywhere from then on.
 
 Failure semantics mirror the wrapper (NIST AU-5 fail-closed): ``raise_error``
 plus ``run_inline`` make the ``tool.invoked`` write complete before the tool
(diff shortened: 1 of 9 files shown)


hamza-56:
@claude-review

sparkth/core/audit/callbacks.py
```diff
@@ -150,7 +150,10 @@ async def on_tool_error(
...
 # the ContextVar's value to every run (the mechanism tracing integrations
-# use). Inactive while the value is None; the execution seams switch over to
-# it (and away from per-tool wrapping) by giving it a default instance.
-audit_tool_callbacks: ContextVar[AuditToolCallbackHandler | None] = ContextVar("audit_tool_callbacks", default=None)
+# use). The default instance makes the handler active everywhere without a
+# set() call, which would not propagate across task contexts; tests deactivate
+# it per-context by setting None.
+audit_tool_callbacks: ContextVar[AuditToolCallbackHandler | None] = ContextVar(
+    "audit_tool_callbacks", default=AuditToolCallbackHandler()
+)
```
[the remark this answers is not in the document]
hamza-56: Fixed in the base PR (42f53d8f on #527): the map is capped now, oldest evicted first, so the global instance can't pile up orphans.

sparkth/main.py
```diff
@@ -7,6 +7,7 @@
 from fastapi.staticfiles import StaticFiles
 from starlette.types import Lifespan
 
+import sparkth.lib.audit.callbacks  # noqa: F401  # registers the process-global LangChain audit callback on import
```
abdulrafey1, line 10: we already make this import in the `main.py` file for registration, why do we need to import it here?

hamza-56: The `chat/tools.py` import exists for a different purpose (it needs `AUDIT_AT_HANDLER_TAG` to prevent double-recording), and it only runs if the chat plugin loads. The loader swallows failures, and plugins are meant to be removable, so relying on it would mean a disabled or broken chat plugin silently stops the RAG agent's runs from being audited. The `main.py` import is the one deterministic registration point, so it stays.

hamza-56:
GitHub auto-closed this PR when its base branch was deleted by the #527 merge, and it cannot be reopened after the restack. Continued in #561 (same branch, rebased onto main).

Issue #434, opened 2026-06-12
[EPIC] Sparkth needs an analytics system to support residential learning


regisb:
# Summary

Sparkth is evolving into a platform that supports residential learning and competency-based education (CBE) at scale. Today the platform has no analytics capability: learning events are not collected in any structured way, there is no aggregation pipeline, and there is no visualization layer for learners, instructors or program operators.

# Why it matters

- **CBE runs on evidence.** Assessing progress against competencies requires capturing fine-grained learning events and aggregating them into per-learner, per-competency state. Without a data collection layer, competency-based features cannot be built at all.
- **Residential learning at scale needs operational visibility.** Instructors and program operators need to monitor cohorts: engagement, progress, struggling learners, completion. None of this is observable today.
- **Analytics data must feed back into the product.** Learner-facing progress views and instructor dashboards are application features, not just back-office reporting — so the application itself needs convenient access to aggregated analytics data.

# Current state

The application stores its operational data in PostgreSQL (`app/models/`). There is no event stream, no analytics storage, no aggregation jobs, and no reporting or dashboarding of any kind.

# Constraints

- **Operational simplicity is a hard requirement.** Sparkth is operated by a small team; every additional stateful service (dedicated analytics database, message broker, BI stack) carries a real, recurring operational cost. The design should minimize the number of moving parts.
- **The architecture must scale with the platform** — both in event write volume (many learners producing many events) and in read patterns (dashboards, aggregations) — without requiring a ground-up redesign.

# Open questions

1. **Storage backend.** Should analytics data live in PostgreSQL — and if so, in the same database/instance as the application, or a separate one? Sharing the application database makes analytics data trivially usable from app features, but mixing a high-volume, append-heavy analytics workload with OLTP traffic may create operational problems over time (write contention, vacuum pressure, backup growth, long analytical queries).
2. **Standards.** Should the event model follow an existing learning-analytics standard (xAPI + Learning Record Store, Caliper), or should it be a purpose-built schema? Standards promise interoperability, but their value to Sparkth specifically is unproven.
3. **Aggregation.** How are raw events turned into queryable aggregates (rollup tables, materialized views, scheduled jobs), and with what freshness guarantees?
4. **Visualization.** What serves dashboards — in-app components, an off-the-shelf BI tool, or both?

_This issue description was written with the assistance of an LLM (Claude)._

regisb:
# Proposed architecture and storage backend discussion

## Architecture overview

[...]

PR #533, opened 2026-07-23
feat(permissions): add GET /permissions/can check endpoint

### LINKED ISSUE #487: Analytics are not visible anywhere in the product UI
### Summary
Even once login activity is rolled up and served by a read API, there is no in-app surface that shows it. Analytics must be integrated into the product, not a separate back-office tool.

### Why it matters
- Feeding analytics back into the product is an explicit requirement of the epic — dashboards are meant to be first-class in-app features over the read API.
- This first dashboard also closes the vertical slice end-to-end (event → rollup → read API → UI), validating the whole read path on a real event.

### What the login-analytics UI needs to show
- **Daily login activity over a recent window** (the read API serves the last N days, default 30, from the `user.logged_in` rollup) — as a simple chart a viewer can scan for trends.
- **At-a-glance summary figures** derived from the same data, e.g. total logins in the window and the busiest day — the milestone exit criterion asks for "a login-activity number/chart".
- **Honest gaps**: the read API omits days with zero logins by contract, so the UI must render those days as zero rather than skipping them or misrepresenting the series.

### Access
- The read API is gated on the `analytics.read` permission (global scope), and the permission engine has no admin bypass — holding the admin role does not imply the permission.
- The frontend currently has no way to know a user's permissions (only a derived `is_admin` flag), so it cannot decide who should see an analytics navigation entry. Users without the permission should not be led to a dead-end page.
- Users who reach the page without the permission must get a clear explanation, not a raw error.

### Expected states
The page must behave sensibly in all cases: data present, no logins in the window, permission denied, and API failure — plus a loading state.

### Where
- `frontend/app/dashboard/` — core dashboard pages exist (resources, llm, admin) but none for analytics.
- `frontend/lib/` — no client wrapper for the analytics read endpoint (`GET /api/v1/analytics/login-activity`, already present in the generated client).
- `frontend/components/AppSidebar.tsx` — no analytics nav entry, and no permission signal to gate one on.
- `sparkth/api/v1/user.py` — `/user/me` exposes `is_admin` only; no effective-permissions information for the frontend.

### What's needed
- A `/dashboard/analytics` page rendering the daily login-activity data from the read API, plus a navigation entry visible only to users who hold `analytics.read`. Minimal presentation is fine for this slice (no new charting dependency required).
- A way for the frontend to know the current user's effective global-scope permissions, so navigation can be gated on the real permission rather than `is_admin`.

zaira-bibi:
## Part of: [#487 — Analytics are not visible anywhere in the product UI](https://github.com/edly-io/sparkth/issues/487)

> Backend prerequisite. The frontend that consumes this (the analytics dashboard, stacked PR #539) closes #487. Part of the [analytics epic #434](https://github.com/edly-io/sparkth/issues/434).

## What
Add a server-side single-permission **check** endpoint so a client (e.g. the web UI) can ask whether the authenticated user holds a specific permission and gate UI precisely — there is no admin bypass, so `is_admin` alone is the wrong signal.

## Changes
- feat(permissions): `GET /api/v1/permissions/can?permission=<name>&scope=<kind>&scope_object_id=<id?>` → `{"allowed": bool}`, auth-only, backed directly by the engine's `can()`; unknown permission/scope names map to `422` via the exception registry
- refactor(permissions): drop the initial approach (a `permissions` list on `/user/me` + a `get_user_permissions` helper) in favour of the check endpoint, per review feedback — keeps the authorization decision on the server (one source of truth, incl. the scope cascade), extends to object-scoped checks, and discloses only the answer asked for

## How to Test
1. Backend suites — all green:
   `uv run pytest tests/api/v1/test_permission_check.py tests/api/v1/test_user.py tests/permissions/ -v`
2. Gates: `make mypy` and `make lint.backend` → clean.
3. Manual (app running): grant a user a role containing `analytics.read` at the global scope
   (`make cli -- roles assign-role <user> <role>`), authenticate, then
   `GET /api/v1/permissions/can?permission=analytics.read` → `{"allowed": true}`;
   a user without the grant → `{"allowed": false}`.
4. An unknown permission (`?permission=does.not.exist`) or scope (`&scope=nope`) → `422`; unauthenticated → `403`.

## Notes
- Supersedes this PR's earlier `/user/me` permissions-list design (reworked per TL feedback). No migration; no new dependencies.
- The check endpoint is a **UI-gating convenience, not a security boundary** — every endpoint still enforces its own permission and returns 403.
- The frontend consumer (analytics dashboard) is the stacked PR #539; it will call this endpoint to gate the nav entry.


zaira-bibi:
@claude-review

sparkth/api/v1/permissions/routes/__init__.py
```diff
@@ -23,4 +30,25 @@ async def list_permissions() -> list[str]:
...
+    """Return whether the current user holds ``permission`` at the given scope.
+
+    A self-check any authenticated user may call about their own grants. It is not a security
+    boundary (endpoints enforce their own permissions and return 403); it lets a client gate UI
+    precisely. The permission and scope names are resolved against the registered vocabulary
+    (unknown → 422) and the decision is delegated to the engine's ``can()``.
+    """
+    permission_obj = get_permission(permission)
+    scope_obj = get_permission_scope(scope)
+    allowed = await can(current_user, permission_obj, scope_obj, scope_object_id, session)
```
abdulrafey1: `permission_obj` and `scope_obj` are only used once, we can call the functions inline, no need to store them in variables

regisb (approved):
yup this is good!

PR #534, opened 2026-07-23
fix(core): store document status as native enum

abdulrafey1:
## What
Chat requests 500 on PostgreSQL because the Document model's `status` mapping references a native `documentstatus` enum type that was never created in the database — queries cast to a type that does not exist.

## Changes
- fix(core): convert `documents.status` to a native `documentstatus` enum (migration + model mapping)
- test(core): guard test asserting the mapped column is the native enum with lowercase value labels

## How to Test
1. `make services.up && make migrations` (PostgreSQL dev services)
2. `make backend.up.dev`, open chat, and send a message — no 500
3. `uv run pytest tests/core/test_documents.py` — all pass

## Notes
Adds migration `4aac177b2327`: creates the `documentstatus` type and converts `documents.status` from varchar, PostgreSQL only (no-op on SQLite, where the enum renders as plain text). Enum labels are the lowercase `DocumentStatus` member values so the rows already stored in the column cast cleanly; the downgrade restores varchar and drops the type.

### DIFF
--- a/sparkth/core/documents/models.py
+++ b/sparkth/core/documents/models.py
@@ -2,11 +2,17 @@
 
 from typing import Optional
 
-from sqlalchemy import String
+from sqlalchemy import Column, Enum
 from sqlmodel import Field
 
 from sparkth.core.documents.enums import DocumentStatus
 from sparkth.core.models.base import SoftDeleteModel, TimestampedModel
+
+
+def _document_status_values(enum_cls: type[DocumentStatus]) -> list[str]:
+    """Enum labels for the documentstatus DB type: the lowercase member values,
+    matching the strings stored in documents.status rows."""
+    return [status.value for status in enum_cls]
 
 
 class Document(TimestampedModel, SoftDeleteModel, table=True):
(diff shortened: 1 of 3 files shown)


abdulrafey1:

- #534 ◀
    - #535

zaira-bibi (approved):

sparkth/core/documents/models.py
```diff
@@ -23,5 +24,8 @@ class Document(TimestampedModel, SoftDeleteModel, table=True):
     name: str = Field(max_length=500)
     mime_type: Optional[str] = Field(default=None, max_length=255)
 
-    status: DocumentStatus = Field(default=DocumentStatus.QUEUED)
+    # Stored as a plain string: the documents.status column is varchar, and a
+    # native-enum mapping would emit ::documentstatus casts that fail on
+    # PostgreSQL because no such type exists.
+    status: DocumentStatus = Field(default=DocumentStatus.QUEUED, sa_type=String)
```
regisb: I don't understand. Postgresql supports Enum types, right? So why not use an Enum to store values?

abdulrafey1: that works too, will add a new migration

PR #535, opened 2026-07-23
feat(core): make frontend serving opt-in via SERVE_FRONTEND

abdulrafey1:
## What
The backend served any leftover `frontend/out` static export in dev — a weeks-stale UI calling removed API routes — so static frontend serving is now opt-in via the `SERVE_FRONTEND` setting.

## Changes
- feat(core): add `SERVE_FRONTEND` setting (default `false`); mount the static export only when enabled
- feat(core): extract mount logic into `mount_frontend()`; startup fails if serving is enabled but the export directory is missing (`StaticFiles` raises `RuntimeError`)
- chore(docker): enable `SERVE_FRONTEND` in the production image via ENV
- ci: set `SERVE_FRONTEND` in the Playwright job so the e2e suite keeps testing the backend-served frontend on :7727
- docs: document the new variable in `.env` and the configuration reference
- test(core): cover disabled-by-default, enabled-with-export, crash-without-export, and the dev default

## How to Test
1. `make backend.up.dev` with a `frontend/out` directory present — `curl http://localhost:7727/` returns 404 JSON, `/docs` still returns 200
2. Restart with `SERVE_FRONTEND=true` — `curl http://localhost:7727/` serves the static export
3. Restart with `SERVE_FRONTEND=true` and no `frontend/out` — startup aborts with `RuntimeError`
4. `uv run pytest tests/core/test_assemble_app.py` — all pass

## Notes
New env var: `SERVE_FRONTEND` (default `false`). The production image sets it to `true` via Dockerfile ENV, so no infra change is needed. Behavior changes: anyone relying on the dev backend serving the static export must use the Next dev server on :3000 or set `SERVE_FRONTEND=true` locally, and enabling serving without a built export now aborts startup instead of coming up API-only (`make frontend.build` first).

### DIFF
--- a/sparkth/main.py
+++ b/sparkth/main.py
@@ -57,17 +57,12 @@ mcp_app = mcp.http_app(path="/mcp")
 def mount_frontend(application: FastAPI) -> None:
     """Mount the static frontend export at "/" when serving is enabled.
 
-    Serving is opt-in via SERVE_FRONTEND (the production image enables it) and
-    additionally requires FRONTEND_DIR to exist.
+    Serving is opt-in via SERVE_FRONTEND (the production image enables it).
+    Startup fails (StaticFiles raises RuntimeError) if FRONTEND_DIR is missing:
+    the app must never come up half-served when asked to serve the frontend.
     """
     frontend_settings = get_settings()
     if not frontend_settings.SERVE_FRONTEND:
-        return
-    if not frontend_settings.FRONTEND_DIR.exists():
-        logger.warning(
-            "SERVE_FRONTEND is enabled but frontend directory %s does not exist; frontend not mounted",
-            frontend_settings.FRONTEND_DIR,
-        )
         return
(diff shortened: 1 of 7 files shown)


abdulrafey1:

- #534
    - #535 ◀

sparkth/main.py
```diff
@@ -54,6 +54,24 @@ async def plugin_lifespan() -> AsyncIterator[None]:
...
+def mount_frontend(application: FastAPI) -> None:
+    """Mount the static frontend export at "/" when serving is enabled.
+
+    Serving is opt-in via SERVE_FRONTEND (the production image enables it) and
+    additionally requires FRONTEND_DIR to exist.
+    """
+    frontend_settings = get_settings()
+    if not frontend_settings.SERVE_FRONTEND:
+        return
+    if not frontend_settings.FRONTEND_DIR.exists():
```
regisb: I think the app should crash if the directory does not exist. We should not attempt to serve the app without serving the frontend. So this check is not necessary.

zaira-bibi (approved):

PR #538, opened 2026-07-23
fix(plugins): read .env.local overrides in plugin settings

abdulrafey1:
## What
Plugin settings classes read only `.env`, so values overridden in the git-ignored `.env.local` — the documented home for sensitive credentials and local overrides — were silently ignored (e.g. a dev `GOOGLE_DRIVE_REDIRECT_URI` override never reached the OAuth flow, breaking Google Drive connects in local development).

## Changes
- fix(plugins): read `.env.local` after `.env` in `GoogleDriveSettings`, `ChatSettings`, and `SlackSettings`, matching core `Settings` and `RAGSettings`
- refactor(core): define the shared env-file list once as `ENV_FILES` in core config, exposed to plugins via `sparkth.lib.settings`, so settings classes cannot drift from the documented order

## How to Test
1. Put `GOOGLE_DRIVE_REDIRECT_URI=http://localhost:3000/api/v1/google-drive/oauth/callback` in `.env.local` (different from the `.env` value)
2. `make backend.up.dev`, then `GET /api/v1/google-drive/oauth/authorize` — the authorization URL carries the `.env.local` redirect URI

## Notes
No new env vars and no migration. Behavior change: plugin-level values set in `.env.local` now take effect (previously ignored) — this is the precedence `.env` and the configuration guide already document.


abdulrafey1:

- #534
    - #535
        - #538 ◀

sparkth/plugins/chat/config.py
```diff
@@ -9,7 +9,7 @@
 class ChatSettings(BaseSettings):
     model_config = SettingsConfigDict(
         env_prefix="CHAT_",
-        env_file=".env",
+        env_file=(".env", ".env.local"),
```
regisb: Could we use this opportunity to store this tuple in a constant that is then imported by plugins?

tests/core/test_config.py
```diff
@@ -0,0 +1,33 @@
+"""Tests for environment-file conventions shared by all Settings classes."""
```
regisb: These tests are superfluous IMHO.

regisb (approved):
Minor comments, otherwise this is good to go.

PR #539, opened 2026-07-24
feat(frontend): add analytics dashboard at /dashboard/analytics

### LINKED ISSUE #487: Analytics are not visible anywhere in the product UI

zaira-bibi:
## Closes: [#487 — Analytics are not visible anywhere in the product UI](https://github.com/edly-io/sparkth/issues/487)

> 📚 **Stacked on [#533](https://github.com/edly-io/sparkth/pull/533)** (`feat(permissions): add GET /permissions/can check endpoint`). Review/merge #533 first; this PR's diff is against #533's branch. Part of the [analytics epic #434](https://github.com/edly-io/sparkth/issues/434).

## What
Makes login activity visible in-app at `/dashboard/analytics` — a hand-rolled SVG bar chart plus two stat tiles, sourced from the existing `GET /api/v1/analytics/login-activity` read API — closing Milestone 1's last (UI) leg.

## Changes
- feat(frontend): `lib/analytics` — typed `fetchLoginActivity(token, { days })` client mirroring `lib/llm`
- feat(frontend): `components/ui/BarChart` — zero-dependency hand-rolled SVG bar chart primitive (shadcn recipe: themed, dark-mode-native, `role="img"`)
- feat(frontend): `components/ui/StatCard` — stat tile composed on `Card`
- feat(frontend): `app/dashboard/analytics` — client page (fetch-once, client-side zero-fill of the sparse response into a continuous 30-day series, two derived stats from that same series, four UI states: loading / 403 / error / ready-or-empty) + route wrapper
- feat(frontend): `lib/permissions` — `checkPermission(token, permission)` client calling #533's `GET /api/v1/permissions/can`
- feat(frontend): sidebar "Analytics" nav entry gated by that check (a `canViewAnalytics` boolean the dashboard layout fetches once and fails closed on error), threaded through the layout + mobile sidebar
- test(frontend): vitest coverage for both clients, the two primitives, the page's four states + derivation helpers, and the sidebar gating

## How to Test
1. `make frontend.install.dev` (if needed), then `make test.frontend.vitest` → all green.
2. `cd frontend && bun run typecheck` and `make lint.frontend` → clean.
3. Manual (`make backend.up.dev` + `make frontend.up.dev`): grant a user a role with `analytics.read` at the global scope (`make cli -- roles assign-role <user> <role>`), log in → the **Analytics** nav entry appears and `/dashboard/analytics` renders the bar chart + two stat tiles.
4. A user **without** `analytics.read` → no nav entry; direct navigation redirects to the dashboard.
5. Sparse-data sanity: with some days missing, the chart still renders a continuous 30-day series and "Total logins" equals the sum of the visible bars.

## Notes
- **Stacked PR** — merge after #533 (`gh stack sync --prune` once #533 lands to rebase onto `main`).
- Nav gating uses #533's `GET /permissions/can` (not a permissions list on `/user/me`) — a UI-gating convenience, not a security boundary; endpoints still enforce their own permissions and return 403.
- No migration; no new dependencies (the chart is hand-rolled SVG per issue #487's guidance).

### DIFF
--- a/frontend/app/dashboard/analytics/AnalyticsPage.test.tsx
+++ b/frontend/app/dashboard/analytics/AnalyticsPage.test.tsx
@@ -1,0 +1,226 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { act, render, screen, waitFor, within } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { redirect } from "next/navigation";
+
+import { ApiRequestError } from "@/lib/api";
+
+// Auth: a mutable token holder so a test can change identity mid-fetch (stale-response guard).
+const auth = vi.hoisted(() => ({ token: "test-token" }));
+vi.mock("@/lib/auth-context", () => ({
+  useAuth: () => ({ token: auth.token }),
+}));
+
+// Analytics data access: keep the real types, stub the fetch.
+vi.mock("@/lib/analytics", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@/lib/analytics")>();
+  return { ...actual, fetchLoginActivity: vi.fn() };
+});
+
+// Denied access redirects; stub redirect so we can assert it (and keep render from navigating).
+vi.mock("next/navigation", () => ({ redirect: vi.fn() }));
+
+import { fetchLoginActivity, type LoginActivityPoint } from "@/lib/analytics";
+import AnalyticsPage from "./AnalyticsPage";
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  auth.token = "test-token";
+});
+
+describe("AnalyticsPage states", () => {
+  it("shows a loading indicator, then the chart and stats", async () => {
+    // Two points with distinct counts so the "Total logins" (5) and "Busiest
+    // day" (3) stat tiles render different text — a single point would make
+    // both tiles show the same value and collide on the `findByText("5")` query.
+    vi.mocked(fetchLoginActivity).mockResolvedValue([
+      { day: "2026-07-23", login_count: 3 },
+      { day: "2026-07-22", login_count: 2 },
+    ]);
+
+    const { container } = render(<AnalyticsPage />);
+
+    // loading first
+    expect(screen.getByText(/loading/i)).toBeInTheDocument();
+
+    // then content: total logins stat + a chart
+    expect(await screen.findByText("5")).toBeInTheDocument();
+    await waitFor(() => expect(container.querySelector("svg[role='img']")).toBeInTheDocument());
+  });
+
+  it("shows an empty state when there are no logins", async () => {
+    vi.mocked(fetchLoginActivity).mockResolvedValue([]);
+
+    render(<AnalyticsPage />);
+
+    expect(await screen.findByText(/no logins in the last 30 days/i)).toBeInTheDocument();
+  });
+
+  it("labels the window as UTC in the subheading so dates are not misread across timezones", () => {
+    vi.mocked(fetchLoginActivity).mockResolvedValue([]);
+
+    render(<AnalyticsPage />);
+
+    // Dates are UTC-bucketed server-side (see lib/analytics reads); the subheading must say
+    // so, or a user in another timezone reads the last bar as "today" and it disagrees.
+    expect(screen.getByText(/over the last 30 days \(UTC\)/i)).toBeInTheDocument();
+  });
+
+  it("redirects to the dashboard on 403 without disclosing permission details", async () => {
+    vi.mocked(fetchLoginActivity).mockRejectedValue(
+      new ApiRequestError({ message: "Permission denied", fieldErrors: {} }, 403),
+    );
+
+    render(<AnalyticsPage />);
+
+    await waitFor(() => expect(redirect).toHaveBeenCalledWith("/dashboard"));
+    // No permission name is disclosed anywhere.
+    expect(screen.queryByText(/analytics\.read/i)).not.toBeInTheDocument();
+  });
+
+  it("shows a generic error on other failures", async () => {
+    vi.mocked(fetchLoginActivity).mockRejectedValue(
+      new ApiRequestError({ message: "boom", fieldErrors: {} }, 500),
+    );
+
+    render(<AnalyticsPage />);
+
+    expect(
+      await screen.findByText(/couldn't load|could not load|failed to load/i),
+    ).toBeInTheDocument();
+  });
+
+  it("re-fetches when the error state's retry button is pressed", async () => {
+    vi.mocked(fetchLoginActivity)
+      .mockRejectedValueOnce(new ApiRequestError({ message: "boom", fieldErrors: {} }, 500))
+      .mockResolvedValueOnce([
+        { day: "2026-07-23", login_count: 3 },
+        { day: "2026-07-22", login_count: 2 },
+      ]);
+
+    render(<AnalyticsPage />);
+
+    await userEvent.click(await screen.findByRole("button", { name: /try again/i }));
+
+    expect(await screen.findByText("Total logins")).toBeInTheDocument();
+    expect(fetchLoginActivity).toHaveBeenCalledTimes(2);
+    expect(screen.queryByRole("button", { name: /try again/i })).not.toBeInTheDocument();
+  });
+});
+
+describe("AnalyticsPage stat/chart window consistency", () => {
+  beforeEach(() => {
+    // shouldAdvanceTime keeps waitFor/findByText's internal polling ticking on the real
+    // clock while Date/new Date() stays pinned to the value set via setSystemTime.
+    vi.useFakeTimers({ shouldAdvanceTime: true });
+    vi.setSystemTime(new Date("2026-07-24T12:00:00Z"));
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("derives the stat tiles from the same 30-day window as the chart, not the raw response", async () => {
+    // The read API's window floor is inclusive (`>= now - days`), so it can return a login
+    // dated exactly `today - DAYS` (2026-06-24, 30 days before 2026-07-24). But the 30-slot
+    // chart built by buildDailySeries only spans [today-29 .. today] (oldest slot 2026-06-25),
+    // so that boundary day has no bar. Only 2026-07-23 is inside the chart's window.
+    vi.mocked(fetchLoginActivity).mockResolvedValue([
+      { day: "2026-06-24", login_count: 5 },
+      { day: "2026-07-23", login_count: 2 },
+    ]);
+
+    render(<AnalyticsPage />);
+
+    const totalCard = (await screen.findByText("Total logins")).parentElement as HTMLElement;
+    // Windowed total is 2 (only 2026-07-23), not 7 (5 + 2 from the raw response).
+    expect(within(totalCard).getByText("2")).toBeInTheDocument();
+
+    const busiestCard = screen.getByText("Busiest day").parentElement as HTMLElement;
+    // Busiest in-window day is 2026-07-23 (count 2), not the out-of-window 2026-06-24 (count 5).
+    expect(within(busiestCard).getByText("2")).toBeInTheDocument();
+    expect(within(busiestCard).getByText("2026-07-23")).toBeInTheDocument();
+
+    expect(screen.queryByText("2026-06-24")).not.toBeInTheDocument();
+    expect(screen.queryByText("7")).not.toBeInTheDocument();
+  });
+});
+
+describe("AnalyticsPage empty-window composition", () => {
+  it("shows the empty state when the API returns rows but all fall outside the chart window", async () => {
+    vi.mocked(fetchLoginActivity).mockResolvedValue([{ day: "2000-01-01", login_count: 5 }]);
+
+    render(<AnalyticsPage />);
+
+    expect(await screen.findByText(/no logins in the last 30 days/i)).toBeInTheDocument();
+    expect(screen.queryByText("Total logins")).not.toBeInTheDocument();
+    expect(screen.queryByText("Busiest day")).not.toBeInTheDocument();
+  });
+});
+
+describe("AnalyticsPage stale-response guard", () => {
+  function deferred<T>() {
+    let resolve!: (value: T) => void;
+    const promise = new Promise<T>((r) => {
+      resolve = r;
+    });
+    return { promise, resolve };
+  }
+
+  it("ignores an earlier request that resolves after a newer one when the token changes mid-fetch", async () => {
+    const today = new Date().toISOString().slice(0, 10); // always in-window
+    const first = deferred<LoginActivityPoint[]>();
+    const second = deferred<LoginActivityPoint[]>();
+    vi.mocked(fetchLoginActivity)
+      .mockReturnValueOnce(first.promise)
+      .mockReturnValueOnce(second.promise);
+
+    auth.token = "token-A";
+    const { rerender } = render(<AnalyticsPage />); // fetches with token-A (first, pending)
+
+    auth.token = "token-B";
+    rerender(<AnalyticsPage />); // token changed → refetch with token-B (second, pending)
+
+    // The newer request (B) resolves first and is shown…
+    await act(async () => {
+      second.resolve([{ day: today, login_count: 2 }]);
+    });
+    const totalCard = (await screen.findByText("Total logins")).parentElement as HTMLElement;
+    expect(within(totalCard).getByText("2")).toBeInTheDocument();
+
+    // …then the stale request (A) resolves last and must NOT overwrite it.
+    await act(async () => {
+      first.resolve([{ day: today, login_count: 9 }]);
+    });
+    expect(within(totalCard).getByText("2")).toBeInTheDocument();
+    expect(screen.queryByText("9")).not.toBeInTheDocument();
+  });
+});
+
+describe("AnalyticsPage window stability", () => {
+  beforeEach(() => {
+    vi.useFakeTimers({ shouldAdvanceTime: true });
+    vi.setSystemTime(new Date("2026-07-24T12:00:00Z"));
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("keeps the window pinned to the fetched data across an unrelated re-render", async () => {
+    vi.mocked(fetchLoginActivity).mockResolvedValue([{ day: "2026-06-25", login_count: 5 }]);
+
+    const { rerender } = render(<AnalyticsPage />);
+
+    const totalCard = (await screen.findByText("Total logins")).parentElement as HTMLElement;
+    expect(within(totalCard).getByText("5")).toBeInTheDocument();
+
+    // A day passes, then something unrelated re-renders the page (same `points`).
+    vi.setSystemTime(new Date("2026-07-25T12:00:00Z"));
+    rerender(<AnalyticsPage />);
+
+    expect(within(totalCard).getByText("5")).toBeInTheDocument();
+    const busiestCard = screen.getByText("Busiest day").parentElement as HTMLElement;
+    expect(within(busiestCard).getByText("2026-06-25")).toBeInTheDocument();
+  });
+});
--- a/frontend/app/dashboard/analytics/AnalyticsPage.tsx
+++ b/frontend/app/dashboard/analytics/AnalyticsPage.tsx
@@ -1,0 +1,133 @@
+"use client";
+
+import { useCallback, useEffect, useMemo, useState } from "react";
+import { redirect } from "next/navigation";
+import { ChartColumn, RefreshCw } from "lucide-react";
+import { useAuth } from "@/lib/auth-context";
+import {
+  buildDailySeries,
+  fetchLoginActivity,
+  LOGIN_ACTIVITY_DAYS,
+  summarize,
+  type LoginActivityPoint,
+} from "@/lib/analytics";
+import { ApiRequestError } from "@/lib/api";
+import { BarChart } from "@/components/ui/BarChart";
+import { StatCard } from "@/components/ui/StatCard";
+import { Alert } from "@/components/ui/Alert";
+import { Button } from "@/components/ui/Button";
+import { Spinner } from "@/components/Spinner";
+
+type State =
+  | { status: "loading" }
+  | { status: "ready"; points: LoginActivityPoint[]; fetchedAt: Date }
+  | { status: "forbidden" }
+  | { status: "error"; message: string };
+
+export default function AnalyticsPage() {
+  const { token } = useAuth();
+  const [state, setState] = useState<State>({ status: "loading" });
+
+  const [reloadKey, setReloadKey] = useState(0);
+  useEffect(() => {
+    if (!token) return;
+    let active = true;
+    setState({ status: "loading" });
+    fetchLoginActivity(token, { days: LOGIN_ACTIVITY_DAYS })
+      .then((points) => {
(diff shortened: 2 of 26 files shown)


zaira-bibi:
@claude-review

frontend/app/dashboard/layout.tsx
```diff
@@ -84,6 +91,7 @@ export default function DashboardLayout({ children }: { children: React.ReactNod
     email: user?.email,
     plan: "Free Plan",
     is_admin: user?.is_admin,
+    permissions: user?.permissions,
```
abdulrafey1: do we still need this?

frontend/components/AppSidebar.tsx
```diff
@@ -230,6 +232,30 @@ export default function AppSidebar({
               </Link>
             </div>
 
+            {user?.permissions?.includes("analytics.read") && (
```
abdulrafey1: Should this check not be moved to the new `permissions/can` endpoint

zaira-bibi:
@claude-review

regisb (approved):
I'm definitely not the right person to review this frontend PR but I approve of this.

frontend/app/dashboard/layout.tsx
```diff
@@ -65,6 +74,24 @@ function DashboardContent({
 export default function DashboardLayout({ children }: { children: React.ReactNode }) {
   const { token, user, isAuthenticated, loading, logout } = useAuth();
 
+  const [canViewAnalytics, setCanViewAnalytics] = useState(false);
+
+  useEffect(() => {
+    if (!token) return;
+    let active = true;
+    checkPermission(token, "analytics.read")
```
hamza-56, line 82: This works fine for one permission, but it's one round trip plus one useState/useEffect per gated nav item, and it'll grow every time we gate something new. Since /user/me no longer returns the permission list, maybe a small permissions context that fetches and caches these (or let `/can` accept multiple permissions)? Fine to punt, just flagging before the pattern spreads.

frontend/app/dashboard/layout.tsx
```diff
@@ -65,6 +74,24 @@ function DashboardContent({
...
+      .then((allowed) => {
+        if (active) setCanViewAnalytics(allowed);
+      })
+      .catch((error) => {
+        // Fail closed (hide the entry) and log — this is UI gating, not enforcement.
+        console.error("Failed to check analytics permission:", error);
```
hamza-56, line 88: Tiny edge case: "fail closed" only holds because the initial state is false. If the token changes (re-login as someone else) and this re-check errors, the previous user's `true` sticks around. Worth a `setCanViewAnalytics(false)` here (guarded by `active`).

frontend/app/dashboard/layout.tsx
```diff
@@ -84,6 +111,7 @@ export default function DashboardLayout({ children }: { children: React.ReactNod
     email: user?.email,
     plan: "Free Plan",
     is_admin: user?.is_admin,
+    canViewAnalytics,
```
hamza-56, line 114: A permission flag living in the user display object is a bit of a smell, and it's also why the casing looks off: name/email/is_admin are snake_case straight from the API, while `canViewAnalytics` is our own frontend invention. I wouldn't rename it, I'd pull it out of `user` entirely (permissions context or a separate prop) so this object stays API-shaped and AppSidebar/MobileSidebar don't all need the extra field.

frontend/app/dashboard/analytics/AnalyticsPage.tsx
```diff
@@ -0,0 +1,139 @@
...
+        {state.status === "ready" && state.points.length > 0 && (
+          <AnalyticsContent points={state.points} />
+        )}
+      </div>
+    </div>
+  );
+}
+
+function AnalyticsContent({ points }: { points: LoginActivityPoint[] }) {
+  const series = buildDailySeries(points, DAYS, new Date());
```
hamza-56: `new Date()` in the render body makes this impure, and the series + summary get rebuilt on every re-render. Wrapping it in `useMemo(..., [points])` fixes both, and pins the window to when the data arrived instead of drifting on unrelated re-renders.

frontend/app/dashboard/analytics/AnalyticsPage.tsx
```diff
@@ -0,0 +1,139 @@
...
+        {state.status === "loading" && (
+          <div className="flex items-center justify-center py-24">
+            <div className="text-center">
+              <Spinner className="mx-auto mb-4" />
+              <p className="text-muted-foreground">Loading analytics…</p>
+            </div>
+          </div>
+        )}
+
+        {state.status === "error" && <Alert severity="error">{state.message}</Alert>}
```
hamza-56: nit: the copy says "Please try again" but there's no retry. `load` is already memoized, so a small "Try again" button here would be cheap.

frontend/lib/analytics/client.ts
```diff
@@ -0,0 +1,17 @@
...
+export async function fetchLoginActivity(
+  token: string,
+  { days = 30 }: { days?: number } = {},
+): Promise<LoginActivityPoint[]> {
+  try {
+    const { data } = await api.GET("/api/v1/analytics/login-activity", {
+      params: { query: { days } },
+      headers: bearer(token),
+    });
+    return data as LoginActivityPoint[];
```
hamza-56: This try/catch + `data as T` is the same shape as the private `call<T>` helper in `lib/api/endpoints.ts`, and permissions/client.ts repeats it a third time. If we export `call` from lib/api, this whole function becomes a one-liner.

frontend/lib/permissions/client.ts
```diff
@@ -0,0 +1,19 @@
...
+  token: string,
+  permission: string,
+  { scope, scopeObjectId }: { scope?: string; scopeObjectId?: string } = {},
+): Promise<boolean> {
+  try {
+    const { data } = await api.GET("/api/v1/permissions/can", {
+      params: { query: { permission, scope, scope_object_id: scopeObjectId } },
+      headers: bearer(token),
+    });
+    return data?.allowed ?? false;
```
hamza-56: Same `call<T>` boilerplate as analytics/client.ts (see comment there). Also, errorMiddleware throws on any non-2xx, so `data` can't really be undefined here. The `?? false` reads like we might quietly answer "no" on a broken response; I'd just do `data.allowed` and let it fail loud.

frontend/components/AppSidebar.tsx
```diff
@@ -230,6 +232,30 @@ export default function AppSidebar({
               </Link>
             </div>
 
+            {user?.canViewAnalytics && (
```
hamza-56, line 235: This is now the fifth copy of this exact Link block (Resources, My Plugins, AI Keys, Analytics, Admin), only href/icon/label differ. Not this PR's fault the pattern exists, but rather than extending it, could we extract a `SidebarNavLink` like we already did with `PluginNavItem`? New entries become one line.

hamza-56 (commented):
Left a few inline comments, mostly structure/DRY stuff on the frontend. Backend side looks clean to me.

frontend/components/AppSidebar.tsx
zaira-bibi: Handled in #558

frontend/app/dashboard/layout.tsx
zaira-bibi: Handled in #558.

frontend/app/dashboard/layout.tsx
zaira-bibi: Handled in #558.

frontend/app/dashboard/analytics/AnalyticsPage.test.tsx
```diff
@@ -0,0 +1,202 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
```
abdulrafey1, line 1: Should we not move these tests to the `lib/tests/` directory?
We have some of the tests in `lib/tests/` and some here, this will be troublesome as we scale

frontend/app/dashboard/layout.tsx
```diff
@@ -32,7 +34,14 @@ function DashboardContent({
   logout,
 }: {
   children: React.ReactNode;
-  user: { name?: string; email?: string; avatar?: string; plan?: string; is_admin?: boolean };
+  user: {
```
abdulrafey1, line 37: Lets not add `canViewAnalytics` in the `user` object. We should keep the user's identity and the user's authorization information separate.

One more thing I noticed is that this `user` object is being used in 3 places, `frontend/app/dashboard/layout.tsx`, `frontend/components/AppSidebar.tsx`, and `frontend/components/MobileSidebar.tsx`, adding one field meant editing three files.

Extracting the shape to one exported type (e.g. `DashboardUser`) removes the triplication, and passing permissions as their own prop or a small context keeps identity and authorization separate. Both get cheaper the moment a second nav entry needs gating.

frontend/app/dashboard/analytics/AnalyticsPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { Alert } from "@/components/ui/Alert";
+import { Button } from "@/components/ui/Button";
+import { Spinner } from "@/components/Spinner";
+
+const DAYS = 30;
+
+// The API omits zero-login days (newest-first, sparse). Build a continuous
+// oldest→newest series of `days` UTC days, filling missing days with 0. `now`
+// is injectable so the derivation is unit-testable without faking the clock.
+export function buildDailySeries(
```
abdulrafey1: These two are pure functions over the API response, exported from a component module so the tests can reach them. Zero-filling a sparse response is data-layer work, and this PR just created the right home for it — `lib/analytics/`.

Moving them to something like `lib/analytics/series.ts` keeps `AnalyticsPage` about rendering, lets the tests import them directly instead of through a component, and puts them next to `fetchLoginActivity`, whose shape they depend on.

frontend/components/AppSidebar.tsx
```diff
@@ -230,6 +232,30 @@ export default function AppSidebar({
...
+              <div>
```
abdulrafey1, line 236: The only change in this file is the changes for the Nav item menu.

Lets create a `<NavItem>` component, it will de-bloat the file and make managing the components easier. This way the next time we have to edit the nav item, the change will only be limited to its file and would be easier to read/comprehend

frontend/app/dashboard/analytics/AnalyticsPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { useAuth } from "@/lib/auth-context";
+import { fetchLoginActivity, type LoginActivityPoint } from "@/lib/analytics";
+import { ApiRequestError } from "@/lib/api";
+import { BarChart, type BarChartDatum } from "@/components/ui/BarChart";
+import { StatCard } from "@/components/ui/StatCard";
+import { Alert } from "@/components/ui/Alert";
+import { Button } from "@/components/ui/Button";
+import { Spinner } from "@/components/Spinner";
+
+const DAYS = 30;
```
abdulrafey1: This constant value shows up here and in `frontend/lib/analytics/client.ts` line 6, lets define a central place for defining constants and we can use the same constant everywhere.

Having them separate will put them at risk of drifting and will be difficult to manage.

frontend/app/dashboard/layout.tsx
```diff
@@ -65,6 +74,24 @@ function DashboardContent({
 export default function DashboardLayout({ children }: { children: React.ReactNode }) {
   const { token, user, isAuthenticated, loading, logout } = useAuth();
 
+  const [canViewAnalytics, setCanViewAnalytics] = useState(false);
+
+  useEffect(() => {
```
abdulrafey1, line 79: lets set `setCanViewAnalytics(false);` in the start of this effect, without it when a second user logs in on the same device, it will still have the value set for the first user until the permission is fetched from the backend.

Resetting at the top of the effect fixes it, and makes "unknown" distinguishable from "denied" if you'd rather not flash the entry at all:

```
useEffect(() => {
  setCanViewAnalytics(false);   // don't carry the previous token's answer over
  if (!token) return;
  // ...
}, [token]);
```

frontend/app/dashboard/layout.tsx
```diff
@@ -65,6 +74,24 @@ function DashboardContent({
...
+    if (!token) return;
```
abdulrafey1, line 80: The early return when !token happens before any state write, so clearing the token on logout leaves canViewAnalytics stuck at its last value.

The `setCanViewAnalytics(false)` reset suggested above covers this case too.

frontend/app/dashboard/analytics/AnalyticsPage.tsx
abdulrafey1: The chart window and the API window are off by one, and it produces a user-visible wrong state rather than just a dropped row.

The read API floors inclusively (`>= now::date - days`, see `core/analytics/reads.py`), so `today − 30` is a row it will return. This builds slots for `[today−29 … today]`, so that row has nowhere to go and is dropped from both the chart and the tiles.

The composition is the real problem: the empty state at line 115 keys off raw `state.points.length`, not the derived series. If every returned row is out of window, `points.length > 0`, the empty state is skipped, and the page renders from all-zero data. With one point `{day: "2026-06-24", login_count: 5} `and the clock at 2026-07-24 I get:

```
Total logins  = 0
Busiest day   = 0  ·  2026-06-25   <- invented; that day had no logins
empty state shown? = false
30 bars, all zero-height           <- visually blank chart
```

Two things worth fixing independently:

**In this PR**: branch the empty state on the derived series (`total === 0`), not on `state.points.length`, so "the API sent rows but none of them count" still renders the honest empty state. This is correct regardless of how the window question is settled.


**The root cause** is the window mismatch itself, and the clean fix is on the read side — flooring at `now - (days - 1)` so the API returns exactly `days` calendar days. Widening the chart to days + 1 slots would also stop dropping the row, but then a heading that says "last 30 days" sits above 31 bars, so I'd avoid that one.

frontend/app/dashboard/analytics/AnalyticsPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+        {state.status === "ready" && state.points.length > 0 && (
+          <AnalyticsContent points={state.points} />
+        )}
+      </div>
+    </div>
+  );
+}
+
+function AnalyticsContent({ points }: { points: LoginActivityPoint[] }) {
+  const { series, total, busiest } = useMemo(() => {
```
abdulrafey1, line 108: `new Date()` inside `useMemo` makes the memo impure — the window it produces depends on when the memo happens to run. The "window stability" test treats the memo's cache as a guarantee, but React explicitly reserves the right to discard memoised values, so this is relying on current implementation behaviour rather than a contract.

Capturing the timestamp alongside the data makes it real state instead of a cache artifact:

```setState({ status: "ready", points, fetchedAt: new Date() });```

…then pass `state.fetchedAt` into `buildDailySeries`. Same behaviour, no reliance on memo identity, and it also makes "the window is the moment we fetched" explicit.

File: frontend/app/dashboard/analytics/AnalyticsPage.tsx [hunk 19]
```diff
@@ -0,0 +1,147 @@
...
+  | { status: "loading" }
+  | { status: "ready"; points: LoginActivityPoint[] }
+  | { status: "forbidden" }
+  | { status: "error"; message: string };
+
+export default function AnalyticsPage() {
+  const { token } = useAuth();
+  const [state, setState] = useState<State>({ status: "loading" });
+
+  const load = useCallback(async () => {
```
abdulrafey1: No cancellation here, so if token changes mid-fetch the earlier response can resolve last and overwrite the newer one. The permission effect you added in dashboard/layout.tsx guards exactly this with an active flag — worth being consistent, either with the same flag or an AbortController.

frontend/app/dashboard/analytics/AnalyticsPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+  points: LoginActivityPoint[],
+  days: number,
+  now: Date,
+): BarChartDatum[] {
+  const counts = new Map(points.map((p) => [p.day, p.login_count]));
+  const series: BarChartDatum[] = [];
+  for (let i = days - 1; i >= 0; i--) {
+    const d = new Date(now);
+    d.setUTCDate(d.getUTCDate() - i);
+    const key = d.toISOString().slice(0, 10); // YYYY-MM-DD (UTC), matches API `day`
```
abdulrafey1: These labels are UTC dates (correctly — the read path buckets in UTC), but they render as bare YYYY-MM-DD with nothing saying so. A user at UTC+12 or UTC−8 will read the last bar as "today" and see it disagree with their own day.

Cheapest fix is qualifying it in the copy — "Daily login activity over the last 30 days (UTC)" in the subheading. Converting to local time would be wrong here, since the buckets themselves are UTC-aligned and can't be re-bucketed client-side.

frontend/app/dashboard/analytics/AnalyticsPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+        setState({ status: "error", message: "We couldn't load analytics. Please try again." });
+      }
+    }
+  }, [token]);
+
+  useEffect(() => {
+    load();
+  }, [load]);
+
+  if (state.status === "forbidden") {
```
abdulrafey1, line 55: This doesn't match the PR description, which says a user without analytics.read "shows the in-page 403 message" — the code redirects to /dashboard with no explanation, and the forbidden variant of the State union renders nothing because redirect() throws first.

Either is a reasonable product call, but they should agree. If the redirect is intended, the description needs updating and the unreachable forbidden render branch is dead weight; if the message is intended, this wants a short "You don't have access to analytics" panel instead. A silent bounce is the one option that leaves the user guessing.

frontend/components/ui/BarChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+        {/* max-value gridline */}
+        <line
+          x1={PAD_X}
+          y1={PAD_TOP}
+          x2={VIEW_W - PAD_X}
+          y2={PAD_TOP}
+          className="stroke-border"
+          strokeWidth={1}
+        />
+        {data.map((d, i) => {
```
abdulrafey1, line 60: A zero value renders a zero-height <rect>, i.e. nothing — so a zero-filled day looks exactly like a missing one. Given buildDailySeries exists specifically to zero-fill, a quiet month renders as a chart full of holes.

A baseline <line> at PAD_TOP + plotH would give the empty days something to sit on and make the chart legible; a 1–2 unit minimum bar height is the other common approach.

frontend/components/ui/BarChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+          const barH = (d.value / max) * plotH;
+          const x = PAD_X + i * slot + (slot - barW) / 2;
+          const y = PAD_TOP + (plotH - barH);
+          return (
+            <rect
+              key={d.label}
```
abdulrafey1: key={d.label} is safe for the analytics caller (dates are unique), but this is a shared components/ui primitive with nothing documenting that constraint — a caller charting ["Mon", "Mon"] would hit duplicate keys. Either key={i} (the order is the identity here, and the list is never reordered) or a note on BarChartDatum that labels must be unique.

frontend/app/dashboard/layout.tsx
abdulrafey1, line 82: This is a bare string literal, and checkPermission takes permission: string, so a typo here fails closed and silently — the nav entry just never appears, with no error anywhere. The backend has a real registry of these (ANALYTICS_READ et al).

A small union or const map in lib/permissions/types.ts (export type PermissionName = "analytics.read" | …) would make typos a typecheck failure. generated.ts doesn't expose the permission vocabulary today, so this is hand-maintained either way — but the union at least centralises it.

same at `lib/permissions/client.ts` line 8

abdulrafey1 (commented):
I went through the PR with the help of claude, found some small issues, nothing major.
I think we all should perform claude-assisted in-depth code reviews for frontend so all of us can get a better idea of what are the common mistakes in the frontend code made by AI and how can we fix them

frontend/app/dashboard/analytics/AnalyticsPage.test.tsx
zaira-bibi: currently `lib/tests/` holds tests for `lib/*` modules only, while the component & page tests are co-located. we can either have per-directory `__tests__/` folders or a single `frontend/__tests__/` to home all the tests (not `frontend/tests/` as it is already being used by e2e tests). wdys?

frontend/app/dashboard/layout.tsx
zaira-bibi: first part handled in #558. will address the second one in that PR.

frontend/components/AppSidebar.tsx
zaira-bibi: Handled in #558

frontend/app/dashboard/layout.tsx
zaira-bibi: Handled in #558

frontend/app/dashboard/analytics/AnalyticsPage.tsx
zaira-bibi: keeping the flooring at `now - (days - 1)` out of this PR as the change touches backend as well.

hamza-56 (approved):

frontend/app/dashboard/analytics/AnalyticsPage.test.tsx
zaira-bibi: delegate this to a separate PR.

abdulrafey1 (approved):

PR #541, opened 2026-07-24
fix(plugins): google drive connect 500s on postgresql

abdulrafey1:
## What
Connecting Google Drive 500s on PostgreSQL: the plugin writes timezone-aware UTC datetimes, but its four datetime columns are `TIMESTAMP WITHOUT TIME ZONE`, and asyncpg rejects aware values on naive columns (`DataError` on the `token_expiry` insert; folder/file sync writes would hit the same on `last_synced_at` / `modified_time`).

## Changes
- fix(plugins): add a `TZDateTime` column type (aware-UTC on write and read for both PostgreSQL and SQLite) and map the four Drive datetime columns (`token_expiry`, `last_synced_at` ×2, `modified_time`) to it
- fix(migrations): convert those columns to `TIMESTAMP WITH TIME ZONE` (`USING ... AT TIME ZONE 'UTC'`; PostgreSQL only, no-op on SQLite)
- test(plugins): guard test asserting the four columns are `TZDateTime`

## How to Test
1. `make services.up && make migrations` (PostgreSQL dev services)
2. `make backend.up.dev`, connect Google Drive from the dashboard — no 500, token stored
3. `uv run pytest sparkth/plugins/googledrive/tests/` — all pass

## Notes
Adds migration `cd057b1ef056` (parent `29952e6a4b1b`). PR #534's migration shares the same parent — whichever merges second must rebase and repoint its `down_revision` to avoid split heads. Existing naive rows are interpreted as UTC, which is what every writer stores.

### DIFF
--- a/sparkth/core/models/base.py
+++ b/sparkth/core/models/base.py
@@ -1,12 +1,36 @@
 from datetime import datetime, timezone
 
-from sqlalchemy import DateTime
+from sqlalchemy import DateTime, Dialect
+from sqlalchemy.types import TypeDecorator
 from sqlmodel import Field, SQLModel
 
 
 def utc_now() -> datetime:
     """Return the current UTC datetime."""
     return datetime.now(timezone.utc)
+
+
+class TZDateTime(TypeDecorator[datetime]):
+    """Column type for datetimes that are timezone-aware UTC on both sides.
+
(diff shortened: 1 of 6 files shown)


regisb (approved):
the differences between sqlite and postgres are getting on my nerves... But I understand there's no other way around the issue.

PR #542, opened 2026-07-24
fix(plugins): failed drive ingestions stay stuck in processing

abdulrafey1:
## What
Retrying a failed Drive ingestion crashes and leaves the document stuck in `processing` forever: `process_folder_rag` loads `DriveFile` rows in a session that closes before per-file processing, and the retry path (document already registered) operates on the detached instance — the first `session.refresh` raises `InvalidRequestError` after `PROCESSING` was already committed. Later folder syncs then skip the document as in-flight, so it never recovers and shows as "INDEXING" in the UI indefinitely.

## Changes
- fix(plugins): re-fetch each `DriveFile` inside the per-file session in `process_folder_rag`, so both the fresh and retry paths operate on a session-bound instance
- fix(plugins): lifecycle backstop in `_process_single_file` — exception types outside the specific catch list now roll back, mark the document `FAILED`, and log with traceback, instead of escaping to the folder-level gather (which only logs) and wedging the document in `processing`
- test(plugins): regression test driving the failed-document retry path through `process_folder_rag` to READY with real sessions
- test(plugins): regression test asserting an unexpected exception type ends the document in `FAILED` with the generic error message

## How to Test
2. Sync a Drive folder containing a file that fails ingestion (e.g. a scanned PDF), then re-sync the folder — the document is retried and reaches a final status instead of sticking in `processing`
3. `uv run pytest sparkth/plugins/googledrive/tests/` — all pass

## Notes
No migration and no new env vars. The specific exception catches stay above the backstop so users keep meaningful error messages; the backstop only covers the unknown tail. Task cancellation and process kills can still leave a document in `processing` — that recovery design (startup sweep vs staleness timeout vs heartbeat) is being discussed separately, and documents already orphaned by earlier crashes are not auto-recovered.

### DIFF
--- a/sparkth/plugins/googledrive/utils.py
+++ b/sparkth/plugins/googledrive/utils.py
@@ -254,10 +254,20 @@ async def _process_with_semaphore(
     access_token: str,
     semaphore: asyncio.Semaphore,
 ) -> None:
-    """Process one file with semaphore-based concurrency control."""
+    """Process one file with semaphore-based concurrency control.
+
+    Re-fetches the DriveFile inside the per-file session: the caller's
+    instances come from a session that is already closed, and a detached
+    instance cannot be refreshed or flushed here.
+    """
     async with semaphore:
         async with session_scope() as file_session:
-            await _process_single_file(drive_file, user_id, access_token, file_session)
+            result = await file_session.exec(select(DriveFile).where(col(DriveFile.id) == drive_file.id))
+            bound_file = result.first()
+            if bound_file is None:
+                logger.warning("Drive file %s disappeared before RAG processing.", drive_file.id)
+                return
+            await _process_single_file(bound_file, user_id, access_token, file_session)
         gc.collect()
         if sys.platform == "linux":
             try:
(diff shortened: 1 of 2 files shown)


sparkth/plugins/googledrive/utils.py
```diff
@@ -243,6 +243,16 @@ async def _process_single_file(
         await session.refresh(drive_file)
         await update_document_status(session, document_id, DocumentStatus.FAILED, "Database integrity error")
         await session.commit()
+    except Exception:
+        # Deliberate lifecycle backstop, not lazy catching: PROCESSING was committed
+        # before ingestion, and later syncs skip in-flight documents, so any unexpected
+        # error escaping here would wedge the document in PROCESSING forever. The
+        # folder-level gather only logs — it cannot fix the status. Roll back the
+        # partial work, record the failure, and swallow (fully handled here).
+        logger.exception("Unexpected error while processing '%s'", log_name)
```
regisb: This should log the actual exception, since we are not re-raising it.

regisb (approved):
Just a minor comment, then this can be merged.

PR #544, opened 2026-07-27
feat(core): add permission groups engine

### LINKED ISSUE #519: Permission system needs groups to grant roles to users in bulk
## Summary

The scoped-RBAC permission system can assign roles to **individual users only**.
`RoleAssignment` is keyed by `user_id`, so granting the same role at the same scope
to N users means N separate assignment rows, each created and maintained
individually. There is no way to grant a role to a *set* of users collectively, and
no entity that represents such a set.

## Where

- `sparkth/core/permissions/models.py` — `RoleAssignment(user_id, role_id, scope, scope_object_id)`; there is no group entity.
- `sparkth/core/permissions/__init__.py` — `can()` / `has_role()` resolve only the user's own assignments at the scope chain.
- `sparkth/cli/roles.py` — the `assign-role` command targets one user at a time.

## Why it matters

- **Bulk grants don't scale.** "Every math instructor gets the teacher role in org 5"
  is one grant conceptually but many rows in practice; membership churn (someone joins
  or leaves the set) means editing individual rows, and the set itself has no home.
- **Organizational structure has no representation.** Departments, cohorts, and class
  sections map naturally to groups, not to repeated per-user grants.
- **No single answer to "who is in this set and what does it grant?"** — the set only
  exists implicitly, spread across identical assignment rows.

## Context

Part of the permission-system EPIC #420 — this is **Phase 3 (Groups)**, described there
as: *"assign a role to a group at a scope, resolve membership transitively. A clean
later extension — no schema rethink."*

abdulrafey1:
## Closes: [Permission system needs groups to grant roles to users in bulk](https://github.com/edly-io/sparkth/issues/519)

Part 1 of 3 stacked PRs (engine → management surface → docs). Does not close the issue on its own.

## What
Adds flat user groups as a second assignee source in the scoped-RBAC engine, so a role granted to a group at a scope is held by every active member through the existing scope cascade.

## Changes
- refactor(core): reshape permission checks around granted-role-id subqueries (behaviour-preserving, existing tests untouched and green before any group code landed)
- feat(core): add `user_group`, `group_membership`, `group_role_assignment` tables with partial-unique active indexes mirroring `uq_role_assignment_active`; `group_membership.source` (`"manual"`) ships now so future rule-driven membership can own its rows without a schema change
- feat(migrations): app-lineage Alembic revision `e69b161ed8ee` for the three tables
- feat(core): `can()` / `has_role()` resolve *own assignments ∪ assignments of groups the user actively belongs to* — `has_role` counting group grants keeps `is_admin` accurate for admin-granted-via-group
- feat(core): group engine functions in `sparkth/core/permissions/groups.py` (CRUD, membership, group-role grants; idempotent race-safe writes, soft-deleted history, `GroupInUse` delete guard) and `GroupNotFound` / `GroupAlreadyExists` / `GroupInUse` exceptions re-exported via `sparkth.lib.permissions.exceptions`

## How to Test
1. `make test.backend.pytest` — full suite green (1492 passed); new coverage in `tests/permissions/test_groups.py` and `tests/permissions/test_group_resolution.py`.
2. `make mypy && make lint.backend` — clean.
3. Engine behaviour: create a group, add a member, insert a `GroupRoleAssignment` at `global` — `can(user, permission, GLOBAL, None, session)` returns True; soft-delete the membership — it returns False.
4. Migration: `make migrations` applies `e69b161ed8ee` (verified up **and** down against PostgreSQL).

## Notes
- **Migration required**: yes — new revision `e69b161ed8ee` (three new tables, no data changes).
- Management surface (REST/CLI/permissions) and docs arrive in the next PRs of the stack.
- Scope posture per the issue's flagged defaults: no per-group delegation scope yet (blocked on #420 Phase 2 object-bearing cascade).


sparkth/core/permissions/exceptions.py
```diff
@@ -38,6 +38,30 @@ def __init__(self, role_id: int) -> None:
...
+
+    def __init__(self, group: str) -> None:
+        super().__init__(f"Group not found: {group}")
+        self.group = group
+
+
+class GroupAlreadyExists(Exception):
+    """Raised when creating or renaming a group to a name that is already taken."""
+
+    def __init__(self, name: str) -> None:
```
regisb, line 52: There's an inconsistency between this "name" attribute and the "group" attribute of the GroupNotFound class. Shouldn't we use the same attribute name?

regisb (approved):
Just a minor comment, otherwise this is good to go :)

hamza-56 (approved):

sparkth/core/permissions/groups.py
```diff
@@ -0,0 +1,271 @@
...
+    RoleNotFound,
+)
+from sparkth.core.permissions.models import Group, GroupMembership, GroupRoleAssignment, Role
+from sparkth.core.permissions.scopes import PermissionScope
+
+
+async def create_group(name: str, description: str | None, session: AsyncSession) -> Group:
+    """Create and return a group. Raises GroupAlreadyExists if the name is already taken."""
+    if (await session.exec(select(Group).where(Group.name == name))).first() is not None:
+        raise GroupAlreadyExists(name)
```
hamza-56, line 34: Check-then-insert race: two concurrent creates with the same name both pass this check, then the unique index makes the loser blow up with a raw IntegrityError (a 500 through the API) instead of GroupAlreadyExists (409). The membership/assignment functions below handle exactly this with a savepoint + re-check, so it'd be easy to do the same here (roles.py has the same gap, but no reason for new code to copy it).

sparkth/core/permissions/groups.py
```diff
@@ -0,0 +1,271 @@
...
+            select(GroupRoleAssignment.id)
+            .where(GroupRoleAssignment.group_id == group_id, GroupRoleAssignment.is_deleted == False)
+            .limit(1)
+        )
+    ).first()
+    if active is not None:
+        raise GroupInUse(group_id)
+    # The group_id foreign keys have no ON DELETE CASCADE, so remove dependents (membership
+    # rows and historical soft-deleted assignments) before deleting the group itself.
+    for membership in (await session.exec(select(GroupMembership).where(GroupMembership.group_id == group_id))).all():
```
hamza-56: This loads every membership and assignment row into memory and issues one DELETE per row. For a big group that's a lot of round trips. A bulk `delete(GroupMembership).where(group_id == ...)` (and same for assignments) does it in two statements.

sparkth/migrations/app/versions/e69b161ed8ee_add_group_tables.py
```diff
@@ -0,0 +1,132 @@
...
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+import sqlmodel
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "e69b161ed8ee"
+down_revision: Union[str, Sequence[str], None] = "b5bf23a5f00e"
```
hamza-56, line 17: The docstring header up top still says `Revises: 4aac177b2327` from before the repoint, so it now contradicts this line. Worth updating.

hamza-56 (commented):
Engine side looks really solid. A few things:

sparkth/core/permissions/exceptions.py
abdulrafey1: Done in c66ac6c — `GroupNotFound` now stores `self.name`, matching `GroupAlreadyExists`.

sparkth/core/permissions/groups.py
abdulrafey1: Fixed in c66ac6c, with race tests for both paths (`update_group`'s rename had the same gap, so it got the same fix). One deviation from the suggestion: instead of a savepoint I translate the IntegrityError at commit time — the CRUD functions own the commit (unlike the membership/assignment helpers, which only flush inside a caller-owned transaction), and a savepoint around a flush of a *persistent* row leaves the renamed object dirty after rollback, poisoning the session on the next autoflush. Catching at commit + rollback gives the loser a clean `GroupAlreadyExists` (409) either way.

sparkth/core/permissions/groups.py
abdulrafey1: Done in c66ac6c — both dependents are removed with bulk DELETE statements now (same pattern as `rag/cleanup.py`).

sparkth/migrations/app/versions/e69b161ed8ee_add_group_tables.py
abdulrafey1: Updated in c66ac6c.

zaira-bibi (approved):

PR #545, opened 2026-07-27
feat(api): add group management surface

### LINKED ISSUE #519: Permission system needs groups to grant roles to users in bulk

abdulrafey1:
## Closes: [Permission system needs groups to grant roles to users in bulk](https://github.com/edly-io/sparkth/issues/519)

Part 2 of 3 stacked PRs — stacked on #544 (engine). Management surface for the group tables the engine PR introduced.

## What
Adds the management surface for permission groups: REST CRUD for group structure, CLI for membership and group-role grants, and the `group.*` permissions gating it — following the split already established for roles (structure via REST, principal/grant edges via CLI).

## Changes
- feat(core): declare `group.create` / `group.read` / `group.update` / `group.delete` permissions and export the group engine surface through `sparkth.lib.permissions`
- feat(migrations): data migration `5c6ccf073bb7` grants the four `group.*` permissions to the seeded `admin` role (no-op when the role isn't seeded)
- feat(api): group CRUD under `/api/v1/permissions/groups` (list / create / get / patch / delete), gated at the global scope; `GroupNotFound → 404`, `GroupAlreadyExists` / `GroupInUse → 409` via the central exception registry
- feat(cli): `groups add-member <user> <group>` and `groups assign-role-to-group <group> <role> [--scope] [--scope-object-id]`
- docs: `group.*` rows in the permissions guide's shipped-permissions table
- chore(frontend): regenerated `frontend/lib/api/generated.ts` from the OpenAPI schema

## How to Test
1. `make test.backend.pytest` — full suite green (1512 passed); new coverage in `tests/api/v1/test_permission_groups.py` and `tests/cli/test_groups.py`.
2. `make migrations` — applies `5c6ccf073bb7` (verified up **and** down against PostgreSQL).
3. As an admin, `POST /api/v1/permissions/groups {"name": "cs-staff"}` → 201; duplicate name → 409 with exact detail body.
4. `make cli -- groups add-member <username> cs-staff` then `make cli -- groups assign-role-to-group cs-staff <role>` — every member now passes `can()` checks for that role's permissions at the granted scope.

## Notes
- **Migration required**: yes — data migration `5c6ccf073bb7` (admin grants only, no schema change).
- Per the issue's flagged defaults: membership/assignment stay CLI-only, and group management gates at GLOBAL scope (a per-group delegation scope waits on #420 Phase 2).

### DIFF
--- a/docs/guides/permissions.md
+++ b/docs/guides/permissions.md
@@ -74,8 +74,8 @@ descendants. Object-bearing multi-level cascade (e.g. a grant at one `org` apply
 | Kind | Names | Notes |
 |---|---|---|
 | **Scopes** | `global`, `whitelist`, `role` | `global` is the objectless root (applies platform-wide). `whitelist` is an objectless singleton nested under `global` for the registration whitelist. `role` is object-bearing (nested under `global`) so role management can be delegated per-role by that role's id. A `global` grant cascades down to both. |
-| **Permissions** | `email.whitelist.{read,create,delete}`; `role.{create,read,update,delete}`; `permission.read` | Gate the registration email-whitelist endpoints, the role-management API, and listing the assignable permission vocabulary respectively. |
-| **Roles** | `admin` | Grants the three `email.whitelist.*`, the four `role.*`, and `permission.read`. A seed migration also assigned it at `global` to every account that was a superuser when the migration ran — a one-time backfill. |
+| **Permissions** | `email.whitelist.{read,create,delete}`; `role.{create,read,update,delete}`; `group.{create,read,update,delete}`; `permission.read` | Gate the registration email-whitelist endpoints, the role-management API, the group-management API, and listing the assignable permission vocabulary respectively. |
+| **Roles** | `admin` | Grants the three `email.whitelist.*`, the four `role.*`, the four `group.*`, and `permission.read`. A seed migration also assigned it at `global` to every account that was a superuser when the migration ran — a one-time backfill. |
 
 Roles and their permission grants are managed at runtime through the role-management REST API
 under `/api/v1/permissions` (gated by `role.*` / `permission.read`). See the REST API docs
--- a/sparkth/api/v1/permissions/routes/groups.py
+++ b/sparkth/api/v1/permissions/routes/groups.py
@@ -1,0 +1,82 @@
+"""Group sub-router (mounted at ``/groups``): CRUD for user groups (issue #519).
+
+Structural definition only — membership and group-role grants are CLI-managed, mirroring
+how user-role assignment is CLI-only. All endpoints authorize at the GLOBAL scope; a
+per-group delegation scope is deferred until object-bearing scope cascade exists
+(EPIC #420 Phase 2). Authored with LLM (Claude) assistance.
+"""
+
+from fastapi import APIRouter, Depends, status
+from sqlmodel.ext.asyncio.session import AsyncSession
+
+from sparkth.api.v1.permissions.schemas import GroupCreate, GroupResponse, GroupUpdate
+from sparkth.core.permissions import groups as group_engine
+from sparkth.core.permissions.models import Group
+from sparkth.lib.db import get_async_session
+from sparkth.lib.permissions import GROUP_CREATE, GROUP_DELETE, GROUP_READ, GROUP_UPDATE
+
+router = APIRouter()
+
+
+def _group_to_response(group: Group) -> GroupResponse:
+    """Assemble a GroupResponse, guarding the persisted-id invariant.
+
+    A persisted/refreshed group always has an id; a real guard (not a bare ``assert``, which
+    ``python -O`` strips) is what keeps the invariant enforced in optimized builds.
+    """
+    if group.id is None:
+        raise RuntimeError(f"Group has no id: {group!r}")
+    return GroupResponse(id=group.id, name=group.name, description=group.description)
+
+
+@router.get(
+    "",
+    response_model=list[GroupResponse],
+    dependencies=[Depends(GROUP_READ.require_in_global_scope())],
+)
+async def list_groups(session: AsyncSession = Depends(get_async_session)) -> list[GroupResponse]:
+    """List all groups. Returns a list of GroupResponse (id, name, description)."""
+    return [_group_to_response(group) for group in await group_engine.list_groups(session)]
+
+
+@router.post(
+    "",
+    response_model=GroupResponse,
+    status_code=status.HTTP_201_CREATED,
+    dependencies=[Depends(GROUP_CREATE.require_in_global_scope())],
+)
+async def create_group(payload: GroupCreate, session: AsyncSession = Depends(get_async_session)) -> GroupResponse:
+    """Create a group. Returns the created GroupResponse (id, name, description)."""
+    return _group_to_response(await group_engine.create_group(payload.name, payload.description, session))
+
+
+@router.get(
+    "/{group_id}",
+    response_model=GroupResponse,
+    dependencies=[Depends(GROUP_READ.require_in_global_scope())],
+)
+async def get_group(group_id: int, session: AsyncSession = Depends(get_async_session)) -> GroupResponse:
+    """Fetch a group by id. Returns its GroupResponse (id, name, description)."""
+    return _group_to_response(await group_engine.get_group(group_id, session))
+
+
+@router.patch(
+    "/{group_id}",
+    response_model=GroupResponse,
+    dependencies=[Depends(GROUP_UPDATE.require_in_global_scope())],
+)
+async def update_group(
+    group_id: int, payload: GroupUpdate, session: AsyncSession = Depends(get_async_session)
+) -> GroupResponse:
+    """Update a group's name and/or description. Returns the updated GroupResponse."""
+    return _group_to_response(await group_engine.update_group(group_id, payload.name, payload.description, session))
+
+
+@router.delete(
+    "/{group_id}",
+    status_code=status.HTTP_204_NO_CONTENT,
+    dependencies=[Depends(GROUP_DELETE.require_in_global_scope())],
+)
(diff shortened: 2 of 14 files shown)


sparkth/cli/groups.py
```diff
@@ -0,0 +1,109 @@
...
+    InvalidScopeObjectId,
+    PermissionScopeNotFound,
+    RoleNotFound,
+)
+from sparkth.lib.plugins import get_plugin_loader
+
+app = typer.Typer(help="User-group management commands")
+
+
+@app.command("add-member")
```
hamza-56, line 53: This is a one-way door right now: we can add members and assign roles, but `remove_group_member` and `revoke_role_from_group` exist in the engine (and are exported through lib) with nothing wired to them, no CLI, no REST. Since membership is CLI-only by design, there's currently no way to remove someone from a group short of raw SQL. The roles CLI has the same assign-only asymmetry, but "remove someone from a group" feels like a day-one operation. Fine as a fast-follow if it's out of scope here, just don't want it forgotten.

sparkth/cli/groups.py
```diff
@@ -0,0 +1,109 @@
...
+        ).first()
+        if user is None or user.id is None:
+            typer.secho(f"User '{identifier}' not found!", fg=typer.colors.RED)
+            raise typer.Exit(code=1)
+        try:
+            group = await get_group_by_name(group_name, session)
+        except GroupNotFound:
+            typer.secho(f"Group '{group_name}' not found!", fg=typer.colors.RED)
+            raise typer.Exit(code=1) from None
+        if group.id is None:
```
hamza-56: This `group.id is None` branch is dead code (a group loaded from the DB always has an id), and the whole GroupNotFound + "not found!" + Exit dance is copy-pasted in both commands. A tiny `_resolve_group(name, session)` helper would remove both copies and the dead branch.

sparkth/api/v1/permissions/schemas.py
```diff
@@ -32,3 +32,27 @@ class RoleResponse(BaseModel):
...
 class RolePermissionAdd(BaseModel):
     permission: str = Field(min_length=1, max_length=100)
+
+
+class GroupCreate(BaseModel):
+    name: str = Field(min_length=1, max_length=50)
+    description: str | None = Field(default=None, max_length=255)
+
+
+class GroupUpdate(BaseModel):
```
hamza-56, line 46: Small gap in the PATCH semantics: `None` means "unchanged", so once a group has a description there's no way to clear it back to null. Inherited from RoleUpdate so I wouldn't block on it, but worth a note in the comment (or an explicit sentinel eventually) so it reads as a known limitation rather than an oversight.

docs/guides/permissions.md
```diff
@@ -74,8 +74,8 @@ descendants. Object-bearing multi-level cascade (e.g. a grant at one `org` apply
...
 |---|---|---|
 | **Scopes** | `global`, `whitelist`, `role` | `global` is the objectless root (applies platform-wide). `whitelist` is an objectless singleton nested under `global` for the registration whitelist. `role` is object-bearing (nested under `global`) so role management can be delegated per-role by that role's id. A `global` grant cascades down to both. |
-| **Permissions** | `email.whitelist.{read,create,delete}`; `role.{create,read,update,delete}`; `permission.read` | Gate the registration email-whitelist endpoints, the role-management API, and listing the assignable permission vocabulary respectively. |
-| **Roles** | `admin` | Grants the three `email.whitelist.*`, the four `role.*`, and `permission.read`. A seed migration also assigned it at `global` to every account that was a superuser when the migration ran — a one-time backfill. |
+| **Permissions** | `email.whitelist.{read,create,delete}`; `role.{create,read,update,delete}`; `group.{create,read,update,delete}`; `permission.read` | Gate the registration email-whitelist endpoints, the role-management API, the group-management API, and listing the assignable permission vocabulary respectively. |
```
hamza-56, line 77: The guide only gains the four permission names, but nothing in it explains groups themselves: that membership and group-role grants are CLI-managed, that `can()`/`has_role` now resolve through group assignments, and that delete is blocked while a group has active assignments. Given the docs rule for permission-system changes, a short Groups section here would go a long way.

hamza-56 (commented):
API layer is clean. Comments are mostly about the CLI and docs:

sparkth/cli/groups.py
abdulrafey1: Done in b0edb29 — added `groups remove-member` and `groups revoke-role-from-group` (with tests), so both doors swing both ways now. Revoking a role that isn't assigned at the exact scope is a no-op, mirroring the engine's `revoke_role_from_group`. The roles CLI asymmetry is untouched (pre-existing, out of scope here).

sparkth/cli/groups.py
abdulrafey1: Done in b0edb29 — extracted `_resolve_group_id` (and `_resolve_user`, since the new remove-member command would have duplicated that lookup too). The `id is None` check survives only inside the helper as a bare one-line mypy narrowing guard with a comment saying it's unreachable.

sparkth/api/v1/permissions/schemas.py
abdulrafey1: Done in b0edb29 — the comment on `GroupUpdate` now names it a known limitation shared with `RoleUpdate` (clearing a description back to null needs an explicit sentinel).

docs/guides/permissions.md
abdulrafey1: This is covered by the next PR in the stack — #546 adds a full **Groups** section to `docs/guides/permissions.md`: flat-by-design rationale, CLI-managed membership/grants (now including the new remove/revoke commands), `can()`/`has_role()` resolving through group assignments with the same scope cascade, soft-delete semantics, and the delete-blocked-while-assignments-active rule. This PR only carries the shipped-vocabulary table rows so each layer documents what it introduces.

sparkth/api/v1/permissions/routes/groups.py
```diff
@@ -0,0 +1,82 @@
...
+
+@router.post(
+    "",
+    response_model=GroupResponse,
+    status_code=status.HTTP_201_CREATED,
+    dependencies=[Depends(GROUP_CREATE.require_in_global_scope())],
+)
+async def create_group(payload: GroupCreate, session: AsyncSession = Depends(get_async_session)) -> GroupResponse:
+    """Create a group. Returns the created GroupResponse (id, name, description)."""
+    return _group_to_response(await group_engine.create_group(payload.name, payload.description, session))
```
regisb, line 50: Are we not catching exceptions thrown by group_engine.create_group in these endpoints? E.g: I believe we should raise 40x if a group already exists?

sparkth/api/v1/permissions/routes/groups.py
```diff
@@ -0,0 +1,82 @@
+"""Group sub-router (mounted at ``/groups``): CRUD for user groups (issue #519).
```
regisb, line 1: I like how this router is really clean.

sparkth/cli/groups.py
```diff
@@ -0,0 +1,177 @@
...
+    contradict each other (enforced by the engine's assign_role_to_group).
+    """
+    # Load plugins first so plugin-declared scope kinds are registered before validation,
+    # and a mistyped --scope fails loudly instead of persisting a no-op assignment.
+    get_plugin_loader()
+    try:
+        permission_scope = get_permission_scope(scope)
+    except PermissionScopeNotFound:
+        typer.secho(f"Unknown scope kind: '{scope}'", fg=typer.colors.RED)
+        raise typer.Exit(code=1) from None
```
regisb, line 127: nit: I'm not too confortable running typer.Exit in non-CLI functions. I understand this specific function is only called in CLI, but it might be a future footgun during a refactoring. Feel free to ignore this comment.

regisb (approved):
Super clean PR. Just one minor comment (feel free to ignore) and question.

sparkth/api/v1/permissions/routes/groups.py
abdulrafey1: `group_engine.create_group` raises `GroupAlreadyExists` (See this [line](https://github.com/edly-io/sparkth/pull/544/changes#diff-f939faef0e4b0633a51b9b3f4ffdeccbab7a3add4f1ea7d27a06fb265114f171R41) in PR #544 )
Furthermore the `GroupAlreadyExists` exception is being added to the central exception management system in this PR in `sparkth/api/v1/permissions/__init__.py`, so it will be mapped to a `409`

sparkth/cli/groups.py
abdulrafey1: Hmmmm, lets revisit this if it ever becomes a problem.

PR #546, opened 2026-07-27
docs: document permission groups

abdulrafey1:
## Closes: [Permission system needs groups to grant roles to users in bulk](https://github.com/edly-io/sparkth/issues/519)

Part 3 of 3 stacked PRs — stacked on #545 (management surface). Merging this one closes #519.

## What
Documents the permission-groups feature so the guide, README, and directory map stay accurate with the code the first two PRs of the stack introduced.

## Changes
- docs: "Groups" section in `docs/guides/permissions.md` — concept (flat groups, bulk grants), resolution semantics (own ∪ group grants through the same scope chain), soft-delete behaviour, the `source` column, management surface, and CLI examples; "The model" table list extended with the three group tables
- docs: README "Permissions" overview section linking the guide
- docs: `CLAUDE.md` key-directories line for `core/permissions/` now mentions groups

## How to Test
1. `make docs` — the docs site builds cleanly and the permissions guide renders the new Groups section.
2. Cross-check the CLI examples against `make cli -- groups --help`.

## Notes
None — documentation only.

_This PR description was written with the assistance of an LLM (Claude)._

PR #549, opened 2026-07-27
feat(core): add inert organization structure (units + memberships)

### LINKED ISSUE #548: Organizational structure has no representation
## Summary

Sparkth has no representation of organizational structure. Departments, faculties,
universities — the containers real institutions are made of — have no entity anywhere in
the data model, and nothing records which unit a user sits in.

## Where

- There is no `sparkth/core/org` package (or any equivalent): no unit table, no
  user↔unit link.
- `sparkth/core/permissions/models.py` — groups (#519) are **flat by design**; they are a
  set of users, not a hierarchy, and deliberately carry no org-chart semantics.
- The market research on #519 concluded every surveyed LMS (Canvas, Totara, Moodle)
  models "teachers belong to departments, which belong to universities" as a separate
  container hierarchy — a shape Sparkth currently cannot express at all.

## Why it matters

- **Rule-driven group membership has nothing to reference.** The planned follow-up to
  #519 (dynamic membership: "everyone in the CS department is in this group") needs an
  org tree to point rules at. Without it, group membership stays forever manual.
- **"Who sits where" has no answer.** Institutional reporting, auditing, and any future
  org-scoped feature would each have to invent their own ad-hoc structure.
- **Bulk operations stay per-user.** An HR event ("this person joined the Faculty of
  Science") has no single row to write; its access consequences must be hand-managed.

## Context

Part of the permission-system EPIC #420, as the second stage of the #519 follow-up
phasing (flat groups → org tree → rules). The org tree is deliberately **inert**: it
grants nothing and no permission check reads it — its permission effect arrives later,
indirectly, through rule-driven group membership.

abdulrafey1:
## Closes: [Organizational structure has no representation](https://github.com/edly-io/sparkth/issues/548)

Part 1 of 3 stacked PRs (core → management surface → docs); stacked on the #519 groups stack (#546). Does not close the issue on its own.

## What
Adds the organizational structure as inert people-classification data — an `OrganizationalUnit` hierarchy and `OrganizationMembership` user↔unit seats — so the org chart finally has a home; it grants nothing and no permission check reads it.

## Changes
- feat(core): new `sparkth/core/organization/` domain — `OrganizationalUnit` (adjacency + materialized id path `/1/7/42/`, sibling-unique names; table `organizational_unit`) and `OrganizationMembership` (soft-deleted history, partial-unique active seat; table `organization_membership`) with migration `94c26cf76059`
- feat(core): unit CRUD engine — create derives the path from the parent; `move_organizational_unit` re-parents with a one-comparison cycle check and rewrites the whole subtree in a single portable UPDATE (SQLite + PostgreSQL); delete refuses with `OrganizationalUnitInUse` while children or active members exist
- refactor(core): the bulk path-rewrite's in-session staleness is handled at the write site (targeted awaited refresh), keeping `get_organizational_unit` on the shared bare `session.get` convention
- feat(core): membership engine — idempotent race-safe `add_organization_member`, soft-delete `remove_organization_member`, `get_organization_members`; membership deliberately does NOT inherit across the tree

## How to Test
1. `make test.backend.pytest` — full suite green; coverage in `tests/organization/` (models, path maintenance, moves, cycle prevention, delete guards, memberships).
2. `make mypy && make lint.backend` — clean.
3. Migration `94c26cf76059` verified up **and** down against PostgreSQL.
4. Inert invariant: `grep -rn "core.organization" sparkth/core/permissions/` — no engine coupling in either direction.

## Notes
- **Migration required**: yes — `94c26cf76059` (two new tables, no data changes).
- The tree is deliberately inert this phase; rule-driven group membership (the phase that connects it to access) comes later, per the #519 follow-up phasing.
- Management surface (permissions, REST, CLI) arrives in the next PR of the stack.


sparkth/core/organization/models.py
```diff
@@ -0,0 +1,62 @@
...
+        ),
+    )
+
+    id: int | None = Field(default=None, primary_key=True)
+    name: str = Field(max_length=100, index=True)
+    # Display/reporting metadata ("university" | "faculty" | "department" | free-form),
+    # never engine input; institutions disagree on structure, so no level enum.
+    kind: str | None = Field(default=None, max_length=50)
+    parent_id: int | None = Field(default=None, foreign_key="organizational_unit.id", index=True)
+    path: str = Field(max_length=500, index=True)
```
hamza-56: The docstring's promise ("one indexable LIKE-prefix comparison") doesn't hold on Postgres as-is: a plain btree on a text column with a normal (non-C) collation can't serve `LIKE 'prefix%'`, so `ix_organizational_unit_path` will be ignored for descendant lookups on the production DB. The index needs `text_pattern_ops` / `varchar_pattern_ops` (via `postgresql_ops`) or a C collation on the column. Worth fixing in this PR since the migration is still unreleased.

sparkth/core/organization/units.py
```diff
@@ -0,0 +1,169 @@
...
+    # objects — e.g. a descendant loaded earlier in this session — rather than the whole
+    # instance. A bare session.get() only re-queries when the *whole* instance is expired,
+    # so a partially-expired object would be handed back unchanged, and the first plain
+    # `.path` access on it would attempt an illegal sync lazy-load under AsyncSession.
+    # Refresh exactly those affected objects now, inside this function, rather than
+    # leaving them expired for whichever caller touches them next — identified via
+    # expired_attributes so this never itself reads a possibly-expired column, and never
+    # disturbing unrelated objects (like the new parent) that the rewrite didn't touch.
+    # Session-wide by design: fetch-sync already limited expiry to UPDATE-touched rows, so filtering more is redundant.
+    for loaded in session.identity_map.values():
```
hamza-56: This loop is one SELECT per stale in-session unit, and it takes a 15-line comment to explain why it's needed. Running the UPDATE with `execution_options(synchronize_session=False)` and doing a single `session.expire_all()` before the commit gets the same safety in constant time, and the comment can go with it.

sparkth/core/organization/units.py
```diff
@@ -0,0 +1,169 @@
...
+    active_member = (
+        await session.exec(
+            select(OrganizationMembership.id)
+            .where(OrganizationMembership.organizational_unit_id == unit_id, OrganizationMembership.is_deleted == False)
+            .limit(1)
+        )
+    ).first()
+    if active_member is not None:
+        raise OrganizationalUnitInUse(unit_id)
+    for membership in (
```
hamza-56: Same note as `delete_group` on #544: this loads every membership row and deletes them one at a time. A single bulk `delete(OrganizationMembership).where(organizational_unit_id == ...)` does it in one statement.

hamza-56 (commented):
Great work overall!

sparkth/core/organization/models.py
abdulrafey1: Fixed in 44c6b6d — the index is now declared explicitly with `postgresql_ops={"path": "varchar_pattern_ops"}` (model and migration; the migration is edited in place since it's unreleased and this branch is its only consumer). Verified against a real Postgres: the DDL renders as `USING btree (path varchar_pattern_ops)`, and up/down both apply cleanly. A metadata test pins the ops so it can't silently regress.

sparkth/core/organization/units.py
abdulrafey1: Pushing back on this one: `synchronize_session=False` + `expire_all()` isn't equivalent under `AsyncSession`. With no fetch-sync the UPDATE expires nothing (in-session units keep silently stale paths), and `expire_all()` then expires *every* object in the session — so the next plain attribute access on any previously loaded unit raises `MissingGreenlet` in the caller instead of lazy-loading. That's exactly the failure this loop absorbs: it refreshes only the UPDATE-touched units (typically 0–2 objects — only ones already in the session), so callers never inherit expired state. We did try the expire-all variant first and it broke callers in exactly this way. What I did take from the comment: the 15-line essay is now 5 lines (44c6b6d), including a note on why expire_all isn't a substitute.

sparkth/core/organization/units.py
abdulrafey1: Done in 44c6b6d — historical membership rows now go in one bulk DELETE, same pattern as the `delete_group` fix on #544.

sparkth/core/organization/memberships.py
```diff
@@ -0,0 +1,83 @@
+"""User ↔ organizational-unit membership management.
+
+Module-level async functions that flush (callers own the transaction boundary), mirroring
+the group-membership engine. Membership is HR truth: it grants nothing, and sitting in a
+unit does not imply sitting in its ancestors or descendants.
```
regisb, line 5: If we don't have ancestor inheritance then I'm not sure what's the point of having a unit hierarchy. The way I see it, sub-orgs should be considered as included in parent orgs. I.e: team purple is part of Edly, which is part of Arbisoft. But maybe I'm missing something? What led you to this approach?

regisb (commented):
I think we need to discuss this feature a little more. It's not an urgently-needed feature. Also, this feature seems to be largely disconnected from the rest of the codebase. So we can take some time to ponder.

For instance:

1. if groups are unrelated to organizations, then do we want organizations to be part of the core or a plugin?
2. In terms of naming convention, I think we should shorten "organization" and "organizational" to "org". It makes the code more readable IMHO.
3. What's the business case that we want to address? If this is related to permissions, then can we see a use case where organizations and permissions interact?
4. If this feature is end-user-facing, then can we see (and discuss) some screenshots?

sparkth/core/organization/memberships.py
abdulrafey1: ***On inheritance / "what's the point of the hierarchy"***: inclusion isn't dropped, it just isn't stored by duplicating rows upward. The materialized path (/1/7/42/) is the inheritance mechanism — "everyone in Arbisoft" is one indexed prefix query over descendant units, unioning their members. So team purple ⊆ Edly ⊆ Arbisoft holds; it's computed at read time from the path rather than written as extra membership rows at every ancestor.

**_The deliberate split is_**: membership = HR truth ("where does this person sit"), inheritance = policy ("does a grant on CS Dept cover its sub-units"). Baking upward inheritance into the membership table would force one answer for everyone and require cascade maintenance on every move/add/remove. Keeping it out leaves the rollup a choice, made in the rule layer — which is the next phase (#553), per the #519 → #548 phasing (flat groups → org tree → rules).

You're right that this PR shows no payoff on its own — the tree is intentionally inert here. The value lands when rules consume the path.

**_Core vs plugin_**: it has to be core. The permission core will read the org tree to drive dynamic group membership, and core can't depend on a plugin — that would invert the plugin→core dependency direction the architecture relies on. It's also cross-cutting reference data (permissions, reporting, audit all need "who sits where"), not an optional integration like the canvas/slack plugins. That's why it lives alongside permissions/, analytics/, audit/.

Business case + how org and permissions interact: #548 has the full case — every surveyed LMS (Canvas, Totara, Moodle) models "teachers → departments → universities" as a container hierarchy, which Sparkth currently can't express. The concrete permission interaction (later phase):

1. HR event: Alice joins Faculty of Science → one membership row.
2. Rule: "members of Faculty of Science and its sub-units → group Science Staff."
3. Group Science Staff carries a GroupRoleAssignment: role Instructor at a scope.
4. Alice gets Instructor access automatically; when she leaves, dropping the one row removes it.

So org structure never grants anything directly — it interacts with permissions indirectly, through rules → groups → role@scope.

**_One clarification worth flagging_**: the org hierarchy and the PermissionScope hierarchy are two separate trees. Org tree = who (people classification, rolls up via path in rules). PermissionScope chain = what (resource boundary, e.g. quiz under course, cascades grants). If the concern was that the org tree should cascade permissions itself, that cascade lives in scopes, not here.

PR #550, opened 2026-07-27
feat(api): add organization management surface

### LINKED ISSUE #548: Organizational structure has no representation

abdulrafey1:
## Closes: [Organizational structure has no representation](https://github.com/edly-io/sparkth/issues/548)

Part 2 of 3 stacked PRs — stacked on #549 (organization core). Management surface for the structure that PR introduces. Does not close the issue on its own.

## What
Adds the management surface for the organizational structure: `organization.unit.*` permissions, REST CRUD + re-parenting for units, CLI membership commands, and the `sparkth/lib/organization.py` public façade — following the established split (structure via REST, principal edges via CLI).

## Changes
- feat(core): declare `organization.unit.create` / `organization.unit.read` / `organization.unit.update` / `organization.unit.delete`, exported via `sparkth.lib.permissions` (constants `ORGANIZATION_UNIT_*`); data migration `ccd39a325759` grants them to the seeded `admin` role (no-op when unseeded)
- feat(core): `sparkth/lib/organization.py` façade — the one import surface for the organization engine (units, memberships, exceptions)
- feat(api): `/api/v1/organization/units` — list/create/get/patch/delete, gated at the global scope; PATCH distinguishes an unset `parent_id` from an explicit `null` (move to root) and re-parents via the engine's cycle-checked move; `OrganizationalUnitNotFound → 404`, `OrganizationalUnitAlreadyExists`/`OrganizationalUnitInUse`/`OrganizationCycleError → 409` via the central registry
- feat(cli): `organization add-member` / `organization remove-member` — units addressed by id (names are only sibling-unique)
- docs: `organization.unit.*` rows in the permissions guide's shipped table
- chore(frontend): regenerated `frontend/lib/api/generated.ts`

## How to Test
1. `make test.backend.pytest` — full suite green; coverage in `tests/organization/test_permissions.py`, `tests/api/v1/test_organization_units.py` (incl. move-to-root-via-null, cycle→409, PATCH/DELETE permission gates), `tests/cli/test_organization.py`.
2. `make migrations` — applies `ccd39a325759` (verified up **and** down against PostgreSQL).
3. As an admin: `POST /api/v1/organization/units {"name": "University X"}` → 201 with `path` `/{id}/`; `PATCH` with `{"parent_id": null}` moves to root; moving a unit under its own descendant → 409.
4. `make cli -- organization add-member <username> <unit-id>` seats the user; `remove-member` soft-deletes the seat.

## Notes
- **Migration required**: yes — data migration `ccd39a325759` (admin grants only, no schema change).
- Gating is GLOBAL-scope only; a per-unit delegation scope waits on object-bearing cascade (#420 Phase 2).
- Known accepted edge (inherited from the design): a combined rename+move PATCH is not atomic across the two engine calls — flagged for a follow-up if combined updates become common.


sparkth/api/v1/organization/routes/units.py
```diff
@@ -0,0 +1,105 @@
...
+    unit_id: int, payload: OrganizationalUnitUpdate, session: AsyncSession = Depends(get_async_session)
+) -> OrganizationalUnitResponse:
+    """Rename/re-kind a unit and/or move it (a provided parent_id — null = make root — re-parents).
+
+    Rename applies before the move so a combined rename+move validates the new name against
+    the destination's siblings.
+    """
+    unit = await unit_engine.get_organizational_unit(unit_id, session)
+    if payload.name is not None or payload.kind is not None:
+        unit = await unit_engine.update_organizational_unit(unit_id, payload.name, payload.kind, session)
```
hamza-56: A combined rename+move isn't atomic: both engine calls commit their own transaction, so if the rename lands and the move then 409s (cycle, or name taken at the destination), the PATCH fails but the rename stays applied. There's also a validation quirk in the ordering: the rename checks the *current* parent's siblings, so "rename to X while moving to B" gets a 409 whenever the current parent already has an X, even if B is free (the docstring says the opposite). The tests cover rename-only and move-only but not the combined path, which is exactly where both of these bite. I'd fold this into a single engine operation (or one transaction without the intermediate commits) and add a combined-PATCH test.

hamza-56 (commented):
One behavioral issue:

sparkth/api/v1/organization/routes/units.py
abdulrafey1: All three points confirmed and fixed. The engine (in #549's layer, b92b0a9) gained `patch_organizational_unit`, which applies rename/re-kind/re-parent in one transaction: validation runs against the *final* state — the name is checked against the destination's siblings, with tests for both directions (rename-to-colliding-name-at-current-parent succeeds when the destination is free, and rename-away-from-a-destination-collision succeeds too) — and nothing commits until every check passes. `update_organizational_unit` / `move_organizational_unit` now delegate to it, unchanged in signature and behaviour. The PATCH route (c3db52a) makes a single engine call, its docstring now tells the truth, and there are combined-PATCH API tests including the atomicity case (rename + cycle-move → 409 and the rename is not applied).

PR #558, opened 2026-07-27
refactor(frontend): gate analytics + admin nav on permissions

zaira-bibi:
> 📚 **Stacked on [#539](https://github.com/edly-io/sparkth/pull/539)** (analytics dashboard) → which is on [#533](https://github.com/edly-io/sparkth/pull/533) (`GET /permissions/can`). Review/merge #533 → #539 → this; the diff is against #539's branch. Follow-up to the analytics-dashboard nav gating; part of the [analytics epic #434](https://github.com/edly-io/sparkth/issues/434).

## What
Render the **Analytics** and **Admin** sidebar entries through a small config-driven, permission-gated mechanism that mirrors how plugin entries render — replacing the ad-hoc `canViewAnalytics` boolean (Analytics) and the coarse `is_admin` flag (Admin).

## Changes
- refactor(frontend): `GATED_NAV` config + `resolveNavPermissions` — resolves each gated entry once per token via `GET /permissions/can` (one check per item, failing closed) into a `Record<key, boolean>`
- refactor(frontend): shared `NavItem` component (extracted from the repeated nav-link blocks)
- refactor(frontend): the dashboard layout resolves `navPermissions` once and threads it to both sidebars, which render `GATED_NAV.filter(i => navPermissions?.[i.key]).map(NavItem)`; `canViewAnalytics` removed
- refactor(frontend): Admin nav gates on `email.whitelist.read` (whitelist scope) instead of `is_admin`
- fix(frontend): the whitelist page's client guard is aligned to the same `email.whitelist.read` check (was `!is_admin`), so nav visibility and page access agree
- test(frontend): `resolveNavPermissions`, `NavItem`, rewritten `AppSidebar` gating (per key), `WhitelistPage` guard

## How to Test
1. `make test.frontend.vitest` → green (190 tests).
2. `cd frontend && bun run typecheck`, `make lint.frontend`, `make test.frontend.api` → all clean.
3. Manual (backend + frontend running): a user with `analytics.read` sees the Analytics entry; a user with `email.whitelist.read` sees the Admin entry and can open `/dashboard/admin/whitelist`; a user with neither sees neither entry, and direct navigation to the whitelist page redirects to `/dashboard`.

## Notes
- **Stacked PR** — merge after #539 (`gh stack sync --prune` once #539 lands).
- No backend / `generated.ts` change; no new dependencies. Single permission check per gated entry (no batch endpoint).
- UI gating is a convenience, **not a security boundary** — every endpoint still enforces its own permission and returns 403; the client checks fail closed.

### DIFF
--- a/frontend/app/dashboard/admin/whitelist/WhitelistPage.tsx
+++ b/frontend/app/dashboard/admin/whitelist/WhitelistPage.tsx
@@ -1,9 +1,10 @@
 "use client";
 
-import { useReducer, useEffect, useCallback, type SubmitEvent } from "react";
+import { useReducer, useEffect, useState, useCallback, type SubmitEvent } from "react";
 import { redirect } from "next/navigation";
 import { Plus, Trash2, Mail, Globe } from "lucide-react";
 import { useAuth } from "@/lib/auth-context";
+import { checkPermission } from "@/lib/permissions";
 import {
   getWhitelist,
   addWhitelistEntry,
@@ -78,10 +79,30 @@ const initialState: WhitelistState = {
 };
 
 export default function WhitelistPage() {
-  const { token, user } = useAuth();
+  const { token } = useAuth();
   const [state, dispatch] = useReducer(reducer, initialState);
 
-  if (user && !user.is_admin) {
+  const [access, setAccess] = useState<"checking" | "allowed" | "denied">("checking");
+
+  useEffect(() => {
+    // Reset on every token change so a re-login can't briefly show the previous user's page
+    // while the new user's check is still in flight (fail closed to the checking state).
+    setAccess("checking");
+    if (!token) return;
+    let active = true;
+    checkPermission(token, "email.whitelist.read", { scope: "whitelist" })
+      .then((allowed) => {
+        if (active) setAccess(allowed ? "allowed" : "denied");
+      })
+      .catch(() => {
+        if (active) setAccess("denied"); // fail closed
+      });
+    return () => {
+      active = false;
+    };
+  }, [token]);
+
+  if (access === "denied") {
     redirect("/dashboard");
   }
 
@@ -101,8 +122,8 @@ export default function WhitelistPage() {
   }, [token]);
 
   useEffect(() => {
-    fetchEntries();
-  }, [fetchEntries]);
+    if (access === "allowed") fetchEntries();
+  }, [access, fetchEntries]);
 
   const handleAdd = async (e: SubmitEvent) => {
     e.preventDefault();
@@ -143,6 +164,17 @@ export default function WhitelistPage() {
       dispatch({ type: "SET_DELETING", deleting: false });
     }
   };
+
+  if (access === "checking") {
+    return (
+      <div className="flex items-center justify-center min-h-screen bg-background">
+        <div className="text-center">
+          <Spinner className="mx-auto mb-4" />
+          <p className="text-muted-foreground">Loading whitelist...</p>
+        </div>
+      </div>
+    );
+  }
 
   if (state.loading) {
     return (
--- a/frontend/app/dashboard/layout.tsx
+++ b/frontend/app/dashboard/layout.tsx
@@ -3,7 +3,7 @@
 import { useState, useEffect } from "react";
 import "@/lib/plugins";
 import { useAuth } from "@/lib/auth-context";
-import { checkPermission } from "@/lib/permissions";
+import { resolveNavPermissions } from "@/components/NavItem";
 import { PluginProvider } from "@/lib/plugins/context";
 import AppSidebar from "@/components/AppSidebar";
 import MobileSidebar from "@/components/MobileSidebar";
@@ -32,6 +32,7 @@ function DashboardContent({
   children,
   user,
   logout,
+  navPermissions,
 }: {
   children: React.ReactNode;
   user: {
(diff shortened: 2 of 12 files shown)


zaira-bibi:
@claude-review

regisb (approved):
Again, I'm definitely not the right person to review this but as far as I'm concerned, this is good to go.

hamza-56 (commented):
Nice follow-up to #539

frontend/app/dashboard/admin/whitelist/WhitelistPage.tsx
```diff
@@ -78,10 +79,27 @@ const initialState: WhitelistState = {
...
+      })
+      .catch(() => {
+        if (active) setAccess("denied"); // fail closed
+      });
+    return () => {
+      active = false;
+    };
+  }, [token]);
+
+  if (access === "denied") {
```
hamza-56, line 105: The access check and `fetchEntries` both run on mount in parallel, so a denied user still fires the `getWhitelist` request, gets a 403 dispatched into the error state, and only then redirects. The old `is_admin` guard was synchronous so the fetch never ran for denied users. Gating the fetch on `access === "allowed"` restores that and saves the wasted 403.

frontend/components/AppSidebar.tsx
```diff
@@ -33,8 +33,8 @@ interface AppSidebarProps {
     avatar?: string;
     plan?: string;
     is_admin?: boolean;
```
hamza-56: `is_admin` is dead now: nothing in AppSidebar or MobileSidebar reads it since the Admin entry moved to GATED_NAV, but both prop interfaces still declare it and the layout still passes it. Dropping it finishes what this PR started, `user` becomes purely display data.

frontend/components/AppSidebar.tsx
```diff
@@ -232,53 +233,17 @@ export default function AppSidebar({
...
-                  title={isCollapsed ? "Admin" : undefined}
-                >
-                  <Shield className="w-5 h-5 flex-shrink-0" />
-                  {!(isCollapsed && variant === "desktop") && (
-                    <span className="font-medium">Admin</span>
-                  )}
-                </Link>
-              </div>
-            )}
+            {GATED_NAV.filter((item) => navPermissions?.[item.key]).map((item) => (
```
hamza-56, line 234: The extraction stopped halfway: Resources, My Plugins, and AI Keys still carry three copies of the original hand-rolled Link block right above this. Migrating them to NavItem too deletes ~60 lines and leaves exactly one place that knows what a sidebar link looks like. Happy for it to be a follow-up if you'd rather keep this PR scoped to the gated entries.

frontend/app/dashboard/admin/whitelist/WhitelistPage.tsx
```diff
@@ -78,10 +79,27 @@ const initialState: WhitelistState = {
...
+      .then((allowed) => {
+        if (active) setAccess(allowed ? "allowed" : "denied");
+      })
+      .catch(() => {
+        if (active) setAccess("denied"); // fail closed
+      });
+    return () => {
+      active = false;
+    };
+  }, [token]);
```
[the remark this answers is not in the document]
abdulrafey1: `access` is never reset when `token` changes, so after a re-login the previous user's `"allowed"` stays in effect until the new user's check settles.

One line at the top of the effect fixes it:
```
useEffect(() => {
  setAccess("checking");
  if (!token) return;
  ...
```

frontend/app/dashboard/layout.tsx
```diff
@@ -74,18 +76,21 @@ function DashboardContent({
...
+      .then((perms) => {
+        if (active) setNavPermissions(perms);
       })
       .catch((error) => {
-        if (active) setCanViewAnalytics(false);
-        console.error("Failed to check analytics permission:", error);
+        // resolveNavPermissions fails closed per item; this is a defensive guard. Clear
+        // explicitly rather than relying on the initial {}: on a token change the state
+        // still holds the previous user's answers.
+        if (active) setNavPermissions({});
```
abdulrafey1: on a token change the state still holds the previous user's answers. Resetting at the top of the effect covers both paths and makes the catch-side clear redundant:
```
useEffect(() => {
  setNavPermissions({});
  if (!token) return;
  ...
```

frontend/components/NavItem/NavItem.tsx
abdulrafey1, line 1: Lets create a new directory for better codebase management. Lets put `NavItem` component in a `components/NavItem` directory and also move the related logic (`gatedNav.ts`) inside the directory.
Furthermore, lets move the tests files into a separate tests directory just like we have for the backend

frontend/app/dashboard/layout.tsx
```diff
@@ -40,9 +41,9 @@ function DashboardContent({
     avatar?: string;
     plan?: string;
     is_admin?: boolean;
```
abdulrafey1: This PR removes the last consumer of `user.is_admin` in the sidebar chain (the Admin entry now gates on `navPermissions.admin`), but the field is still declared and threaded through `layout.tsx → AppSidebar → MobileSidebar` and populated in 1userInfo. canViewAnalytics1 was cleaned out of these same interfaces in this PR — `is_admin` deserves the same treatment, or the next reader will assume something still consumes it. (The `is_admin` field on the API user type itself is fine — this is only about the sidebar prop plumbing.)

frontend/components/AppSidebar.tsx
zaira-bibi: Delegating this to a separate PR.

hamza-56 (approved):

abdulrafey1 (approved):

PR #563, opened 2026-07-29
feat(docker): add production compose stack

### LINKED ISSUE #562: Self-hosted production deployments need a production-grade Docker Compose stack
## Summary

Sparkth has no supported way to run the full stack in production with Docker Compose. The only compose file in the repository, [`docker-compose.yml`](../blob/main/docker-compose.yml), is explicitly dev-only: it provides just the backing services (Postgres/TimescaleDB, Redis, Mailpit) for a natively running backend and frontend, with ports published on `127.0.0.1`.

hamza-56:
## What
Adds a self-contained production Docker Compose stack that runs the whole platform (app, TimescaleDB with the app and analytics databases, Redis), plus a one-shot `sparkth migrate` CLI command that applies both Alembic lineages and the continuous-aggregate backfill, replacing the README's "convert the dev compose yourself" instructions.

## Changes
- feat(docker): add docker-compose.prod.yml running the app, TimescaleDB (app + analytics DBs), and Redis; every service loads `.env` and `.env.local` explicitly, and compose fails to start if `.env.local` is missing
- feat(cli): add a `migrate` command that applies both Alembic lineages and backfills continuous aggregates in one step; `make migrations` now delegates to it
- docs: rewrite the README production section around the new stack (plain `docker compose` commands, migrations run explicitly after `up`)
- docs: note in .env, docker-compose.yml, and CLAUDE.md how the prod stack expects the connection URLs in `.env.local` and how the migrate command covers the backfill

## How to Test
1. `make docker.build` (or let compose pull `ghcr.io/edly-io/sparkth:latest`)
2. Create `.env.local` with the production overrides, pointing `DATABASE_URL`, `ANALYTICS_DATABASE_URL`, and `REDIS_URL` at the bundled containers (see the README's "Production" section)
3. `docker compose -f docker-compose.prod.yml up -d`
4. `docker compose -f docker-compose.prod.yml run --rm sparkth python -m sparkth.cli.main migrate` shows both Alembic lineages applied and the aggregate backfill run
5. `docker compose -f docker-compose.prod.yml ps` shows `db` and `redis` healthy and `sparkth` running
6. Open http://localhost:7727 (bundled frontend is served) and http://localhost:7727/docs
7. `docker compose -f docker-compose.prod.yml down`, then `up -d` again: data volumes survive
8. Remove `.env.local` temporarily and re-run `up -d`: compose refuses to start
9. `uv run pytest tests/cli/test_migrate.py` covers the new CLI command

## Notes
- New Alembic migrations: none. App code changes are limited to the new `sparkth/cli/migrate.py` command.
- `.env.local` is now mandatory for the production stack and must carry the connection URLs (`@db:5432`, `redis://redis:6379`).
- New compose-level knobs (interpolation-only, not app settings, so not added to .env): `SPARKTH_TAG` (image tag, default latest), `SPARKTH_HTTP_PORT` (published port, default 7727), `SPARKTH_HTTP_BIND` (bind address; set to 127.0.0.1 behind a same-host reverse proxy).
- Migrations are not run automatically on startup; run the migrate command on first start and after every upgrade.
- TLS and email are intentionally not part of the stack: run a reverse proxy in front (set `TRUSTED_PROXY_HOPS`) and point `SMTP_*` at a real provider in `.env.local`.


hamza-56:
@claude-review

regisb (changes requested):
Review pending. Please wait until I review this fully.

docker-compose.prod.yml
```diff
@@ -0,0 +1,117 @@
...
+#
+# Run via the `make prod.*` targets (prod.up, prod.down, prod.logs, ...), or
+# pass both env files yourself:
+#
+#   docker compose -f docker-compose.prod.yml --env-file .env --env-file .env.local up -d
+#
+# Both --env-file flags are required so the POSTGRES_* values from .env.local
+# reach the interpolated connection URLs below.
+
+name: sparkth-prod
```
regisb: This is unnecessary, let's remove this line.

docker-compose.prod.yml
```diff
@@ -0,0 +1,117 @@
...
+
+x-logging: &default-logging
```
regisb: No need to override default logging. Let users decide that for themselves. We should remove this section and all "logging" settings.

docker-compose.prod.yml
```diff
@@ -0,0 +1,117 @@
...
+  driver: json-file
+  options:
+    max-size: "10m"
+    max-file: "5"
+
+# .env.local overrides .env (later entries win) and may be absent.
+x-env-files: &env-files
```
regisb: I have a big issue with how we handle environment files here. It really boils down to the fact that .env.local is mandatory. So we should explicitely load both .env and .env.local in all services:

```
env_file: 
  - .env
  - .env.local
```

No need to check for presence in Makefile: let's just crash if .env.local is not present.

Also, no need to create x-env-files: we can just duplicate the .env/.env.local list in all services.

docker-compose.prod.yml
```diff
@@ -0,0 +1,117 @@
...
+  - .env
+  - path: .env.local
+    required: false
+
+# In-network URLs for the bundled services, overriding the localhost defaults
+# committed in .env. POSTGRES_PASSWORD is embedded, so it must be URL-safe.
+x-app-env: &app-env
```
regisb: Wait, why is this necessary? I suggest we add the list of explicit env variables to the sparkth service.

docker-compose.prod.yml
```diff
@@ -0,0 +1,117 @@
...
+      migrations:
+        condition: service_completed_successfully
+      db:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
+
+  # Applies both Alembic lineages, then backfills TimescaleDB continuous
+  # aggregates (idempotent). The app starts only after this succeeds.
+  migrations:
```
regisb: Let's not run migrations automatically on every application start. Instead, add instructions in the README that explain how to run migrations. The fact that we need to run 3 different commands should tell us that we need to simplify the way we run migrations...

docker-compose.prod.yml
```diff
@@ -0,0 +1,117 @@
...
+
+  db:
+    image: docker.io/timescale/timescaledb-ha:pg18
+    restart: unless-stopped
+    # Postgres parallel workers use /dev/shm; Docker's 64 MB default is too
+    # small under analytical load.
+    shm_size: 256mb
+    env_file: *env-files
+    volumes:
+      - postgres_data:/var/lib/postgresql
```
regisb, line 46: I personally prefer bind-mounted data volumes, but let's stick with named volumes for consistency with the dev environment (nothing to change here).

Makefile
```diff
@@ -145,6 +145,34 @@ docs.serve: ## Serve the documentation site with live reload (http://127.0.0.1:8
 docker.build: ## Build the Docker image
 	docker build -t ghcr.io/edly-io/sparkth:latest .
 
+##@ Production (Docker Compose)
+# Self-contained single-host deployment: app + TimescaleDB + Redis, with
+# migrations applied before the app starts (see docker-compose.prod.yml).
+# The prod compose file builds the app's connection URLs from POSTGRES_*, so
+# compose interpolation must read .env.local (when present) on top of .env;
+# these targets pass both --env-file flags for that reason.
+COMPOSE_PROD = docker compose -f docker-compose.prod.yml --env-file .env$(if $(wildcard .env.local), --env-file .env.local)
```
regisb: This Makefile is going to replicate all the docker compose commands (logs, up, down, etc.) but without all the nice CLI options. So I suggest we don't add `make` commands. Instead, the README instructs to run `docker compose -f docker-compose.prod.yml up -d`, then apply migrations.

regisb (changes requested):

docker-compose.prod.yml
hamza-56: Removed.

docker-compose.prod.yml
hamza-56: Removed the logging section and all logging settings.

docker-compose.prod.yml
hamza-56: Done, every service now lists .env and .env.local explicitly and compose crashes if .env.local is missing.

docker-compose.prod.yml
hamza-56: Dropped the anchor, the connection URLs now go in .env.local and the sparkth service only sets SERVE_FRONTEND.

docker-compose.prod.yml
hamza-56: Removed the auto-run job and collapsed the three commands into a single `sparkth migrate` CLI command, documented in the README.

Makefile
hamza-56: Dropped the make targets, the README now uses plain docker compose commands.

hamza-56:
@claude-review

regisb (approved):
good to go!

PR #566, opened 2026-07-31
feat(plugins): backend-declared plugin names and frontend metadata

### LINKED ISSUE #565: Plugin identity and display metadata are split across disconnected backend and frontend registries
## Summary

Item 4 of #372. A plugin's identity and display metadata are split across two disconnected registries, one in the backend and one in the frontend, tied together only by a matching kebab-case name string. Nothing declares or validates that name: the backend derives it from the Python class name, and the frontend repeats it by hand. There is no single source of truth for what a plugin is.

## Where it happens

- `sparkth/core/plugins/base.py` — `SparkthPlugin` carries nothing frontend-facing: no display name, no description, no icon, no nav entry.
- `frontend/lib/plugins/registry.ts`, `frontend/lib/plugins/types.ts` — the frontend maintains a completely independent `PluginDefinition` registry (`name`, `displayName`, `loadComponent`, routes, events), registered by hand.
- `frontend/lib/plugins/index.ts` and `generateStaticParams` — the plugin name must be repeated manually here as well.

## Why it matters

- The kebab-case name is an implicit, uncheckable convention. The frontend can register a plugin the backend has never heard of, and vice versa, with no error anywhere.
- Renaming a backend plugin class silently changes the plugin's identity (this has already required two cleanup migrations).
- The registries are already drifting: google-drive is missing from the static params list, and backend-only plugins like canvas and open-edx show up in settings with a blank description because display metadata only exists in the frontend registry.

## How to observe

1. Enable a backend plugin that has no matching frontend `PluginDefinition` — no UI appears and nothing reports the mismatch.
2. Open the settings page — canvas and open-edx render with empty descriptions.
3. Compare the plugin names in `frontend/lib/plugins/index.ts` and `generateStaticParams` against the backend loader output — google-drive is absent from the static params.

hamza-56:

First PR of the sequence proposed in #372 (hooks + API exposure). Frontend consumption and the drift gate follow separately, so this does not close #565 yet.

## What

The backend becomes the single source of truth for a plugin's identity and display metadata: plugins declare their own name explicitly (nothing is derived from the class name anymore) and contribute display info, sidebar entry, and a frontend-app marker through per-concern hooks, which the user-plugins API exposes read-only.

## Changes

- feat(plugins): plugins declare their own kebab-case name positionally via `super().__init__()`; the loader constructs classes with no arguments and stops deriving names from class names
- feat(plugins): add `DISPLAY_INFO`, `SIDEBAR_ENTRIES`, and `FRONTEND_APPS` hooks (`sparkth/lib/frontend`) with name-based lookup helpers
- feat(plugins): built-in plugins register their display info (canvas and open-edx finally get one), sidebar entries (chat, slack), and frontend-app markers (chat, google-drive, slack)
- feat(api): expose `display`, `sidebar`, and `has_frontend` read-only on `UserPluginResponse`; regenerate `frontend/lib/api/generated.ts`
- test(plugins): loader explicit-name contract, hook lookup helpers, per-plugin identity tests, API response coverage
- docs(plugins): replace the "Plugin Name Derivation" section with explicit declaration, document the frontend metadata hooks

## How to Test

1. `uv run pytest` (new suites: `tests/lib/test_frontend.py`, `tests/core/plugins/test_loader.py`, `sparkth/plugins/*/tests/test_plugin.py`)
2. `make backend.up.dev`, authenticate, then `GET /api/v1/user-plugins/`: every plugin entry carries `display`, `sidebar`, and `has_frontend`; canvas and open-edx now have a display name and description
3. `make docs`: the Plugin authoring reference renders the new `sparkth.lib.frontend` section

## Notes

- **Breaking for plugin authors**: a plugin `__init__` no longer receives the name from the loader. Declare it explicitly: `def __init__(self) -> None: super().__init__("my-app")`. Old-style plugins are skipped at load time with an error in the logs.
- No migration needed: all five built-in plugins declare the exact names previously derived (`canvas`, `chat`, `google-drive`, `open-edx`, `slack`), so existing DB rows keep matching.
- Plugin names are now validated as kebab-case slugs; `super().__init__()` raises `ValueError` otherwise.

### DIFF
--- a/.claude/docs/architectural_patterns.md
+++ b/.claude/docs/architectural_patterns.md
@@ -14,13 +14,20 @@ a `PluginHook` holds one; `sparkth/lib/hooks.py` also defines a `SingleNamedItem
 a flat, name-keyed set not grouped by plugin — used for the permission/scope vocabulary):
 
 ```
-SparkthPlugin.__init__
+SparkthPlugin.__init__          # super().__init__("my-app") declares the plugin name
   ├── register_router(self, router)                                   # sparkth/lib/routes.py
   ├── MCP_TOOLS.add_item(self, Tool(self.my_tool, category="..."))   # sparkth/lib/mcp/hooks.py
   ├── CONFIG_SCHEMAS.add_item(self, MyPluginConfig)                  # sparkth/lib/config/hooks.py
   ├── CONFIG_ADAPTERS.add_item(self, MyConfigAdapter())              # sparkth/lib/config/hooks.py
+  ├── DISPLAY_INFO.add_item(self, DisplayInfo("My App", "..."))      # sparkth/lib/frontend/hooks.py
+  ├── SIDEBAR_ENTRIES.add_item(self, SidebarEntry("My App", ...))    # sparkth/lib/frontend/hooks.py
+  ├── FRONTEND_APPS.add_item(self, FrontendApp())                    # sparkth/lib/frontend/hooks.py
   └── register_event_schema(self, MyEvent)                           # sparkth/lib/analytics
 ```
+
+Every plugin declares its own kebab-case name by passing it positionally to
+`super().__init__()`; the loader (`sparkth/core/plugins/loader.py`) constructs each class
+with no arguments and never derives a name from the class name.
 
 `register_router` (`sparkth/lib/routes.py`) mounts the router at `/api/v1/<plugin-name>`,
 deriving both the prefix and OpenAPI tags from the plugin instance automatically.
--- a/docs/guides/plugins.md
+++ b/docs/guides/plugins.md
@@ -77,7 +77,7 @@ Both methods default to `None` on the base class, so non-LMS plugins require no 
 
 Register a Pydantic configuration class when your plugin has user-configurable settings that the system should validate and normalize — and always for LMS plugins that rely on credential injection. Plugins with no user-facing configuration can skip this entirely.
 
-When you do register one, contribute your config class to the `CONFIG_SCHEMAS` hook from your plugin's `__init__`, right after calling `super().__init__(...)`. The system resolves config classes by the plugin's _derived_ name (the value passed to `__init__` — see [Plugin Name Derivation](#plugin-name-derivation) below), so no name string is needed at the call site.
+When you do register one, contribute your config class to the `CONFIG_SCHEMAS` hook from your plugin's `__init__`, right after calling `super().__init__(...)`. The system resolves config classes by the plugin's _declared_ name (the value you pass to `super().__init__()`, see [Plugin Names](#plugin-names) below), so no name string is needed at the call site.
 
 ```python
 # sparkth/plugins/myappplugin/plugin.py
@@ -88,8 +88,8 @@ from sparkth.lib.plugins import SparkthPlugin
 
 
 class MyAppPlugin(SparkthPlugin):
-    def __init__(self, plugin_name: str) -> None:
-        super().__init__(plugin_name)
+    def __init__(self) -> None:
+        super().__init__("my-app")
         CONFIG_SCHEMAS.add_item(self, MyAppPluginConfig)
 ```
 
@@ -134,8 +134,8 @@ from sparkth.lib.plugins import SparkthPlugin
 
 
 class MyAppPlugin(SparkthPlugin):
-    def __init__(self, plugin_name: str) -> None:
-        super().__init__(plugin_name)
+    def __init__(self) -> None:
+        super().__init__("my-app")
         CONFIG_SCHEMAS.add_item(self, MyAppPluginConfig)
         CONFIG_ADAPTERS.add_item(self, MyAppPluginConfigAdapter())
 ```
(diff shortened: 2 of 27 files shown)


hamza-56:
@claude-review

.claude/docs/architectural_patterns.md
```diff
@@ -14,14 +14,21 @@ a `PluginHook` holds one; `sparkth/lib/hooks.py` also defines a `SingleNamedItem
...
   ├── CONFIG_ADAPTERS.add_item(self, MyConfigAdapter())              # sparkth/lib/config/hooks.py
+  ├── DISPLAY_INFO.add_item(self, DisplayInfo("My App", "..."))      # sparkth/lib/frontend/hooks.py
+  ├── SIDEBAR_ENTRIES.add_item(self, SidebarEntry("My App", ...))    # sparkth/lib/frontend/hooks.py
+  ├── FRONTEND_APPS.add_item(self, FrontendApp())                    # sparkth/lib/frontend/hooks.py
   └── register_event_schema(self, MyEvent)                           # sparkth/lib/analytics
 ```
 
+Every plugin declares its own kebab-case name by passing it positionally to
+`super().__init__()`; the loader (`sparkth/core/plugins/loader.py`) constructs each class
+with no arguments and never derives a name from the class name.
```
regisb, line 30: Oh right, this had escaped me. Please ignore my comments from the standup meeting earlier today.

docs/guides/plugins.md
```diff
@@ -280,10 +279,41 @@ Analytics event schemas register through `register_event_schema(self, MyEvent)`:
...
+
+class MyAppPlugin(SparkthPlugin):
+    def __init__(self) -> None:
+        super().__init__("my-app")
+        DISPLAY_INFO.add_item(self, DisplayInfo("My App", "What the plugin does, in one line", icon="sparkles"))
+        SIDEBAR_ENTRIES.add_item(self, SidebarEntry("My App", icon="sparkles", order=5))
+        FRONTEND_APPS.add_item(self, FrontendApp())
+```
+
+- **`DISPLAY_INFO`**: the human-facing identity (display name, description, optional icon) shown in settings and catalogs. Every plugin should register one; without it the plugin appears with only its slug name.
```
regisb: This sort of reference information would be more appropriate as docstring to be included in the generated docs.

sparkth/core/plugins/loader.py
```diff
@@ -52,36 +53,58 @@ def _load_all(self) -> None:
...
             Check logs for any failures.
         """
-        for plugin_name, plugin_class in self.iter_plugin_classes():
+        for plugin_class in self.iter_plugin_classes():
             try:
-                plugin_instance = plugin_class(plugin_name)
+                # Concrete plugin classes declare a no-argument __init__ that
+                # passes their name to super().__init__(); only the base class
+                # takes the name parameter, hence the call-arg ignore.
+                plugin_instance = plugin_class()  # type: ignore[call-arg]
```
regisb, line 58: Any reason why we need `type: ignore`? Also the comment is redundant.

sparkth/lib/frontend/hooks.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@dataclass(frozen=True)
+class FrontendApp:
+    """Marks the plugin as shipping a frontend page (``/dashboard/<plugin-name>``).
+
+    Registering it is the declaration. The marker carries no data yet; fields
+    describing the frontend counterpart (entry route, bundle info) belong here
+    as the frontend plugin system grows.
+    """
```
regisb, line 64: 👍

regisb (approved):
Just a few minor comments but this looks great! I like it.

docs/guides/plugins.md
hamza-56: Done. Moved into the `sparkth.lib.frontend.hooks` module docstring, rendered on the plugin authoring reference page; the guide now links to it.

sparkth/core/plugins/loader.py
hamza-56: The reason it is needed: `iter_plugin_classes()` yields `Type[SparkthPlugin]`, whose `__init__` takes `name`, but the loader calls `plugin_class()` because each concrete subclass passes its own name up. mypy flags the missing arg on the base signature, which the subclasses no longer have.

docs/guides/plugins.md
```diff
@@ -207,26 +207,25 @@ class MyAppPluginConfigAdapter(LLMConfigAdapter):
...
+```python
+class MyAppPlugin(SparkthPlugin):
+    def __init__(self) -> None:
+        super().__init__("my-app")
+```
 
-This derived name is what gets passed to your `__init__`, what the `CONFIG_SCHEMAS` hook resolves config classes by, and what `get_plugin_adapter` uses to look up your adapter from `CONFIG_ADAPTERS`. Name your class so the derived name is what you want.
+The name must be a kebab-case slug (lowercase letters, digits, single hyphens: `PLUGIN_NAME_PATTERN` in `sparkth/core/plugins/base.py`); `super().__init__()` raises `ValueError` otherwise. It appears in URLs (`/api/v1/<name>`, `/dashboard/<name>`), keys the plugin's database row, and is what every hook lookup (`CONFIG_SCHEMAS`, `CONFIG_ADAPTERS`, the frontend metadata hooks) resolves by, so treat it as permanent once released.
 
+The loader (`sparkth/core/plugins/loader.py`) constructs every plugin as `plugin_class()` with no arguments and indexes it under the declared name. A plugin that declares a name already taken by another plugin is skipped with an error in the logs.
```
abdulrafey1, line 222: Do we want to skip a duplicate plugin name or would it be better to crash?

sparkth/core/plugins/base.py
```diff
@@ -7,42 +7,59 @@
...
+- frontend metadata via ``DISPLAY_INFO`` / ``SIDEBAR_ENTRIES`` / ``FRONTEND_APPS``
+  (``sparkth.lib.frontend.hooks``)
 """
 
+import re
+
+# A plugin name is a kebab-case slug: it appears in URLs (``/api/v1/<name>``,
+# ``/dashboard/<name>``) and is the key joining the backend plugin, its DB row,
+# and its frontend counterpart.
+PLUGIN_NAME_PATTERN = re.compile(r"[a-z0-9]+(-[a-z0-9]+)*")
```
abdulrafey1: lets move this to `sparkth.core.plugins.constants` to keep things consistent with rest of the codebase

sparkth/lib/frontend/hooks.py
```diff
@@ -0,0 +1,75 @@
...
+@dataclass(frozen=True)
+class SidebarEntry:
+    """A dashboard sidebar navigation entry pointing at the plugin's frontend page.
+
+    ``order`` sorts entries ascending; entries that keep the default sort last.
+    """
+
+    label: str
+    icon: str | None = None
+    order: int = 100
```
abdulrafey1, line 54: If we want the default `order` to be the highest, then lets use `float('inf')` which always returns `True` on a `x < float('inf')` comparison

sparkth/plugins/chat/plugin.py
```diff
@@ -14,8 +22,13 @@
...
-        super().__init__(plugin_name)
+    def __init__(self) -> None:
+        super().__init__("chat")
         register_router(self, chat_router)
         CONFIG_SCHEMAS.add_item(self, ChatUserConfig)
         CONFIG_ADAPTERS.add_item(self, ChatConfigAdapter())
+        DISPLAY_INFO.add_item(
+            self, DisplayInfo("Create Course", "Transform your resources into courses with AI", icon="plus")
+        )
+        SIDEBAR_ENTRIES.add_item(self, SidebarEntry("Create Course", icon="plus", order=1))
```
abdulrafey1, line 33: having these values defined in the config file for the plugin would make more sense, define there, then import and pass here
Same for all other plugins

docs/guides/plugins.md
hamza-56: Intentional, the loader's contract is "one bad plugin never crashes the app" (constructor failures, missing names all skip too). Hooks are weakref-keyed, so the dropped duplicate gets GC'd with everything it registered.

sparkth/core/plugins/base.py
hamza-56: Done in bbde223.

sparkth/lib/frontend/hooks.py
hamza-56: `order` crosses the wire and JSON has no Infinity, so `inf` would break the API response and the generated TS. The default just means "late", not "highest", an explicit order=200 sorting after defaults is fine.

sparkth/plugins/chat/plugin.py
hamza-56: `config.py` means env settings + the user-editable `PluginConfig` schema, and display name/icon are static identity, not config, so I'd rather keep them at the registration site like all other hook registrations. What would the indirection buy?

abdulrafey1 (approved):

PR #567, opened 2026-07-31
feat(frontend): render backend-declared plugin metadata

hamza-56:

Second PR of the sequence proposed in #372, stacked on #566 (hooks + API exposure). The drift gate follows in #568.

## What

The frontend stops keeping its own copy of plugin display metadata and renders what the backend declares: `PluginDefinition` shrinks to the plugin name plus component loaders, and display name, description, icon, and sidebar config come from the user-plugins API.

## Changes

- feat(frontend): `PluginDefinition` shrinks to what only the frontend can own (name, `loadComponent`, `loadSettingsComponent`); `displayName`, `description`, `icon`, `category`, and the sidebar fields are removed
- feat(frontend): `UserPluginState` carries `display`, `sidebar`, and `has_frontend`, typed from the generated OpenAPI schema
- feat(frontend): sidebar, dashboard cards, settings list, and plugin pages read the backend-declared metadata; backend-only plugins (canvas, open-edx) now show a proper name and description in settings
- feat(frontend): icons resolve from declared lucide names via `lib/plugins/icons.ts`; plugins register custom brand icons under their name (slack)
- fix(frontend): add google-drive to `generateStaticParams` (was missing)
- test(frontend): metadata helpers, icon resolution, and settings list rendering

## How to Test

1. `cd frontend && bun run test` (new suites: `lib/tests/plugin-metadata.test.ts`, `lib/tests/plugin-icons.test.ts`, `components/settings/ListItem.test.tsx`)
2. Run the app, open the dashboard: sidebar entries (Create Course, Slack TA Bot) and cards come from the API metadata
3. Open Settings > My Plugins: canvas and open-edx show display names and descriptions instead of slugs and blanks

## Notes

- Stacked on #566 (requires the backend hooks and API fields).
- Removed frontend registry APIs that existed only for the frontend-side metadata copy: `getSidebarPlugins`, `getPluginsByCategory`, `getCorePlugins`, `searchPlugins`, `useSidebarPlugins` (all unused outside the registry).
- No migrations, no env vars, no dependency changes.

### DIFF
--- a/frontend/app/dashboard/[pluginName]/page-client.tsx
+++ b/frontend/app/dashboard/[pluginName]/page-client.tsx
@@ -3,8 +3,8 @@
 import { useParams, redirect } from "next/navigation";
 import { useAuth } from "@/lib/auth-context";
 import { Spinner } from "@/components/Spinner";
-import { usePlugin } from "@/lib/plugins/context";
-import { getPlugin } from "@/lib/plugins";
+import { usePlugin, usePluginContext } from "@/lib/plugins/context";
+import { displayNameOf, getPlugin } from "@/lib/plugins";
 import PluginRenderer from "@/components/PluginRenderer";
 import Link from "next/link";
 
@@ -15,7 +15,11 @@ export default function PluginPageClient() {
   const pluginName = params?.pluginName as string;
 
   const { isEnabled } = usePlugin(pluginName);
+  const { userPlugins } = usePluginContext();
   const pluginDef = getPlugin(pluginName);
+
+  const userPlugin = userPlugins.find((p) => p.plugin_name === pluginName);
+  const displayName = userPlugin ? displayNameOf(userPlugin) : pluginName;
 
   if (!authLoading && !isAuthenticated) {
     redirect(`/login?redirect=/dashboard/${pluginName}`);
@@ -125,7 +129,7 @@ export default function PluginPageClient() {
           </div>
           <h2 className="text-2xl font-bold text-foreground mb-2">Plugin Not Enabled</h2>
           <p className="text-muted-foreground mb-4">
-            {pluginDef.displayName} is not enabled for your account.
+            {displayName} is not enabled for your account.
           </p>
           <div className="flex justify-center gap-3">
             <Link
--- a/frontend/app/dashboard/[pluginName]/page-client.tsx
+++ b/frontend/app/dashboard/[pluginName]/page-client.tsx
@@ -15,7 +15,7 @@ export default function PluginPageClient() {
   const pluginName = params?.pluginName as string;
 
   const { isEnabled } = usePlugin(pluginName);
-  const { userPlugins } = usePluginContext();
+  const { userPlugins, loading: pluginsLoading } = usePluginContext();
   const pluginDef = getPlugin(pluginName);
 
   const userPlugin = userPlugins.find((p) => p.plugin_name === pluginName);
@@ -103,6 +103,16 @@ export default function PluginPageClient() {
             Go to Dashboard
           </Link>
         </div>
+      </div>
+    );
+  }
+
+  // Until the plugin list loads, neither isEnabled nor displayName is
+  // meaningful, so don't render a verdict yet.
+  if (pluginsLoading) {
+    return (
+      <div className="flex items-center justify-center h-full min-h-screen bg-background">
+        <Spinner />
       </div>
     );
   }
--- a/frontend/app/dashboard/[pluginName]/page.tsx
+++ b/frontend/app/dashboard/[pluginName]/page.tsx
@@ -1,7 +1,7 @@
 import PluginPageClient from "./page-client";
 
 export function generateStaticParams() {
-  return [{ pluginName: "chat" }, { pluginName: "slack" }];
+  return [{ pluginName: "chat" }, { pluginName: "google-drive" }, { pluginName: "slack" }];
 }
 
 export default function PluginPage() {
--- a/frontend/app/dashboard/[pluginName]/page.tsx
+++ b/frontend/app/dashboard/[pluginName]/page.tsx
@@ -1,7 +1,10 @@
+// The barrel, not @/lib/plugins/registry — the registerPlugin() calls live in
+// the barrel, so importing the leaf would return an empty registry.
+import { getAllPlugins } from "@/lib/plugins";
 import PluginPageClient from "./page-client";
 
 export function generateStaticParams() {
-  return [{ pluginName: "chat" }, { pluginName: "google-drive" }, { pluginName: "slack" }];
+  return getAllPlugins().map((plugin) => ({ pluginName: plugin.name }));
 }
 
 export default function PluginPage() {
--- a/frontend/app/dashboard/page.tsx
+++ b/frontend/app/dashboard/page.tsx
@@ -1,8 +1,7 @@
 "use client";
 
 import { usePluginContext } from "@/lib/plugins/context";
-import { displayNameOf } from "@/lib/plugins/metadata";
-import { resolvePluginIcon } from "@/lib/plugins/icons";
+import { displayNameOf, resolvePluginIcon } from "@/lib/plugins";
 import Link from "next/link";
 import { ArrowRight } from "lucide-react";
 import { Spinner } from "@/components/Spinner";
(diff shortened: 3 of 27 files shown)


hamza-56:
@claude-review

frontend/lib/plugins/types.ts
```diff
@@ -170,13 +154,29 @@ export interface EnabledPlugin {
...
  */
 export interface UserPluginState {
   plugin_name: string;
   enabled: boolean;
   config: Record<string, string>;
   is_core: boolean;
+  display?: PluginDisplayInfo | null;
+  sidebar?: PluginSidebarEntry | null;
+  has_frontend?: boolean;
 }
```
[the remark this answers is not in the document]
hamza-56: fixed

frontend/app/dashboard/[pluginName]/page-client.tsx
```diff
@@ -15,8 +15,12 @@ export default function PluginPageClient() {
   const pluginName = params?.pluginName as string;
 
   const { isEnabled } = usePlugin(pluginName);
+  const { userPlugins } = usePluginContext();
   const pluginDef = getPlugin(pluginName);
 
+  const userPlugin = userPlugins.find((p) => p.plugin_name === pluginName);
+  const displayName = userPlugin ? displayNameOf(userPlugin) : pluginName;
```
Comment (abdulrafey1) [inline frontend/app/dashboard/[pluginName]/page-client.tsx:22]: `displayName` falls back to the slug whenever `userPlugins.find(...)` misses — and that includes every render before the plugin list loads, not just an unknown plugin. Since `isEnabled` is also false in that window, the "Plugin Not Enabled" screen renders with the slug in it.

Both components already destructure loading off the context, so the cheap fix is to not render a verdict yet:
```
const { userPlugins, loading } = usePluginContext();
// ...
if (loading) return <Spinner />;   // before the !isEnabled branch
```

Same thing in `PluginRenderer.tsx:35` — its "Plugin Disabled" copy and both `PluginLoadingFallback` labels read from the same fallback.

frontend/lib/plugins/metadata.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const DEFAULT_SIDEBAR_ORDER = 100;
+
+/**
+ * Sidebar items for the enabled plugins that declared a sidebar entry,
+ * sorted by their declared order.
+ */
+export function sidebarItemsFrom(userPlugins: UserPluginState[]): SidebarItem[] {
+  return userPlugins
+    .filter((plugin) => plugin.enabled && plugin.sidebar)
```
abdulrafey1: `sidebarItemsFrom` filters on `enabled && sidebar`, so any plugin the backend declares a `SidebarEntry` for gets a nav link — including one with no frontend registration. Before this PR the list came from `getPluginsByNames(enabledNames)`, so that couldn't happen; the registry lookup was the guard.

This should fix the problem:
```
.filter((plugin) => plugin.enabled && plugin.has_frontend && plugin.sidebar)
```

frontend/app/dashboard/[pluginName]/page.tsx
```diff
@@ -1,7 +1,7 @@
 import PluginPageClient from "./page-client";
 
 export function generateStaticParams() {
-  return [{ pluginName: "chat" }, { pluginName: "slack" }];
+  return [{ pluginName: "chat" }, { pluginName: "google-drive" }, { pluginName: "slack" }];
```
Comment (abdulrafey1) [inline frontend/app/dashboard/[pluginName]/page.tsx]: `output: "export"` in production makes this array the definitive list of plugin pages that exist — which is why `google-drive` 404'd until this PR added it. That's three places a plugin name has to be written before the UI works: the frontend registry, the backend's FRONTEND_APPS, and here.

`getAllPlugins()` is already exported from `registry.ts`, and it's exactly the set this route can render:

```
// The barrel, not @/lib/plugins/registry — the registerPlugin() calls live in
// the barrel, so importing the leaf would return an empty registry.
import { getAllPlugins } from "@/lib/plugins";

export function generateStaticParams() {
  return getAllPlugins().map((plugin) => ({ pluginName: plugin.name }));
}
```

frontend/app/dashboard/page.tsx
```diff
@@ -1,12 +1,15 @@
 "use client";
 
-import { useEnabledPlugins } from "@/lib/plugins/context";
+import { usePluginContext } from "@/lib/plugins/context";
+import { displayNameOf } from "@/lib/plugins/metadata";
```
abdulrafey1: `ListItem.tsx:5`, `PluginRenderer.tsx:5`, and `page-client.tsx:7` import this from `@/lib/plugins`.
Here the import is from `@/lib/plugins/metadata`. Same symbol, two paths, one PR

frontend/lib/plugins/metadata.ts
```diff
@@ -0,0 +1,43 @@
...
+  icon?: string | null;
+  description?: string | null;
+}
+
+/** The backend-declared display name, falling back to the plugin name. */
+export function displayNameOf(plugin: UserPluginState): string {
+  return plugin.display?.display_name ?? plugin.plugin_name;
+}
+
+const DEFAULT_SIDEBAR_ORDER = 100;
```
abdulrafey1: `SidebarEntry.order` is `int = 100` on the backend and non-optional in the generated schema (`order: number`, `@default`) , so the value is always present when `sidebar` is. `?? DEFAULT_SIDEBAR_ORDER` is unreachable because the `.filter()` one line up already excluded `sidebar: null`; the `?.` is only there because TS can't narrow across the `.sort()`. So this is a second home for a default that can't be reached, in the PR whose point is that there should be one home for it.

abdulrafey1 (commented):
Few small comments:

`useEnabledPlugins` and the provider's `enabledPlugins` state are now dead, and they are the frontend-side copy this PR set out to retire. Files: `frontend/lib/plugins/context.tsx:27, 146, 318-325`

frontend/app/dashboard/[pluginName]/page-client.tsx
Comment (hamza-56) [reply frontend/app/dashboard/[pluginName]/page-client.tsx:22]: Fixed in e0ec1b7: both the plugin page and PluginRenderer now destructure loading off the context and render a spinner before any verdict branch, so neither the "Plugin Not Enabled" copy nor the loading-fallback labels can render with the slug.

frontend/lib/plugins/metadata.ts
hamza-56: Fixed in cb8a30d: the filter now requires `enabled && has_frontend && sidebar`, restoring the guard the registry lookup used to provide.

frontend/app/dashboard/[pluginName]/page.tsx
Comment (hamza-56) [reply frontend/app/dashboard/[pluginName]/page.tsx]: Fixed in d4c9a31: `generateStaticParams` now maps over `getAllPlugins()` from the barrel (with the barrel-vs-leaf caveat noted in a comment), so the frontend registry is the single frontend-side place a plugin name is written.

frontend/app/dashboard/page.tsx
hamza-56: Fixed in 3cbb82e: the dashboard page now imports `displayNameOf` and `resolvePluginIcon` from `@/lib/plugins` like the other consumers.

frontend/lib/plugins/metadata.ts
hamza-56: Fixed in cb8a30d: the filter is now a type guard that narrows `sidebar` as non-null through the sort, so both the `?? DEFAULT_SIDEBAR_ORDER` fallback and the `!` assertions are gone.

hamza-56:
Addressed @abdulrafey1's review comment about `useEnabledPlugins`: the hook and the provider's `enabledPlugins` state are removed in c214408 (their only remaining reference was a stale mock in `layout.test.tsx`). The inline comments are fixed in cb8a30d, e0ec1b7, d4c9a31, and 3cbb82e. Type check, lint, and the full frontend suite (231 tests) pass.

frontend/lib/plugins/icons.ts
```diff
@@ -0,0 +1,28 @@
...
+ * declarations are mapped explicitly (keeping the bundle tree-shakeable), and
+ * plugins can register custom (non-lucide) icon components under a name.
+ */
+
+import type { ComponentType } from "react";
+import { Plus } from "lucide-react";
+
+export type PluginIcon = ComponentType<{ className?: string }>;
+
+const icons = new Map<string, PluginIcon>([["plus", Plus]]);
```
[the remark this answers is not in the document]
hamza-56: Fixed in 0425dad: the resolver now logs a dev-only `console.warn` naming the missing icon and where to register it, and the frontend plugin guide documents the "add the lucide name to `icons.ts`" step.

abdulrafey1 (approved):

PR #568, opened 2026-07-31
feat(plugins): gate plugin registries against drift

### LINKED ISSUE #565: Plugin identity and display metadata are split across disconnected backend and frontend registries

hamza-56:

Final PR of the sequence proposed in #372, stacked on #567. Together with #566 and #567 this closes #565.

## What

A mismatch between the backend's plugin declarations and the frontend registry becomes a red build instead of a silently missing dashboard page: the frontend's plugin route list is generated from the backend `FRONTEND_APPS` declarations, and tests on both tiers fail on drift.

## Changes

- feat(plugins): `scripts/dump_frontend_plugins.py` dumps the `FRONTEND_APPS` declarations offline (the loader is DB-free) as `frontend/lib/plugins/generated.ts`
- feat(frontend): `generateStaticParams` derives the dashboard plugin routes from the generated list instead of a hand-maintained array
- chore: `make frontend.build.plugins` regenerates the committed list
- test(plugins): pytest fails when the committed list is stale vs the declarations (`tests/core/test_dump_frontend_plugins.py`)
- test(frontend): vitest fails when the frontend registry does not match the list (`lib/tests/plugin-registry-drift.test.ts`)
- docs: plugin guides describe the regeneration workflow

## How to Test

1. `uv run pytest tests/core/test_dump_frontend_plugins.py`
2. `cd frontend && bun run test lib/tests/plugin-registry-drift.test.ts`
3. Remove `FRONTEND_APPS.add_item(...)` from a plugin without regenerating: step 1 fails. Regenerate with `make frontend.build.plugins`: step 2 fails until the frontend definition is removed too.

## Notes

- Stacked on #567 (requires the shrunken `PluginDefinition` and the metadata hooks from #566).
- No CI workflow changes needed: both gates run inside the existing `make test.backend` / `make test.frontend` lanes.
- No migrations, no env vars, no dependency changes.


hamza-56:
@claude-review

frontend/lib/plugins/generated.ts
```diff
@@ -0,0 +1,7 @@
+// Generated by scripts/dump_frontend_plugins.py. Do not edit.
```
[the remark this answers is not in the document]
hamza-56: fixed

Makefile
```diff
@@ -84,6 +84,10 @@ frontend.build.api: ## Regenerate frontend API types from the backend OpenAPI sc
 	uv run python scripts/dump_openapi.py > openapi.json
 	cd frontend && bunx openapi-typescript ../openapi.json -o ./lib/api/generated.ts
 
+.PHONY: frontend.build.plugins
+frontend.build.plugins: ## Regenerate the frontend plugin list from the backend FRONTEND_APPS declarations
+	uv run python scripts/dump_frontend_plugins.py > frontend/lib/plugins/generated.ts
```
regisb: Could we maybe run this as `uv run python scripts/dump_frontend_plugins.py -o frontend/lib/plugins/generated.ts`? And if there is no `-o` argument print to stdout. This should make the script easier to debug.

frontend/lib/plugins/generated.ts
```diff
@@ -0,0 +1,7 @@
...
+// Backend plugins that declare a frontend page (FRONTEND_APPS hook).
+// Regenerate with `make frontend.build.plugins`; tests fail on drift.
+
+export const FRONTEND_PLUGIN_NAMES = ["chat", "google-drive", "slack"] as const;
```
regisb, line 5: This is effectively a leak of plugins into the core. This means that when a user wants to enable some custom plugins, they will need to re-generate this file, which is not really appropriate. I don't have a great solution though. Could we instead generate one file per plugin, in their own plugin directory?

Makefile
hamza-56: Done in 5c392e7

I left `scripts/dump_openapi.py` on redirection for now to keep this PR scoped; happy to align the two dump scripts in a follow-up.

abdulrafey1 (approved):
Approving this for now, but not sure how we want to stop the plugins leak into the core.
We can have this implementation for now but we do need to address what regis has pointed out

PR #571, opened 2026-07-31
refactor(frontend): centralize unit tests and name test dirs by kind

zaira-bibi:
## What
Give the frontend one home per test kind, named for what it holds: all vitest unit tests move into a single `frontend/tests/` mirror of the source tree, and the Playwright suite moves to `frontend/e2e-tests/`, so a developer can tell which suite lives where from the directory name alone.

## Changes
- test(frontend): move co-located component/page tests into a central mirror at `<same path>` and rewrite their code-under-test imports to the `@/` alias
- test(frontend): fold the `lib/tests/` suite (incl. `test-utils.ts`) into the mirror's `lib/`
- refactor(frontend): rename the unit mirror `__tests__/` → `tests/`, and the Playwright suite `tests/` → `e2e-tests/`
- fix(frontend): repoint `vitest.config.ts`'s exclude at `e2e-tests/**` and `playwright.config.ts`'s `testDir` at `./e2e-tests`
- docs: document the two-directory convention in CLAUDE.md, and correct the e2e path in README.md and `scripts/e2e_seed.py`

## How to Test
1. `cd frontend && bun run test` — 25 files / 212 tests pass, all discovered from `tests/`
2. `cd frontend && bunx playwright test --list` — 12 tests in 4 files, all discovered from `e2e-tests/`. This is the check that matters most: it proves the two suites resolve to different directories and neither runner picks up the other's files.
3. `cd frontend && bun run typecheck` — clean (relocated tests resolve the code under test via `@/`)
4. `cd frontend && bun run lint && bun run format:check` — oxlint 0 warnings/0 errors, oxfmt clean
5. Confirm nothing lives beside the source anymore: `cd frontend && find . -name node_modules -prune -o -name '*.test.*' -print` — every path is under `./tests/`

## Notes
Pure test relocation — no source or behavior changes, no new dependencies, no migrations.

**The one thing worth reviewing closely:** `frontend/vitest.config.ts` excluded `"tests/**"` to keep the Playwright specs out of vitest. Left untouched, that pattern would have silently excluded the *entire* unit suite once the mirror became `tests/` — a green run with zero tests, which is the failure mode that looks like success. It now excludes `"e2e-tests/**"`; step 1 above is what proves it (212 tests, not 0).

**No CI workflow changes were needed, and one line deliberately left alone:** `.github/workflows/playwright.yml` has `- tests/**` in its paths filter, which refers to the **backend Python** suite at the repo root, not the frontend directory — changing it would have broken the filter. Its `- frontend/**` entry already matches both directories under any name, and both runners resolve their directories from the two config files above, so `ci.yml` (`make test.frontend`) needed no edit either.


frontend/playwright.config.ts
```diff
@@ -6,14 +6,14 @@ import "dotenv/config";
...
- * `tests/config.ts`.
+ * `e2e-tests/config.ts`.
  *
  * See https://playwright.dev/docs/test-configuration.
  */
 const BASE_URL = process.env.PLAYWRIGHT_BASE_URL ?? "http://localhost:3000";
 
 export default defineConfig({
-  testDir: "./tests",
+  testDir: "./e2e-tests",
```
regisb, line 16: FYI this is how bunx detects e2e tests (AFAIU).

abdulrafey1 (approved):

frontend/vitest.config.ts
```diff
@@ -8,8 +8,9 @@ export default defineConfig({
     environment: "jsdom",
     globals: true,
     setupFiles: ["./vitest.setup.ts"],
-    // Playwright e2e specs live in tests/ and run via `make test.e2e`, not vitest.
-    exclude: [...configDefaults.exclude, "tests/**"],
+    // Vitest unit tests live in tests/; Playwright e2e specs live in e2e-tests/
+    // and run via `make test.e2e`, not vitest.
+    exclude: [...configDefaults.exclude, "e2e-tests/**"],
```
hamza-56: Since this centralizes unit tests under tests/, worth enforcing it with an explicit `include: ["tests/**/*.test.{ts,tsx}"]` rather than relying on convention. Also makes the e2e-tests exclude below redundant.

hamza-56 (commented):
Looks good overall

hamza-56 (approved):

PR #572, opened 2026-08-03
feat(analytics): add emit_event for analytics producers

zaira-bibi:
> **Stack 1 of 4** — Analytics Milestone 2. This is the bottom layer; the three
> chat-instrumentation PRs above it build on the primitive introduced here.

## What

Analytics producers each carried their own session plumbing and `except` ladder around `ingest_event`; `emit_event` now owns session acquisition in one place and nothing else, and the existing login producer is retrofitted onto it.

## Changes

- feat(analytics): add `emit_event` — opens its own analytics session, validates and lands the event, and propagates every failure
- refactor(auth): emit `user.logged_in` through the primitive instead of hand-rolled session scope + `except` ladder
- test(analytics): add contract tests pinning down which failure reaches the caller for each failure source (unknown event type, invalid payload, gateway error, unexpected error, dead analytics DB)
- test(analytics): move the login producer tests' patch targets to `sparkth.lib.analytics`, and assert that an analytics failure propagates out of the login background task
- docs: point producers at `emit_event` in `README.md` and `docs/guides/plugins.md`, and correct the stale swallowing-contract prose in `sparkth/lib/audit` and the `user.logged_in` schema docstring

## How to Test

1. Run the primitive's contract tests — each pins one failure source to the exception it must surface:
   ```bash
   uv run pytest tests/analytics/test_emit_event.py -v
   ```
   Expected: 6 passed. The two payload-failure cases also assert no row was written, proving there is no partial write.
2. Run the login producer tests:
   ```bash
   uv run pytest tests/analytics/test_login_producer.py -v
   ```
   Expected: 4 passed — one happy path that lands a row, three proving an analytics failure is not hidden.
3. Confirm no regressions across the analytics suite and the auth surface:
   ```bash
   uv run pytest tests/analytics/ -v
   uv run pytest tests/ -k "auth or login"
   ```
   Expected: 61 passed / 3 skipped, and 76 passed / 3 skipped. The skips are the `pg`-marked TimescaleDB lane, which skips without `ANALYTICS_TEST_PG_URL`.
4. Confirm nothing in the analytics write path catches broadly:
   ```bash
   grep -rn "except Exception" sparkth/lib/analytics/
   ```
   Expected: no output.

## Notes

- **Error-handling policy — please read.** An earlier revision of this branch had the primitive swallow every failure behind a documented `except Exception`. That was rejected: analytics failures must not be caught and ignored. `emit_event` now catches **nothing** — `UnknownEventTypeError`, `ValidationError` and `SQLAlchemyError` all reach the caller.
- **What that means at runtime.** Every producer emits from a FastAPI background task, which runs *after* the response is flushed. A failed analytics write therefore does not change the user's request outcome in production; it surfaces as an unhandled ASGI exception that uvicorn logs with a traceback. Under the httpx ASGI test transport the background task runs inside the request, so the same failure is re-raised there — which is exactly what the three inverted login tests assert.
- **No migration.** `raw_events` already exists; new event types are rows, not DDL.
- **No breaking change.** `ingest_event` keeps its signature and behaviour; it remains the right call for a caller that already holds an analytics session.


zaira-bibi:
@claude-review

abdulrafey1 (approved):
looks great

hamza-56 (approved):

PR #573, opened 2026-08-03
refactor(frontend): derive PaginatedResponse from the generated schema

hamza-56:
## What
`PaginatedResponse<T>` in `frontend/lib/drive.ts` hand-duplicated the envelope of the generated `PaginatedResponse_*` schemas field for field, the same latent-drift pattern `UserPluginState` had before #567; the page shapes are now direct aliases of the generated schemas.

## Changes
- refactor(frontend): replace the hand-written `PaginatedResponse<T>` with `PaginatedFolders` / `PaginatedFiles`, direct `Schema<...>` aliases like every other type in the module (openapi-typescript emits one monomorphic paginated schema per item type, so there is no generic to preserve)
- refactor(frontend): make `fetchAllPages`'s fetcher parameter structural (just the `items`/`total` fields the loop reads), keeping it polymorphic without a named envelope type

## How to Test
1. `cd frontend && bun run typecheck` passes.
2. `make test.frontend` passes (25 files, 212 tests).
3. `listFolders` / `listFiles` return types are now exactly the generated schemas, so any backend change to the pagination envelope fails `tsc` after `make frontend.build.api` with no extra gate needed.

## Notes
Type-level only, no runtime change. No migration, no env vars, no dependencies.

### DIFF
--- a/frontend/lib/drive.ts
+++ b/frontend/lib/drive.ts
@@ -10,12 +10,16 @@ export type SyncStatus = Schema<"SyncStatusResponse">;
 export type FileRagStatus = Schema<"FileRagStatusResponse">;
 export type FolderRagStatusResponse = Schema<"FolderRagStatusResponse">;
 
-export interface PaginatedResponse<T> {
-  items: T[];
-  total: number;
-  skip: number;
-  limit: number;
-}
+// The pagination envelope, derived from a generated schema rather than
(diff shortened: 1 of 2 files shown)


hamza-56:
@claude-review

abdulrafey1 (approved):

PR #574, opened 2026-08-04
feat(plugins): add the chat analytics event vocabulary

zaira-bibi:
> **Stack 2 of 4** — Analytics Milestone 2. Based on
> [#572](https://github.com/edly-io/sparkth/pull/572), which adds the `emit_event` primitive
> this PR's helpers call. Review #572 first.

## What
Adds the chat plugin's analytics vocabulary — four `chat.*` event schemas, a tool-name normaliser, a tool-category lookup, and four emit helpers — with no route wired up yet, so the whole thing is additive and independently reviewable.

## Changes
- feat(plugins): add four `AnalyticsEventSchema` subclasses (`chat.conversation_started`, `chat.message_sent`, `chat.completion_served`, `chat.tool_invoked`) plus the frozen `CompletionAnalyticsContext`, registered from `ChatPlugin.__init__`
- feat(plugins): add `tool_names()` — one normaliser reading tool executions from both completion paths' differently-shaped records
- feat(plugins): record MCP tool categories during registry discovery and expose `ToolRegistry.category_for()`, defaulting to `UNKNOWN_TOOL_CATEGORY`
- feat(plugins): add four `emit_*` helpers, each constructing its schema and handing it to
  `emit_event` through a shared `_emit` — the payload is dumped from the instance, so a field
  added to a schema cannot be forgotten at the emission site
- refactor(plugins): move `UNKNOWN_TOOL_CATEGORY` into the plugin's existing `constants.py`
- refactor(plugins): drop the unused `refresh_tools` and the `ToolRegistry.reset` it was the
  only non-test caller of

## How to Test
1. Schemas, normaliser, and helpers:
   ```bash
   uv run pytest sparkth/plugins/chat/tests/test_analytics_schemas.py -v
   ```
   Expected: 31 passed.
2. Registry category lookup:
   ```bash
   uv run pytest sparkth/plugins/chat/tests/test_tool_registry.py -v
   ```
3. No regressions in the plugin, and no `DuplicateEventTypeError` from the new registration:
   ```bash
   ```
   Expected: 296 passed.
4. Confirm no import cycle from `analytics.py` now importing `tools.py`:
   ```bash
   uv run python -c "import sparkth.main"
   ```
   Expected: no output, exit 0.

## Notes
- **The non-negative counts move validation to the producer.** `message_length` and
  `tool_call_count` are `NonNegativeInt`, so the schema is now constructed — and can raise
  `ValidationError` — inside the `emit_*` helper rather than inside `emit_event`. Every call
  site in stacks 3/4 queues these through `background_tasks.add_task`, so the raise still
  lands after the response is sent. These helpers must stay queued, never awaited inline in a
  request path.
- **No migration.** New event types are rows in the existing `raw_events` table, not DDL.
- **Payload safety is enforced structurally.** Every payload carries identifiers, lengths, flags and names only. The base schema's `extra="forbid"` rejects anything else, and `tool_names()` reads *only* the name key — never `tool_input` or `output`, which can hold course content and learner-identifying data.
- **`tool_names()` exists because the two completion paths disagree on shape**: the streaming path's `tool_end` events yield `{"name": …}`, while the non-streaming provider records `{"tool", "tool_input", "output"}`. Keeping that mapping in one place is what stops the two seams in stack 3/4 from drifting apart.
- **Nothing here catches exceptions**, matching `emit_event`'s contract from #572 — a failed analytics write propagates rather than being hidden.
- **No route is wired yet.** Nothing in this PR emits during a real request; that starts in stack 3.


zaira-bibi:
@claude-review

sparkth/plugins/chat/analytics.py
```diff
@@ -0,0 +1,196 @@
...
+
+    Grouping by ``tool_name`` answers what was authored; by ``tool_category``,
+    which LMS. Counts attempted executions, so this is not proof the authoring
+    action succeeded.
+    """
+
+    event_type = "chat.tool_invoked"
+    version = 1
+
+    conversation_id: str
```
abdulrafey1, line 74: `conversation_id` comes up in every schema, could we create a parent class for this?

sparkth/plugins/chat/analytics.py
```diff
@@ -0,0 +1,196 @@
...
+                break
+    return names
+
+
+async def emit_conversation_started(conversation_id: str, provider: str, model: str, actor_id: str) -> None:
+    """Emit ``chat.conversation_started``."""
+    await emit_event(
+        ChatConversationStarted.event_type,
+        ChatConversationStarted.version,
+        {"conversation_id": conversation_id, "provider": provider, "model": model},
```
abdulrafey1, line 127: lets indent this dict just like its indented below

sparkth/plugins/chat/tools.py
```diff
@@ -12,12 +12,17 @@
 
 logger = get_logger(__name__)
 
+# Category reported for a tool the registry never discovered. A missing category
+# degrades one analytics dimension; it must never break emission.
+UNKNOWN_TOOL_CATEGORY = "unknown"
```
abdulrafey1, line 17: lets move this to the existing `chat/constants.py`

abdulrafey1 (commented):
just a few comments

hamza-56 (approved):

sparkth/plugins/chat/analytics.py
```diff
@@ -0,0 +1,196 @@
...
+class ChatMessageSent(AnalyticsEventSchema):
+    """An instructor turn was persisted. ``message_length`` is a character count."""
+
+    event_type = "chat.message_sent"
+    version = 1
+
+    conversation_id: str
+    provider: str
+    model: str
+    message_length: int
```
hamza-56, line 41: `message_length` and `tool_call_count` are plain ints, so a producer bug (e.g. -1) validates fine and skews aggregates. Worth using `NonNegativeInt` / `Field(ge=0)` here.

sparkth/plugins/chat/analytics.py
```diff
@@ -0,0 +1,196 @@
...
+    and learner-identifying data, and must never reach an analytics payload.
+    """
+    names: list[str] = []
+    for record in records:
+        for key in _TOOL_NAME_KEYS:
+            value = record.get(key)
+            if isinstance(value, str) and value:
+                names.append(value)
+                break
+    return names
```
hamza-56, line 119: Records with no usable name are dropped silently. If either provider ever renames the key, this zeroes tool analytics with no trace. Worth a `logger.warning` on the miss (keys only, not values).

sparkth/plugins/chat/analytics.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+async def emit_conversation_started(conversation_id: str, provider: str, model: str, actor_id: str) -> None:
```
hamza-56, line 122: Each emit_* helper hand-lists its schema's fields in a dict, so a field added to the schema but forgotten here only fails at runtime. Building the payload from the schema instance (`event.model_dump(mode="json")`) would catch that at type-check time and collapse these four helpers into one.

sparkth/plugins/chat/tests/test_analytics_schemas.py
```diff
@@ -0,0 +1,207 @@
...
+    )
+    assert event.tool_name == "openedx_create_xblock"
+    assert event.tool_category == "openedx-course"
+
+
+@pytest.mark.parametrize("schema", ALL_SCHEMAS, ids=lambda s: s.event_type)
+def test_schemas_forbid_extra_fields(schema: type[AnalyticsEventSchema]) -> None:
+    """extra="forbid" on the base class stops a producer smuggling content into a payload."""
+    with pytest.raises(ValidationError):
+        schema(unexpected_field="leak")  # type: ignore[call-arg]
```
hamza-56, line 93: This passes for the wrong reason: every schema has required fields, so this raises ValidationError on the missing fields even without extra="forbid". It'd stay green if the smuggling guard were removed. Use a fully valid payload plus one extra key, and assert on extra_forbidden.

sparkth/plugins/chat/tests/test_analytics_schemas.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def test_completion_context_is_frozen() -> None:
+    context = CompletionAnalyticsContext(provider="openai", model="gpt-4o", rag_used=True, actor_id="7")
+    with pytest.raises(Exception):
```
hamza-56, line 98: `pytest.raises(Exception)` is too broad here, it'd pass on an unrelated bug too. A frozen dataclass raises FrozenInstanceError specifically, assert that.

sparkth/plugins/chat/analytics.py
zaira-bibi: the natural approach fails at import time:

```
class ChatEvent(AnalyticsEventSchema):      # no event_type/version
    conversation_id: str
# TypeError: ChatEvent must declare ['event_type', 'version'] as class attributes
```
as `core/analytics/schemas/base.py` enforces that every subclass declares its own identity. if we give it dummy `event_type/version` then a concrete schema that forgot to declare its own would silently inherit them and register under the wrong type, defeating that guard.

if we use a plain `BaseModel` and do something like:

```
class ConversationScoped(BaseModel):
    conversation_id: str

class ChatToolInvoked(ConversationScoped, AnalyticsEventSchema):
    event_type = "chat.tool_invoked"
    version = 1
    tool_name: str
    tool_category: str
```
 this works but the DRY win is small (as it's one str field across four classes), and it also costs a non-obvious trick that exists purely to dodge a safety check. i'd rather skip this and keep the current version.

sparkth/plugins/chat/tools.py
```diff
@@ -56,6 +61,17 @@ async def get_tool_schemas(self) -> list[dict[str, Any]]:
...
         ]
 
+    def category_for(self, tool_name: str) -> str:
+        """Return the MCP category a tool was contributed under.
+
+        LangChain tools carry no category — ``_convert_mcp_to_langchain_tool`` drops
+        it — so the mapping is recorded during discovery. Unknown names resolve to
+        :data:`UNKNOWN_TOOL_CATEGORY` rather than raising.
+        """
+        if not self._initialized:
```
regisb, line 71: This pattern is being repeated multiple times across the ToolRegistry class. I would recommend writing instead:

```
class ToolRegistry:
    """Registry for managing LangChain tools."""

    def __init__(self) -> None:
        self._tools: dict[str, BaseTool] | None = None

    def tools_dict(self) -> dict[str, BaseTool]:
        if self._tools is None:
            self._tools = self._discover_plugin_tools()
        return self._tools
```

Interestingly, while writing this, I realised that we could replace many instance methods by static methods: _convert_mcp_to_langchain_tool, _get_handler_type_hints, _build_args_schema_from_handler, _json_schema_to_pydantic, _resolve_ref, _get_python_type etc. could all be moved out of the ToolRegistry class and made normal functions.

Also, `refresh_tools` is unused anywhere, and we should thus gete rid of the `reset` method.

Maybe this might be worth having a look if you take on this PR?

sparkth/plugins/chat/tests/test_analytics_schemas.py
```diff
@@ -0,0 +1,207 @@
...
+        tool_category="openedx-course",
+    )
+    assert event.tool_name == "openedx_create_xblock"
+    assert event.tool_category == "openedx-course"
+
+
+@pytest.mark.parametrize("schema", ALL_SCHEMAS, ids=lambda s: s.event_type)
+def test_schemas_forbid_extra_fields(schema: type[AnalyticsEventSchema]) -> None:
+    """extra="forbid" on the base class stops a producer smuggling content into a payload."""
+    with pytest.raises(ValidationError):
```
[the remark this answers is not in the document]
abdulrafey1: Done. `test_completion_context_is_frozen` now asserts `dataclasses.FrozenInstanceError` instead of bare `Exception`, so an unrelated failure on the assignment no longer makes the test pass.

sparkth/plugins/chat/tests/test_analytics_schemas.py
abdulrafey1: Done. `test_completion_context_is_frozen` now asserts `dataclasses.FrozenInstanceError` instead of bare `Exception`, so an unrelated failure on the assignment no longer makes the test pass.

sparkth/plugins/chat/tests/test_tool_registry.py
```diff
@@ -273,3 +273,26 @@ async def handler(payload: CreateCoursePayload) -> dict[str, Any]:
...
+    def test_unknown_tool_name_falls_back(self) -> None:
+        registry = ToolRegistry()
+        assert registry.category_for("never_registered_tool") == UNKNOWN_TOOL_CATEGORY
+
+    def test_discovered_tool_reports_its_mcp_category(self) -> None:
+        registry = ToolRegistry()
+        registry.discover_plugin_tools()
+        # Every discovered tool must resolve to some category string, never raise.
+        for name in registry._categories:
+            assert isinstance(registry.category_for(name), str)
```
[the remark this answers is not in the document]
abdulrafey1: Done, and taken further than the suggested guard. Asserting `registry._categories` is non-empty would still tie the test to whatever the process happened to load, so `TestCategoryLookup` now stubs `sparkth.plugins.chat.tools.MCP_TOOLS` and contributes one known `Tool(openedx_create_xblock, category="openedx-course")`, then asserts `category_for("openedx_create_xblock") == "openedx-course"` through the public method. A second test covers a tool contributed with no category falling back to `UNKNOWN_TOOL_CATEGORY`.

That also removes the private-attribute reach you flagged — nothing in the class touches `_categories` any more. `test_reset_clears_categories` is gone along with `reset` itself: `reset` was only reachable from `refresh_tools`, which had zero callers in the repo, so both were dead (see regisb's thread).

sparkth/plugins/chat/analytics.py
abdulrafey1: Agreed, let's keep it as it is. One more argument for leaving it alone beyond what you listed: with `class ChatToolInvoked(ConversationScoped, AnalyticsEventSchema)` a plain `BaseModel` sits ahead of the base in the MRO, and `extra="forbid"` on `AnalyticsEventSchema.model_config` is the whole payload-safety guarantee of this PR. Making that setting depend on a config merge across the MRO to save one `str` field across four classes isn't a trade worth making. Resolving.

sparkth/plugins/chat/analytics.py
abdulrafey1: This one dissolved rather than being fixed directly: addressing hamza's point below, the hand-listed payload dicts are gone from all four helpers — each now constructs its schema instance and passes it to a shared `_emit`. The single-line dict you flagged is now a multi-line constructor call with a trailing comma, matching the three below it (the trailing comma is what keeps `ruff format` from collapsing it back).

sparkth/plugins/chat/tools.py
abdulrafey1: Done. `UNKNOWN_TOOL_CATEGORY` (and its comment) now lives in `sparkth/plugins/chat/constants.py`; `tools.py` and the registry test import it from there, no re-export left behind. Re-ran `uv run python -c "import sparkth.main"` since `constants.py` pulls the provider SDKs and reads the prompt assets at import — clean, exit 0.

sparkth/plugins/chat/analytics.py
abdulrafey1: Done. `ChatMessageSent.message_length` and `ChatCompletionServed.tool_call_count` are both `NonNegativeInt` now, with the reason recorded in each schema's docstring. Two tests assert `greater_than_equal` is the only error raised.

One consequence worth stating explicitly, since it changes *where* the failure lands: the payload is now built by constructing the schema inside the `emit_*` helper (your next comment), so a negative count raises `ValidationError` in the helper's own frame, before `emit_event` is ever awaited. I checked stacks 3 and 4 — every call site is `background_tasks.add_task(emit_*, …)`, so the raise still happens in a background task after the response has been sent, exactly where a `ValidationError` from `emit_event` would have surfaced. Nothing changes for the request.

The constraint that inherits from this: these helpers must keep being queued, not awaited inline in a request path, or an analytics guard would start failing a user's request. A test pins the behaviour — `test_a_negative_count_is_rejected_before_anything_is_emitted` asserts the raise *and* `emit.assert_not_awaited()`, so no skewed row can be written.

sparkth/plugins/chat/analytics.py
abdulrafey1: Done. `tool_names` now logs `logger.warning` per dropped record, and deliberately logs `sorted(record)` — the keys only — since `tool_input`/`output` values are exactly what must not reach a log line any more than a payload. The name lookup moved into a small `_record_name(record) -> str | None` so the drop is a single explicit branch. A test asserts the key appears in the log while the record's values do not.

sparkth/plugins/chat/analytics.py
abdulrafey1: Took the first half of this and not the second. All four helpers now construct their schema instance and hand it to one module-level `_emit(event, actor_id)`, which does `emit_event(event.event_type, event.version, event.model_dump(mode="json"), actor_id=actor_id)`. With `init_typed`/`init_forbid_extra` on the pydantic mypy plugin, a field added to a schema and forgotten at the call site is now a type error, which is the failure you wanted moved earlier — and `event_type`/`version` no longer get re-paired by hand at each site.

I kept the four named helpers rather than collapsing to one, because two of them aren't pure payload assembly: `emit_tool_invoked` resolves the category off the tool registry, and `emit_completion_served` reads the frozen context while taking `streamed` from the seam. Collapsing them would push both back out to the call sites in stacks 3/4, which is what the named helpers exist to prevent.

The four `emit.assert_awaited_once_with(...)` assertions passed unedited, which is the evidence that `model_dump(mode="json")` reproduces the previous payloads byte for byte.

sparkth/plugins/chat/tests/test_analytics_schemas.py
abdulrafey1: Correct, thanks. There's now a `VALID_PAYLOADS` table of one complete payload per schema, and the test does `schema.model_validate({**payload, "unexpected_field": "leak"})` and asserts the error list is exactly `["extra_forbidden"]` — so a missing-field error can no longer carry it, and removing `extra="forbid"` turns it red. (`model_validate` rather than the kwargs form because `init_forbid_extra = true` makes the kwargs form a mypy error, which also let the `# type: ignore[call-arg]` go.)

sparkth/plugins/chat/tools.py
abdulrafey1: Good catch on both counts, and it lines up with the repo's own "avoid classes for logical implementations" rule. Splitting it by risk:

**Done here:** `refresh_tools` was dead (definition only, zero callers anywhere in the repo) and it was the only non-test caller of `reset`, so both are gone along with the test that existed only to cover `reset`.

**Not here:** the lazy `tools_dict()` and moving `_convert_mcp_to_langchain_tool`, `_get_handler_type_hints`, `_build_args_schema_from_handler`, `_json_schema_to_pydantic`, `_resolve_ref` and `_get_python_type` out to module-level functions. That's ~200 lines of `tools.py` unrelated to this PR's analytics vocabulary, on stack 2 of 4 — stacks 3 and 4 are already written against the current shape and call `category_for`, which is precisely the lazy-discovery path being restructured. Doing it here means re-review of an approved PR plus conflict resolution down the whole stack.

It is not being dropped: a dedicated PR follows immediately after this one merges, covering the lazy accessor and the class → module-function split.

abdulrafey1: Follow-up is open: #588, stacked on this PR so its diff is the restructure alone.

It does the lazy `tools_dict()` you sketched, moves all eight conversion steps out to module
functions, and drops `get_tool`/`get_tool_schemas` as well — neither had a caller either. The
two accessors that remain were `async` while awaiting nothing, so they are sync now.

One thing your suggestion surfaced that I hadn't expected: `None` vs an empty dict matters.
The old `_initialized` flag meant a hook contributing nothing was still "discovered"; a naive
`if not self._tools` would have re-run discovery on every read. That is now explicit and tested.

abdulrafey1:
@hamza-56 flagging a change to `register_event_schema` from #572 that merging `main` forced, since
it inverts a test you wrote deliberately.

**The collision.** #566 gave every plugin a `test_plugin.py` that constructs the plugin directly
(`ChatPlugin()`, in all four plugins). This PR registers analytics schemas from `__init__`, and
`register_event_schema` raised `DuplicateEventTypeError` on *any* second registration of an
`(event_type, version)` — so after the merge `ChatPlugin` became unconstructable twice and two of
main's tests failed. No textual conflict; only the test run surfaced it.

**The resolution.** Re-registering the *identical* class is now a no-op; a *different* class
claiming the identity still raises. Reasons:

- `KeyedClassHook` in `sparkth/lib/hooks.py` already documents exactly this rule for class
  registries — "re-adding the same class under its key is a no-op so module re-imports stay
  idempotent. Only a *different* class claiming an occupied key is a collision" — and
  `AuditEventTypeHook` follows it. `ANALYTICS_EVENTS` stores classes too, but on `KeyedItemHook`,
  which raises on any duplicate.
- The guard's own docstring gives its purpose as stopping a plugin squatting a name. Only a
  different class can squat; the same class cannot.
- The alternative was rewriting chat's `test_plugin.py` to avoid constructing, which would make
  it the odd one out among four identical files and turn "never construct `ChatPlugin` twice"
  into an undocumented process-wide constraint.

`test_register_event_schema_rejects_re_registration` is now
`test_register_event_schema_re_registering_the_same_class_is_a_no_op`;
`test_register_event_schema_raises_on_duplicate_schema` (the different-class case) is untouched
and still green. The `DuplicateEventTypeError` docstring and `docs/guides/plugins.md` are updated.
Rationale is in the merge commit body.

Happy to revert to the strict guard and change chat's test instead if you'd rather keep #572's
semantics — say so and I'll switch it.

PR #576, opened 2026-08-04
feat(plugins): emit conversation and message analytics events

zaira-bibi:
> **Stack 3 of 4** — Analytics Milestone 2. Based on
> [#574](https://github.com/edly-io/sparkth/pull/574), which adds the schemas and emit helpers
> this PR wires up. Review #572 and #574 first.

## What
Wires the first two analytics events into the real chat request path: a `chat.conversation_started` when a conversation is created, and a `chat.message_sent` per instructor turn.

## Changes
- feat(plugins): emit `chat.conversation_started` from `get_or_create_conversation`, on the create branch only
- feat(plugins): return the stored rows from `persist_incoming_messages` and emit `chat.message_sent` per user turn from the completions route
- test(plugins): add end-to-end producer tests driving the real `POST /api/v1/chat/completions` endpoint

## How to Test
1. The producer tests, which drive the real endpoint and assert on rows actually landed in the analytics database:
   ```bash
   ```
   Expected: 3 passed.
2. No regressions in the plugin — in particular the conversation-title suite, which shares the background-task queue this PR touches:
   ```bash
   ```
   Expected: 295 passed.
3. Confirm a second turn on an existing conversation does not re-emit `conversation_started`:
   ```bash
   uv run pytest sparkth/plugins/chat/tests/test_analytics_producers.py -k second_turn -v
   ```

## Notes
- **Emit ordering is deliberate and load-bearing.** Starlette's `BackgroundTasks.__call__` is a plain `for task in self.tasks: await task()` with no per-task isolation, and this milestone's analytics emits propagate their failures by design (see #572). An emit queued *ahead* of real work would therefore let an analytics outage silently stop that work — concretely, `generate_conversation_title` would never run. `emit_conversation_started` is queued last in `get_or_create_conversation` for exactly this reason, with a comment at the call site so it does not get "tidied" back up next to `create_conversation`. **Please keep analytics emits queued after functional background work in any future change to these seams.**
- **`persist_incoming_messages` changed its return type** from `None` to `list[Message]`. It has one call site, which this PR updates. The rows are returned so the caller can measure the *stored* content: this function is the only place that flattens a request's content blocks into message text, so deriving a length from the raw request would duplicate that flattening and could drift from it.
- **No message text reaches any payload.** `chat.message_sent` carries a character count and an attachment flag; the tests assert the question string appears nowhere in the landed payload.
- **No migration.** New event types are rows in the existing `raw_events` table.

### DIFF
--- a/sparkth/plugins/chat/routes/completions.py
+++ b/sparkth/plugins/chat/routes/completions.py
@@ -18,6 +18,7 @@ from sparkth.lib.llm import (
 )
 from sparkth.lib.log import get_logger
 from sparkth.lib.models import User
+from sparkth.plugins.chat.analytics import emit_message_sent
 from sparkth.plugins.chat.classifier import HistoryTurn
 from sparkth.plugins.chat.config import ChatSettings, get_chat_settings
 from sparkth.plugins.chat.constants import LLM_PROVIDER_API_ERRORS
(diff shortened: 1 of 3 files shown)


zaira-bibi:
Closing this as the base has changed by a lot. Will open a new PR on updated codebase.

PR #580, opened 2026-08-07
feat(core): add supported languages and DEFAULT_LANGUAGE setting

abdulrafey1:
## What
Adds the supported-language allowlist and the platform-wide `DEFAULT_LANGUAGE` setting that later work reads from.

## Changes
- feat(core): add `LanguageInfo`, `SUPPORTED_LANGUAGES` (`en`/`es`/`fr`) and `is_supported_language` to `sparkth/core/config.py`
- feat(core): add `Settings.DEFAULT_LANGUAGE`, validated against the allowlist at startup
- docs(core): document `DEFAULT_LANGUAGE` in the configuration reference and add it to `.env`

## How to Test
1. `uv run pytest tests/core/test_language.py -q` — 8 tests pass.
2. Set `DEFAULT_LANGUAGE=klingon` in `.env` and start the app: it fails at startup with `DEFAULT_LANGUAGE must be one of: en, es, fr` rather than accepting the value.
3. Restore `DEFAULT_LANGUAGE=en` and confirm the app boots.

## Notes
New env var: `DEFAULT_LANGUAGE` (defaults to `en`, committed in `.env`). No migration, no API change.

Tags are BCP 47 (RFC 5646) and hyphenated — the form `HTML lang`, `Accept-Language` and the JS `Intl` API consume. The list is deliberately short: LLM output quality varies by language, so a language is added only once generated content in it has been reviewed by a speaker.

`is_supported_language` lives beside the allowlist so the membership rule has exactly one expression — `Settings.DEFAULT_LANGUAGE`'s validator reuses it, which stops the platform default from ever accepting a value the API would reject.

### DIFF
--- a/.env
+++ b/.env
@@ -42,6 +42,10 @@ SERVE_FRONTEND=false
 # Trusted reverse-proxy hops. 0 = ignore X-Forwarded-For (client-forgeable) and use
 # the socket peer address; N = trust the last N entries appended by your own proxies.
 TRUSTED_PROXY_HOPS=0
+# Platform-wide default language: the fallback value for users who have not chosen
+# one in their profile. BCP 47 tag; must be one of the supported languages
+# (en, es, fr) or the app refuses to start.
+DEFAULT_LANGUAGE=en
 
 # Database
 POSTGRES_USER=sparkth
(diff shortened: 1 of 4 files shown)


hamza-56 (approved):
LGTM 👍

PR #588, opened 2026-08-11
refactor(plugins): reshape the chat tool registry around its state

abdulrafey1:
> **No longer stacked — [#574](https://github.com/edly-io/sparkth/pull/574) has merged** (`b053901`),
> and this PR targets `main` directly. Its `tools.py` changes (the `UNKNOWN_TOOL_CATEGORY` move and
> the dead `refresh_tools`/`reset` removal) are in `main` now, so the diff here is the restructure
> alone, as intended.
>
> #574 landed as a squash commit, which conflicted with the same content sitting on this branch as
> several commits. Resolved by merging `main` in and keeping this branch's `tools.py` and
> `test_tool_registry.py`: `main`'s copies of both are byte-identical to `2d2abb1`, the #574 tip
> this branch was built on, so nothing from `main` is dropped.

## What
`ToolRegistry` carried ten methods that never touched instance state, and repeated
`if not self._initialized: self.discover_plugin_tools()` in five accessors as its lazy gate.
This reshapes the module around the only thing that is actually state — the converted tools and
their MCP categories — as [requested in review on #574](https://github.com/edly-io/sparkth/pull/574#discussion_r3727615426).

## Changes
- refactor(plugins): keep only the tool/category cache on `ToolRegistry`; a single
  `tools_dict()` replaces the `_initialized` flag repeated across five accessors. `None` rather
  than an empty dict means "not discovered", so a hook contributing nothing is discovered once
  instead of on every read
- refactor(plugins): move discovery to a module-level `discover_plugin_tools()` returning the
  tools and categories together, which is what lets `category_for` gate on the same lazy pass
- refactor(plugins): move the eight conversion steps — `get_handler_type_hints`,
  `build_args_schema_from_handler`, `convert_mcp_to_langchain_tool`,
  `convert_args_to_handler_types`, `convert_to_pydantic`, `json_schema_to_pydantic`,
  `resolve_ref`, `get_python_type` — out to plain module functions; none read `self`
- refactor(plugins): drop `get_tool` and `get_tool_schemas` (zero callers) and `register_tool`
  (absorbed into discovery)
- refactor(plugins): `get_all_tools` and `get_tools_by_names` were `async` while awaiting
  nothing — now sync, and `resolve_tools` drops the two `await`s
- test(plugins): cover what had no test — chat's own tools being skipped, a tool that fails to
  convert being skipped rather than fatal, discovery running exactly once, and both accessors

## How to Test
1. The registry suite, including the new discovery and laziness tests:
   ```bash
   uv run pytest sparkth/plugins/chat/tests/test_tool_registry.py -v
   ```
   Expected: 31 passed.
2. No behavioural regression anywhere in the plugin, including the completion routes that
   consume `resolve_tools`:
   ```bash
   ```
   Expected: 305 passed.
3. Whole suite and types — the accessors changed shape, so callers must be checked:
   ```bash
   uv run pytest -q && uv run mypy sparkth/ tests/
   ```
   Expected: 1697 passed, 3 skipped; mypy clean on 416 files.
4. No import cycle from the module reshuffle:
   ```bash
   uv run python -c "import sparkth.main"
   ```
   Expected: no error, exit 0.

## Notes
- **Pure refactor — no behaviour change intended**, with two deliberate exceptions worth a
  reviewer's eye: `get_all_tools`/`get_tools_by_names` are no longer awaitable, and a registry
  whose hook contributes nothing now stops re-running discovery on every read.
- **`register_tool`, `get_tool` and `get_tool_schemas` are gone.** All three were unreachable
  from outside `tools.py` (`grep` over `sparkth/` and `tests/`); `register_tool`'s body was two
  lines now inlined into discovery.
- The two closures inside `convert_mcp_to_langchain_tool` stay nested — they capture the
  handler and its hints for `StructuredTool`, so they are genuine closures rather than helpers
  that could move out.
- No migration, no new env vars, no dependency changes.


abdulrafey1:
@claude-review

hamza-56 (approved):
LGTM 👍

PR #589, opened 2026-08-11
fix(plugins): enforce the per-user plugin access gate

### LINKED ISSUE #586: Users can call endpoints of plugins they have disabled
## Summary

`PluginAccessMiddleware` never blocks anything. A user who has a plugin disabled can still call that plugin's endpoints — the middleware's 403 branch is unreachable, so the per-user plugin access gate is inert.

## Where

`sparkth/core/plugins/middleware.py:67-72`

```python
def _get_route_plugin_name(self, request: Request) -> str | None:
    for route in request.app.routes:
        match, _ = route.matches(request.scope)
        if match == Match.FULL:
            return get_route_plugin_name(route)
    return None
```

Since FastAPI 0.140, `include_router()` no longer copies sub-routes into `application.routes` — it appends a single lazy `_IncludedRouter` branch. Plugin routers are mounted with `include_router` (`sparkth/main.py:_register_plugin_routes`), so the object that matches here is an `_IncludedRouter`, not the underlying `APIRoute`.

`get_route_plugin_name` (`sparkth/core/routes/__init__.py`) reads `PLUGIN_NAME_ATTRIBUTE` off `route.endpoint`. `_IncludedRouter` has no `.endpoint`, so it returns `None`.

`dispatch` then takes its pass-through branch:

```python
plugin_name = self._get_route_plugin_name(request)
if not plugin_name:
    response = await call_next(request)
    return response
```

The same FastAPI change is already documented in `tests/core/test_assemble_app.py:_route_paths`, which uses `fastapi.routing.iter_route_contexts` to flatten those branches.

## Reproduce

```python
from sparkth.main import assemble_app
from sparkth.core.plugins.middleware import PluginAccessMiddleware

app = assemble_app()
mw = PluginAccessMiddleware(app)
scope = {"type": "http", "method": "POST", "path": "/api/v1/chat/completions",
         "root_path": "", "headers": [], "query_string": b""}

class R:
    def __init__(self, app, scope): self.app, self.scope = app, scope

print(mw._get_route_plugin_name(R(app, scope)))
```

Prints `None` for a route that belongs to the `chat` plugin. Inspecting the object that matched confirms why:

```
matched object: _IncludedRouter | has .endpoint: False
```

## Why it matters

This is the enforcement point for per-user plugin access. With it inert:

- A user can call the endpoints of any plugin they have disabled in their settings.
- A plugin disabled system-wide is still reachable over HTTP (`_check_plugin_access` passes `check_system_enabled=True`, but is never called).
- The failure is silent — no error, no log. The middleware runs on every request and does nothing.

## Test coverage

There is none. The only test naming `PluginAccessMiddleware` is `tests/lib/test_plugins.py:49`, which asserts it is a type — a façade-export check. No test exercises the gate, which is why the regression went unnoticed.

abdulrafey1:
## What

`PluginAccessMiddleware` never blocked anything — a user could call the endpoints of a plugin they had disabled, and a plugin disabled system-wide stayed reachable over HTTP. This makes the gate enforce, and covers it with the behavioural tests it never had.

The gate was inert at **three** independent points, not the one the issue describes. Each silently passed the request on, so the failure was invisible.

| # | Where | Why it did nothing |
|---|---|---|
| 1 | `sparkth/core/routes/__init__.py:28` | `register_router`'s stamping loop iterated `prefixed_router.routes`, which since FastAPI 0.137 is a single lazy `_IncludedRouter` branch with no `.endpoint`. **`PLUGIN_NAME_ATTRIBUTE` was never set on any endpoint** — the loop body never ran. |
| 2 | `sparkth/core/plugins/middleware.py:67` | The lookup matched that same `_IncludedRouter` branch, so `get_route_plugin_name` returned `None`. This is the layer #586 describes. |
| 3 | `sparkth/core/plugins/middleware.py:40` | `dispatch` read `request.state.user`, which **nothing in the codebase writes** — that line was the only mention of `request.state` in `sparkth/`. Auth is a dependency, so it resolves *after* all middleware has run; `user` was always `None`. |

Layers 1 and 2 are the same FastAPI 0.137 change hitting two places: `include_router()` no longer copies sub-routes into the parent's `routes`. Both now flatten those branches with `iter_route_contexts`.

Layer 3 needed a decision rather than a repair, since the middleware cannot be handed a user that has not been resolved yet. It now identifies the caller from the request's bearer token, reusing the helpers `get_current_user` is built from so the two paths cannot drift.

> [!NOTE]
> The snippet in the issue's proposed-fix comment does not work as written, as that comment anticipated: `context.original_route.matches(scope)` returns `Match.NONE`, because `original_route.path` is the *unprefixed* `/completions`. Matching goes through `context.matches(scope)` (whose `path_regex` carries the include prefix); `original_route` is only used to read the stamp.

**Anonymous requests deliberately pass the gate.** Plugin routers carry unauthenticated endpoints — Slack's `oauth_callback` is called by Slack itself, with no token — and a per-user preference is meaningless without a user. Endpoints that do require a caller are still rejected by their own auth dependency.

## Changes

- `fix(plugins)`: stamp the plugin name on the routes `include_router` actually mounts, via `iter_route_contexts`
- `fix(plugins)`: resolve the route's plugin through the flattened route contexts instead of the `_IncludedRouter` branch
- `fix(plugins)`: identify the caller from the request's bearer token, since no middleware populates `request.state.user`
- `fix(plugins)`: drop the `"/"` entry from the default `exclude_paths` — entries are matched with `startswith`, so it excluded every path
- `refactor(auth)`: extract `decode_token_username` and `get_user_by_username` from `get_current_user`, so the gate and the dependency share one implementation
- `refactor(plugins)`: move `BEARER_SCHEME` into `core/plugins/constants.py`, alongside `PLUGIN_NAME_PATTERN`
- `fix(deps)`: raise the `fastapi` floor to `>=0.138.0`, the first version exposing `iter_route_contexts`
- `test(core)`: cover the gate's behaviour — blocking, letting through, failing open on anonymous, failing closed on a database error — plus the plugin-name stamp and the auth helpers
- `docs(plugins)`: document that plugin routes are gated, and that unauthenticated endpoints are not

## How to Test

1. `make mypy && make lint.backend && make lint.format.backend check=1` — all clean.
2. `python -m pytest -q` — 1680 passed, 3 skipped. No existing test changed behaviour.
3. `python -m pytest tests/core/plugins/test_middleware.py tests/core/test_routes.py tests/lib/test_auth.py -q` — 26 passed, the new coverage.
4. Confirm the tests fail without the fix: `git stash` the `sparkth/` changes and re-run step 3. `test_blocks_a_plugin_the_user_disabled` returns **422** instead of 403 — the request sailed past the gate and reached body validation — and `test_register_router_stamps_the_plugin_name_on_its_routes` asserts `[None] == ['stamp-test']`.
5. Run the reproduction from the issue against this branch. It prints `chat` where it printed `None`:

   ```python
   from sparkth.main import assemble_app
   from sparkth.core.plugins.middleware import PluginAccessMiddleware

   app = assemble_app()
   mw = PluginAccessMiddleware(app)
   scope = {"type": "http", "method": "POST", "path": "/api/v1/chat/completions",
            "root_path": "", "headers": [], "query_string": b""}

   class R:
       def __init__(self, app, scope): self.app, self.scope = app, scope

   print(mw._get_route_plugin_name(R(app, scope)))   # -> "chat"
   ```

6. End-to-end, against a running app: disable the chat plugin for a user, then `POST /api/v1/chat/completions` with that user's token. Expect `403` naming the plugin, and a `WARNING` in the logs. Re-enable it and repeat — the request reaches the handler.

## Notes

- No migration and no new env var. One dependency floor raised, see below.
- **Behaviour change, intended:** endpoints of a disabled plugin now return `403` where they previously served the request. That is the bug being fixed, but anyone relying on the gate being inert will see requests start failing.
- A request to a plugin path with the wrong HTTP method matches `Match.PARTIAL`, not `Match.FULL`, so it reaches the router and returns `405` rather than `403`. No plugin functionality is reachable that way; called out because it is visible.
- **Dependency floor raised:** `fastapi[standard]>=0.121.2` → `>=0.138.0`. `iter_route_contexts` does not exist below 0.138, so the previously-permitted range would `ImportError` at startup. Measured per version: `include_router` stopped copying sub-routes in **0.137** (where the gate actually broke, with no `iter_route_contexts` to fix it), and 0.138–0.140 behave identically for everything this code relies on. The `Since FastAPI 0.140` comments, including the pre-existing one in `tests/core/test_assemble_app.py`, are corrected to 0.137.
- Gated requests cost one extra database round trip (resolve the caller, then check access) in a single session. The route scan in `_get_route_plugin_name` runs on every non-excluded request as it always has — unchanged by this PR; only the token/DB cost is new, and that falls solely on plugin requests.
- **The gate covers HTTP routes only — a plugin's MCP tools are not gated at all.** `/ai/mcp` is a mount, so no plugin name resolves, and the MCP surface has no authenticated caller to check a preference against. Disabling a plugin does not stop its tools being called over MCP. Documented in `docs/guides/plugins.md` and tracked in #591; deliberately not fixed here, since the correct seam is a call-time FastMCP middleware rather than anything in this gate.
- Worth a follow-up, out of scope here: this gate would fit the codebase better as a route dependency (`dependencies=[...]`, as `CLAUDE.local.md` prescribes for gatekeeping) than as middleware. It would run after routing, so it would need neither its own token decoding nor any coupling to FastAPI's routing internals — which is what broke here, in two places at once. That is an architecture change, and this security fix should be reviewed on its own.

### DIFF
--- a/docs/guides/plugins.md
+++ b/docs/guides/plugins.md
@@ -297,6 +297,11 @@ callback, an inbound webhook — keep working. Those endpoints are reachable reg
 anyone's plugin settings, so authenticate them the way you would any other public endpoint
 (a signed state parameter, a request signature) rather than relying on the gate.
 
+That applies to the system-wide switch too: disabling a plugin turns its endpoints off for
+every *identified* caller, but its unauthenticated endpoints stay reachable, because the
+gate has no caller to check them against. Treat "disabled" as a per-caller answer, not as a
+kill switch for the plugin's HTTP surface.
+
 ## Frontend Metadata
 
 The backend is the single source of truth for what the frontend shows about a plugin. Declare it through three per-concern hooks from `sparkth.lib.frontend.hooks`, registered in `__init__` like every other hook. Human-facing names are passed positionally:
--- a/docs/guides/plugins.md
+++ b/docs/guides/plugins.md
@@ -302,6 +302,13 @@ every *identified* caller, but its unauthenticated endpoints stay reachable, bec
 gate has no caller to check them against. Treat "disabled" as a per-caller answer, not as a
 kill switch for the plugin's HTTP surface.
 
+The gate covers HTTP routes and nothing else — **a plugin's MCP tools are not gated at
+all**. `/ai/mcp` is a mount rather than a plugin route, so no plugin name resolves for a
+request to it, and the MCP surface carries no caller identity to check a preference
+against. Disabling a plugin, for one user or system-wide, does not stop its tools being
+called over MCP. If a tool must not run for someone, enforce that in the tool itself; do
+not rely on the plugin being switched off.
+
 ## Frontend Metadata
 
 The backend is the single source of truth for what the frontend shows about a plugin. Declare it through three per-concern hooks from `sparkth.lib.frontend.hooks`, registered in `__init__` like every other hook. Human-facing names are passed positionally:
--- a/docs/guides/plugins.md
+++ b/docs/guides/plugins.md
@@ -283,6 +283,20 @@ Analytics event schemas register through `register_event_schema(self, MyEvent)`:
 the plugin instance's declared name. The router above is reachable at
 `http://localhost:7727/api/v1/my-app/`.
 
+### Who can reach plugin routes?
+
+Every route you register is gated by `PluginAccessMiddleware`. A request from a caller who
+has turned your plugin off in their settings — or a request to a plugin disabled
+system-wide — is answered with `403` before it reaches your handler, so a handler never has
+to check plugin access itself.
+
+The gate identifies the caller from the request's bearer token. A request carrying no
+readable token passes straight through it: a per-user preference is meaningless without a
+user, so endpoints that are called by someone other than a logged-in user — an OAuth
+callback, an inbound webhook — keep working. Those endpoints are reachable regardless of
+anyone's plugin settings, so authenticate them the way you would any other public endpoint
+(a signed state parameter, a request signature) rather than relying on the gate.
+
 ## Frontend Metadata
 
 The backend is the single source of truth for what the frontend shows about a plugin. Declare it through three per-concern hooks from `sparkth.lib.frontend.hooks`, registered in `__init__` like every other hook. Human-facing names are passed positionally:
--- a/sparkth/core/plugins/middleware.py
+++ b/sparkth/core/plugins/middleware.py
@@ -10,14 +10,13 @@ from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.routing import Match
 
 from sparkth.core.models.plugin import Plugin, UserPlugin
+from sparkth.core.plugins.constants import BEARER_SCHEME
 from sparkth.core.routes import get_route_plugin_name
 from sparkth.lib.auth import decode_token_username, get_user_by_username
 from sparkth.lib.db import session_scope
 from sparkth.lib.log import get_logger
 
 logger = get_logger(__name__)
-
-BEARER_SCHEME = "bearer"
 
 
 def _bearer_token_username(request: Request) -> str | None:
--- a/sparkth/core/plugins/middleware.py
+++ b/sparkth/core/plugins/middleware.py
@@ -1,8 +1,8 @@
 from typing import Any, Awaitable, Callable, cast
 
-from fastapi import Request, Response, status
+from fastapi import FastAPI, Request, Response, status
 from fastapi.responses import JSONResponse
-from fastapi.routing import iter_route_contexts
+from fastapi.routing import RouteContext, iter_route_contexts
 from sqlalchemy.exc import DatabaseError, OperationalError
 from sqlmodel import Session, select
 from sqlmodel.ext.asyncio.session import AsyncSession
@@ -10,6 +10,7 @@ from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.routing import Match
 
 from sparkth.core.models.plugin import Plugin, UserPlugin
+from sparkth.core.models.user import User
 from sparkth.core.plugins.constants import BEARER_SCHEME
 from sparkth.core.routes import get_route_plugin_name
 from sparkth.lib.auth import decode_token_username, get_user_by_username
@@ -53,6 +54,10 @@ class PluginAccessMiddleware(BaseHTTPMiddleware):
 
     def __init__(self, app: Any, exclude_paths: list[str] | None = None) -> None:
         super().__init__(app)
+        # Filled on the first request rather than here: ``app`` is the next ASGI app in the
+        # chain, not the FastAPI instance holding the routes, and ``assemble_app`` adds this
+        # middleware before it registers the plugin routers anyway. See _flattened_routes.
+        self._route_contexts: list[RouteContext] | None = None
         # Entries are matched with startswith, so every one of them must be a real path
         # prefix: a bare "/" would exclude every path there is and leave the gate
         # enforcing nothing.
@@ -78,7 +83,12 @@ class PluginAccessMiddleware(BaseHTTPMiddleware):
             response = await call_next(request)
             return response
 
-        has_access = await self._user_may_use_plugin(username, plugin_name)
+        user, has_access = await self._resolve_caller_access(username, plugin_name)
+        if user is not None:
+            # Hand the route the user already loaded here: get_current_user reuses it rather
+            # than decoding the same token and re-reading the same row a second time.
+            request.state.user = user
+
         if not has_access:
             logger.warning(
                 f"User '{username}' attempted to access disabled plugin '{plugin_name}' at path {request.url.path}"
@@ -110,28 +120,59 @@ class PluginAccessMiddleware(BaseHTTPMiddleware):
         and would never match the request — while original_route is the real route whose
         endpoint holds the stamp.
         """
-        for context in iter_route_contexts(request.app.routes):
+        for context in self._flattened_routes(request.app):
             match, _ = context.matches(request.scope)
             if match == Match.FULL:
                 return get_route_plugin_name(context.original_route)
         return None
 
-    async def _user_may_use_plugin(self, username: str, plugin_name: str) -> bool:
-        """Whether the user this token names still has the plugin enabled.
+    def _flattened_routes(self, app: FastAPI) -> list[RouteContext]:
+        """The app's route table, flattened on first use and kept.
+
+        The table is fixed once ``assemble_app`` returns — routes are registered at import
+        time, never per request — so flattening it again on every request is repeated work
+        on a path every request takes. The matching itself still runs per request: request
+        paths carry parameters (``/api/v1/canvas/courses/{course_id}``), so there is no
+        path-keyed answer to cache, only the flattened list to match against.
+
+        Args:
+            app: The FastAPI instance serving the request, taken from its scope.
+
+        Returns:
+            Every route on the app, with nested includes flattened into their own contexts.
+        """
+        if self._route_contexts is None:
+            self._route_contexts = list(iter_route_contexts(app.routes))
+        return self._route_contexts
+
+    async def _resolve_caller_access(self, username: str, plugin_name: str) -> tuple[User | None, bool]:
+        """The user this token names, and whether they may still use the plugin.
+
+        Both answers come out of one session because the caller needs both and the route
+        needs the user again: returning it lets ``get_current_user`` skip a second lookup.
 
         A token naming a user that no longer exists passes: there is no preference to
         enforce, and the route's own auth dependency rejects the request anyway. A database
         failure blocks — the gate cannot confirm access, so it must not grant it.
+
+        Args:
+            username: The username the request's bearer token names.
+            plugin_name: The plugin owning the route being requested.
+
+        Returns:
+            The user, or None when the token names nobody or the lookup failed, paired with
+            whether the request may proceed.
         """
         try:
             async with session_scope() as session:
                 user = await get_user_by_username(username, session)
                 if user is None or user.id is None:
-                    return True
-                return await _check_plugin_access_async(user.id, plugin_name, session, check_system_enabled=True)
+                    return None, True
+                allowed = await _check_plugin_access_async(user.id, plugin_name, session, check_system_enabled=True)
+                return user, allowed
         except (DatabaseError, OperationalError) as e:
             logger.error(f"Database error checking plugin access for user '{username}' and plugin '{plugin_name}': {e}")
-            return False
+            return None, False
 
 
 async def _check_plugin_access_async(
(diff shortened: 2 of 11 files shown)


abdulrafey1:
@claude-review

sparkth/core/plugins/middleware.py
```diff
@@ -65,19 +102,36 @@ def _is_excluded_path(self, path: str) -> bool:
...
-            return False  # Fallback if session generator doesn't yield
+            async with session_scope() as session:
+                user = await get_user_by_username(username, session)
+                if user is None or user.id is None:
+                    return True
+                return await _check_plugin_access_async(user.id, plugin_name, session, check_system_enabled=True)
         except (DatabaseError, OperationalError) as e:
-            logger.error(f"Database error checking plugin access for user {user_id} and plugin '{plugin_name}': {e}")
+            logger.error(f"Database error checking plugin access for user '{username}' and plugin '{plugin_name}': {e}")
             return False
```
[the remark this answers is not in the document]
abdulrafey1: Both branches now have a test, and both were verified to fail when the branch they cover is inverted (flip `return False` → `return True` here and `test_blocks_when_the_access_lookup_fails` fails; flip the `user is None` branch and `test_lets_through_a_token_naming_a_user_that_does_not_exist` fails). The DB-error test patches `get_user_by_username` to raise `OperationalError` and asserts 403.

On the wording: the docstring already says the DB-error path blocks, not that it fails open —

> A database failure blocks — the gate cannot confirm access, so it must not grant it.

The PR body's "failing open" refers to the *anonymous* path in the test-coverage bullet, not to database errors. Neither describes a DB error as failing open, so there's nothing to align here.

abdulrafey1:
## Addressed — with a correction on the version range

### 🔴 fastapi floor — real problem, wrong numbers

The finding holds: `>=0.121.2` permits versions where `iter_route_contexts` does not exist, so the app `ImportError`s at startup. Bumped — but to **`>=0.138.0`**, not `>=0.140.0`, because the suggested range does not match what the packages actually contain. Measured against real installs of each version:

| fastapi | `include_router()` yields | `iter_route_contexts` |
|---|---|---|
| ≤ 0.136.0 | `APIRoute` (has `.endpoint`) — the old code worked | absent |
| 0.137.0 | `_IncludedRouter` — **the gate breaks here** | **absent** — breakage with no way to fix it |
| 0.138.0 | `_IncludedRouter` | present |
| 0.139.0 | `_IncludedRouter` | present |
| 0.140.0 | `_IncludedRouter` | present |

```
0.137.0 -> ImportError: cannot import name 'iter_route_contexts' from 'fastapi.routing'
0.138.0 -> has iter_route_contexts
```

`>=0.140.0` would have worked, but it over-constrains downstream consumers by two minor versions for no reason. 0.138.0 is both necessary and sufficient: the semantics this PR depends on — flattening the branch, `context.matches()` resolving against the *prefixed* path, `original_route.endpoint` carrying the stamp — were exercised on 0.138.0, 0.139.0 and 0.140.0 and all three resolve `chat` identically.

Two consequences worth noting beyond the pin:

- **The gate broke in 0.137, not 0.140.** The `Since FastAPI 0.140` comments — including the pre-existing one in `tests/core/test_assemble_app.py` this PR copied its framing from — were wrong, and are corrected to 0.137.
- **0.137 is a genuine dead zone**: the lazy branches are there, `iter_route_contexts` is not. Nothing to do about it other than exclude it, which the new floor does.

### 🟡 Untested error branches — fixed

`test_blocks_when_the_access_lookup_fails` (patches `get_user_by_username` to raise `OperationalError`, asserts 403) and `test_lets_through_a_token_naming_a_user_that_does_not_exist`. Both were verified honest by inverting the branch each covers and confirming it fails.

The wording nit is answered in the inline thread: the docstring says the DB path *blocks*, and the PR body's "failing open" refers to the anonymous path.

### 🟢 System-wide disable on anonymous endpoints — documented

Correct, and worth stating outright, so `docs/guides/plugins.md` now says it: disabling a plugin turns its endpoints off for every *identified* caller, but its unauthenticated endpoints stay reachable, because the gate has no caller to check them against. "Disabled" is a per-caller answer, not a kill switch for the plugin's HTTP surface.

### 🟢 Route scan — agreed, the note was imprecise

The full scan on every non-excluded request is pre-existing and unchanged; only the token/DB cost is new. Fair point that "non-plugin requests are unaffected" reads as broader than it is.

### 🟢 Route-dependency refactor

Agreed, and the argument is stronger than "eventually": this is now the second FastAPI routing-internals change to break this code, and 0.137 shows the failure can arrive in a release with no escape hatch at all. It stays out of this PR so the security fix is reviewable on its own.

docs/guides/plugins.md
```diff
@@ -283,6 +283,25 @@ Analytics event schemas register through `register_event_schema(self, MyEvent)`:
...
+to check plugin access itself.
+
+The gate identifies the caller from the request's bearer token. A request carrying no
+readable token passes straight through it: a per-user preference is meaningless without a
+user, so endpoints that are called by someone other than a logged-in user — an OAuth
+callback, an inbound webhook — keep working. Those endpoints are reachable regardless of
+anyone's plugin settings, so authenticate them the way you would any other public endpoint
+(a signed state parameter, a request signature) rather than relying on the gate.
+
+That applies to the system-wide switch too: disabling a plugin turns its endpoints off for
```
hamza-56, line 300: This section spells out the gate's holes but leaves out the biggest one: a plugin's MCP tools are not gated at all. `register_plugin_tools` registers every plugin's tools on the shared FastMCP server, and the `/ai/mcp` mount is a Mount with no endpoint, so `_get_route_plugin_name` returns None and the gate always passes. An admin who disables a plugin system-wide, expecting it off, still has all of its tools callable over MCP. Worth stating here so nobody reads "disabled" as covering the plugin's whole surface.

abdulrafey1: Confirmed and documented in `aca30f6`. `/ai/mcp` is not excluded and `_get_route_plugin_name` returns `None` for it, so the gate passes every MCP request through — and `register_plugin_tools` registers every tool unconditionally.

Worth splitting the gap in two, because the halves are not equally fixable: the MCP surface has no authenticated caller at all (`FastMCP` takes no auth provider, the mount has no dependency, handlers take no user id), so *per-user* enforcement has nothing to check against. *System-wide* enforcement needs no identity and is doable now — which is the half your admin example lands on.

The new paragraph says both, and points at the tool itself as the place to enforce rather than the plugin switch.

Fix tracked in #591 with a proposed approach: a call-time FastMCP middleware beside `ToolCallAuditMiddleware`, not registration-time filtering — tools register once at startup while `enabled` flips at runtime. Keeping it out of this PR so the HTTP fix stays reviewable on its own.

One thing that turned up while checking: `check_user_plugin_access`, `get_user_enabled_plugins` and `get_user_disabled_plugins` in `middleware.py` have no callers anywhere. They look like the tool-side gate that was never wired up. Noted in #591 as either its implementation or a deletion.

hamza-56 (approved):

sparkth/core/plugins/middleware.py
```diff
@@ -2,28 +2,64 @@
...
         super().__init__(app)
+        # Entries are matched with startswith, so every one of them must be a real path
+        # prefix: a bare "/" would exclude every path there is and leave the gate
+        # enforcing nothing.
         self.exclude_paths = exclude_paths or [
             "/docs",
             "/redoc",
             "/openapi.json",
-            "/",
             "/api/v1/auth",
```
hamza-56, line 68: Plain `startswith` matches past the segment boundary, so `/api/v1/auth` would also exclude a hypothetical `/api/v1/authors`. Matching `path == p or path.startswith(p + "/")` keeps exclusions scoped to their own subtree.

sparkth/core/plugins/middleware.py
```diff
@@ -65,19 +101,36 @@ def _is_excluded_path(self, path: str) -> bool:
...
+        """Name of the plugin owning the route this request targets, or None for a core route.
+
+        Since FastAPI 0.137 include_router() no longer copies the sub-routes into
+        app.routes: it appends a single lazy _IncludedRouter branch, which has no .endpoint
+        carrying the plugin name. iter_route_contexts flattens those branches into contexts
+        that match on the *prefixed* path — the underlying route's own .path is unprefixed
+        and would never match the request — while original_route is the real route whose
+        endpoint holds the stamp.
+        """
+        for context in iter_route_contexts(request.app.routes):
```
hamza-56: This re-flattens and scans the whole route table on every request, and Starlette then repeats the same matching to dispatch. Routes are static after startup, so building the flattened list (or a path to plugin-name cache) once in `__init__` would be cheap. Not blocking, just runs on every single request.

sparkth/core/plugins/middleware.py
```diff
@@ -65,19 +101,36 @@ def _is_excluded_path(self, path: str) -> bool:
...
+            match, _ = context.matches(request.scope)
             if match == Match.FULL:
-                return get_route_plugin_name(route)
+                return get_route_plugin_name(context.original_route)
         return None
 
-    async def _check_plugin_access(self, user_id: int, plugin_name: str) -> bool:
+    async def _user_may_use_plugin(self, username: str, plugin_name: str) -> bool:
```
hamza-56: Heads up that an authenticated plugin request now costs 3 sequential queries here (user, plugin, user_plugin), and then `get_current_user` decodes the same token and fetches the same user again. Could collapse into one joined query and/or stash the user on `request.state` for the dependency to reuse. Fine to punt, just worth knowing.

hamza-56 (commented):
A few small things, nothing blocking.

sparkth/core/plugins/middleware.py
abdulrafey1: Fixed in `8a69bbb`. Measured before and after over 3000 resolutions: **49.5 µs → 21.4 µs** per request.

Two adjustments to the suggestion, both of which turned out to matter:

**It can't go in `__init__`.** `BaseHTTPMiddleware.__init__` receives the next ASGI app in the chain, not the FastAPI instance holding the routes — the table is only reachable via `request.app` from the scope. And `assemble_app` calls `add_middleware` *before* `_register_plugin_routes`, so even with a reference the routes wouldn't exist yet. It's built on first use and kept instead.

**A path → plugin-name cache would have broken the gate.** Plugin routes carry path parameters (`/api/v1/canvas/courses/{course_id}`), so a map keyed on the request path misses every parameterised route — the gate would silently stop enforcing on exactly the routes it appears to cover, which is the failure mode this PR exists to fix. Only the flattened list is cached; matching still runs per request.

sparkth/core/plugins/middleware.py
abdulrafey1: Confirmed by counting statements on a real request, and fixed in `8a69bbb`: **4 → 2**.

Before: `user` (gate) → `plugins` → `user_plugins` → `user` again (`get_current_user`).
After: `user`, then one `plugins LEFT OUTER JOIN user_plugins`.

Both halves of the suggestion, since they're independent:

- The plugin and the preference now come from one outer join. It's keyed on the user as well as the plugin — a join keyed on the plugin alone would let one user's disabled plugin answer for everyone — and it's an *outer* join so a plugin with no preference row keeps its own row, or the system switch would silently stop applying to every user who never opened their settings. Both cases have their own test; an inner join fails the second, a join missing `user_id` fails the first.
- The gate leaves the user it loaded on `request.state`, and `get_current_user` reuses it rather than decoding the same token and re-reading the same row. Ungated routes never reach the gate, so nothing is stashed and the dependency resolves the caller itself — covered by its own test.

On the reuse being safe: the gate's session has closed by the time the route runs, so the instance is detached. That's harmless here because `User` maps only columns and nothing mutates `current_user` across its 28 consumers. Since adding a relationship to `User` is what would make it unsafe, there's a test asserting `User` has none — it fails loudly at that point rather than at runtime.

sparkth/core/plugins/middleware.py
abdulrafey1: We can fix this once it becomes and issue, I think we can leave this as is for now.

PR #592, opened 2026-08-13
fix(mcp): gate plugin tools on the plugin's enabled state

### LINKED ISSUE #591: Plugin tools remain callable over MCP after the plugin is disabled
## Summary

Disabling a plugin does not stop its MCP tools from being called. Neither the per-user preference nor the system-wide `enabled` flag reaches the MCP surface, so an admin who turns a plugin off — expecting it off — still has every one of its tools callable over `/ai/mcp`.

## Where

`sparkth/mcp/server.py:74` (`register_plugin_tools`), `sparkth/main.py:124` (the mount), and `sparkth/core/plugins/middleware.py` (the gate that does not cover it).

Two independent reasons, both needed to explain the gap:

**1. The gate cannot see MCP requests.** `PluginAccessMiddleware` resolves the owning plugin from the name `register_router` stamps on plugin *endpoints*. `/ai/mcp` is a `Mount`, not a plugin route, so nothing resolves and `dispatch` takes its pass-through branch:

```python
from fastapi import Request
from sparkth.main import assemble_app
from sparkth.core.plugins.middleware import PluginAccessMiddleware

app = assemble_app()
mw = PluginAccessMiddleware(app, ["/docs", "/redoc", "/openapi.json", "/plugins", "/api/v1/auth"])
scope = {"type": "http", "method": "POST", "path": "/ai/mcp",
         "root_path": "", "headers": [], "query_string": b"", "app": app}

print(mw._is_excluded_path("/ai/mcp"))                    # False — not excluded
print(mw._get_route_plugin_name(Request(scope)))          # None  — so the gate passes
```

**2. Every tool is registered unconditionally.** `register_plugin_tools` iterates the `MCP_TOOLS` hook and registers each tool on the shared FastMCP server at startup. Nothing consults `Plugin.enabled` or `UserPlugin.enabled`, at registration or at call time.

## Why it matters

The system-wide flag is an administrative control. `_check_plugin_access_async(..., check_system_enabled=True)` treats a system-disabled plugin as unreachable over HTTP, so "disabled" reads as off — while the same plugin's tools keep running over MCP. A plugin disabled because it is broken, unsafe, or mid-incident is still fully callable through the other door.

The two halves of the gap are not equally fixable:

- **System-wide enforcement is possible today.** It needs no caller identity, only the `Plugin.enabled` row.
- **Per-user enforcement is not.** The MCP surface has no authenticated caller: `FastMCP` is constructed with no auth provider, `mcp.http_app()` is mounted with no dependency, and tool handlers take no user id. There is nothing to check a per-user preference against until that changes.

## Reproduce

1. Register a plugin that contributes an MCP tool (`chat`, `slack`, and `canvas` all do).
2. Disable it system-wide — set `Plugin.enabled = False` for that row.
3. Call any of its tools over `/ai/mcp`.

The tool runs. The equivalent HTTP request to one of that plugin's routes is answered with `403`.

## Related

`sparkth/core/plugins/middleware.py` defines `check_user_plugin_access`, `get_user_enabled_plugins` and `get_user_disabled_plugins`. None of the three has a caller anywhere in the repository — they are shaped like the tool-side gate this issue describes, and either belong to its implementation or should be deleted.

Documented as a known limitation in `docs/guides/plugins.md` by #589, which fixed the HTTP half of the gate.

abdulrafey1:
## What

Disabling a plugin now disables its MCP tools: `/ai/mcp` refuses `tools/call` for a tool
whose plugin is switched off system-wide, and stops advertising it in `tools/list`.

Previously nothing on the MCP surface consulted `Plugin.enabled`, so an administrator
switching a plugin off — because it is broken, unsafe, or mid-incident — still had every
one of its tools reachable through that door, while the equivalent HTTP request was
answered with a `403`.

## Changes

- `fix(mcp)`: add `PluginToolAccessMiddleware` (`sparkth/mcp/access.py`), refusing
  `tools/call` and filtering `tools/list` for system-disabled plugins
- `fix(mcp)`: stamp the contributing plugin's name into each tool's FastMCP `meta` at
  registration, so the gate can resolve ownership per call
- `feat(core)`: add `system_disabled_plugin_names()` — one query per request, returning
  *disabled* names so a plugin with no registry row stays reachable
- `refactor(core)`: remove `check_user_plugin_access`, `get_user_enabled_plugins` and
  `get_user_disabled_plugins`, which had no callers, plus the private sync access check
  whose only caller was one of them
- `test(mcp)`: 15 tests covering call rejection, listing filtering, both failure modes,
  and the audit interaction
- `docs`: rewrite the MCP section of the plugins guide; correct §2 and extend §11 of the
  architectural patterns

### Design notes

**The flag is read per request, not at registration.** Tools register once during the
lifespan while `Plugin.enabled` flips at runtime, so filtering at registration would leave
a plugin disabled afterwards fully callable, and one re-enabled afterwards dark until the
next restart.

**The two failure modes are deliberately opposite.** A database error refuses a *call* —
a lookup that could not complete is not permission to run — but leaves the *listing*
unfiltered, because `sparkth/main.py:80` lists tools inside the lifespan to log its startup
count and a raising filter would take startup down with it. The listing is discoverability;
the call is the boundary, and a tool that slips into the listing is still refused when
called.

**Middleware order is load-bearing.** FastMCP runs the first-added middleware outermost, so
`ToolCallAuditMiddleware` staying first means the calls this gate refuses still leave a
`tool.failed` audit record. A test pins the ordering.

**Enforcement is system-wide only.** `/ai/mcp` carries no authenticated caller, so there is
no user whose per-user preference could be checked.

## How to Test

1. `uv run pytest tests/mcp/test_access.py -v` — 15 tests, all green.
2. `uv run pytest` — full suite: 1695 passed, 3 skipped.
3. Verify the tests are honest rather than merely green. Each of these inversions must
   fail the tests named beside it:
   - make `on_call_tool` return `await call_next(context)` unconditionally →
     `rejects_a_tool_whose_plugin_is_disabled_system_wide`,
     `reads_the_flag_at_call_time_not_at_registration_time`,
     `a_database_error_blocks_the_call`, `a_rejected_call_is_still_audited`
   - make `on_list_tools` return `tools` unfiltered → `hides_tools_whose_plugin_is_disabled`
   - drop `meta={PLUGIN_NAME_META_KEY: plugin_name}` from `_register_tool` →
     `a_registered_plugin_tool_carries_its_plugin_name`
   - flip `system_disabled_plugin_names` to select `Plugin.enabled == True` and invert both
     membership checks → `runs_a_tool_whose_plugin_has_no_database_row` (this is the test
     that catches an inverted default making every tool unreachable until its row exists)
4. End to end, against a running server:
   ```bash
   # Disable a plugin system-wide, then call one of its tools over /ai/mcp.
   # The call is refused, and the tool no longer appears in tools/list.
   ```
5. `uv run ruff check sparkth tests && uv run ruff format --check sparkth tests && uv run mypy sparkth tests`

## Notes

- **No migration, no new env var, no dependency change.**
- **Behaviour change on a surface that has never blocked anything.** Callers of a disabled
  plugin's tools now receive a `ToolError` instead of a result. That is the point of the
  change, but it is worth knowing before deploying.
- **Stacked on #589** (`fix/586-plugin-access-middleware-gate`), which owns the plugins-guide
  paragraph this PR rewrites. Merge #589 first.
- **Two gaps remain open, both documented in the plugins guide rather than silently left:**
  - Per-user enforcement is impossible until `/ai/mcp` has an authenticated caller.
  - The chat agent builds LangChain tools from the same `MCP_TOOLS` hook
    (`sparkth/plugins/chat/tools.py`) and that path is ungated — tracked as #593.
- `sparkth/mcp/access.py` imports `sparkth.core.plugins.service` directly rather than
  through the `sparkth.lib.plugins` façade, by decision, to avoid widening the public
  surface for a single internal query.


sparkth/core/plugins/middleware.py
```diff
@@ -177,90 +177,3 @@ async def _check_plugin_access_async(
         return True
 
     return bool(user_plugin.enabled)
```
hamza-56: With the sync helpers gone, `check_system_enabled` has exactly one caller left and it always passes `True`, so the flag can go and the system-level check become unconditional. While here, `enabled` is already a `bool`, so the `bool()` wrapper isn't needed.

hamza-56 (commented):
Looks good overall!

sparkth/core/plugins/middleware.py
abdulrafey1: Both correct — applied in `c602591`.

`check_system_enabled` is gone and the system check is now unconditional. Confirmed it had exactly one caller left, passing `True`, so the flag was configurability nothing used.

The `bool()` wrapper is gone too. Checked rather than assumed, since SQLite has no native boolean and stores 0/1: SQLAlchemy's `Boolean` type applies its result processor to a bare column selection as well as to an ORM load, so both columns come back as real Python `bool` on the backend the suite runs against.

One adjacent thing I took while in there: renamed `_check_plugin_access_async` to `_user_may_use_plugin`. The `_async` suffix only existed to distinguish it from the synchronous twin this PR deletes, and its docstring still described it as "shared" logic when one caller is all that remains.

No behaviour change — verified by removing the system check and watching `test_blocks_a_plugin_disabled_system_wide` and `test_blocks_a_plugin_disabled_system_wide_for_a_user_with_no_preference` both fail, then restoring it.

hamza-56 (approved):

Issue #510, opened 2026-07-07
Sparkth UI, emails, API errors, and AI-generated content are English-only


regisb:
# Summary

Sparkth is English-only. Users cannot use the platform, receive emails, or reliably generate course content in another language. For a learning platform aimed at international audiences, this blocks adoption by non-English-speaking institutions and learners.

# Where the problem shows up

There is currently no i18n infrastructure anywhere in the codebase:

- **Frontend UI** — all user-facing strings are hardcoded in JSX (e.g. `frontend/app/dashboard/settings/page.tsx`); no i18n library is installed and there are no message catalogs. Note the production constraint: the frontend is a static export (`output: "export"` in `frontend/next.config.ts`) served by FastAPI, so any solution requiring server-side locale negotiation or middleware routing will not work.
- **API errors** — ~151 `HTTPException` occurrences across `app/` with hardcoded English `detail` strings and no machine-readable error codes, so the frontend cannot localize them.
- **Transactional emails** — verification emails are rendered from English f-strings in `app/services/email_verification.py` (subject and body), with no locale awareness in `app/core/email.py`.
- **AI-generated content** — the language of chat responses and generated courses is only handled implicitly by prompt instructions ("in the user's language": `app/mcp/prompts/course_generation_prompt.txt:22`, `app/core_plugins/chat/assets/system_prompt.txt:63`). No explicit language is ever injected, so the model guesses from conversation text.
- **User model** — `app/models/user.py` has no `language`/`locale` field, and there is no user-preferences update endpoint (`app/api/v1/user.py` only exposes `GET /me`) nor any language setting in the frontend settings UI. There is nowhere to store a user's language preference.

# How to observe

1. Log in and browse the dashboard: every page is English with no way to switch language.
2. Register a new account with any locale: the verification email arrives in English.
3. Ask the chat plugin to generate a course while writing short/ambiguous messages: the output language is unpredictable because it is inferred from the conversation rather than a stored preference.

_This description was written with the assistance of an LLM (Claude)._

regisb:
# Proposed solution

Four independently shippable phases, cheapest and highest-leverage first.

## 1. A single source of truth for locale

- Add a `language` column to `User` (BCP-47 string, default `"en"`) via a new Alembic migration.
- Add a `PATCH /api/v1/user/me` endpoint and a language picker in the frontend settings page.
[...]

PR #603, opened 2026-08-14
feat(plugins): mark plugin user-facing strings for translation

hamza-56:
Third layer of the i18n stack (#601 → #602 → #603 → #604 → #606 → #607).

## What
Marks the user-facing strings owned by the plugins and the RAG ingestion path, covering route error details, the chat streaming error map, the Slack bot's canned replies, and each plugin's frontend metadata.

## Changes
- feat(plugins): `gettext_noop`-mark the `DisplayInfo`/`SidebarEntry` registrations of all five plugins
- feat(plugins): chat — completion/attachment/conversation route details and the `streaming_error_message` / `rag_retrieval_error_message` maps wrapped in `_()`
- feat(plugins): slack — the bot's canned messages become `lazy_gettext` constants rendered with `str()` at dispatch; OAuth and workspace route details marked
- feat(plugins): googledrive — route details and the persisted document-failure messages marked; throwaway `_` unpacking variables renamed so they cannot shadow the gettext alias
- feat(rag): `ScannedPDFError.USER_MESSAGE` becomes a lazy translation rendered where the failure is stored

## How to Test
1. `uv run pytest sparkth/plugins/chat/tests/test_i18n.py sparkth/plugins/slack/tests/test_i18n.py tests/rag/test_extraction_scanned_pdf.py`
2. Full run: `uv run pytest` (1760 passed on this layer)
3. `make mypy` and `make lint.backend` are clean

## Notes
Deferred, per the phase-1 plan: the chat refusal message (lives in `scope_keywords.json`, entangled with the LLM system prompt), LLM prompts, MCP tool descriptions, Slack webhook plumbing responses, and hand-rolled pagination-parameter validation messages.

### DIFF
--- a/sparkth/plugins/canvas/plugin.py
+++ b/sparkth/plugins/canvas/plugin.py
@@ -8,6 +8,7 @@ from typing import Any
 import sparkth.plugins.canvas.tools as canvas_tools
 from sparkth.lib.config.hooks import CONFIG_SCHEMAS
 from sparkth.lib.frontend.hooks import DISPLAY_INFO, DisplayInfo
+from sparkth.lib.i18n import gettext_noop
 from sparkth.lib.mcp.hooks import MCP_TOOLS, Tool
 from sparkth.lib.plugins import SparkthPlugin
 from sparkth.plugins.canvas.config import CanvasConfig
@@ -32,7 +33,11 @@ class CanvasPlugin(SparkthPlugin):
         super().__init__("canvas")
         CONFIG_SCHEMAS.add_item(self, CanvasConfig)
         DISPLAY_INFO.add_item(
-            self, DisplayInfo("Canvas", "Canvas LMS integration with tools for courses, modules, pages, and quizzes")
+            self,
+            DisplayInfo(
+                gettext_noop("Canvas"),
+                gettext_noop("Canvas LMS integration with tools for courses, modules, pages, and quizzes"),
+            ),
         )
         tools_per_category: list[tuple[str, list[Callable[..., Any]]]] = [
             ("canvas-auth", [canvas_tools.canvas_authenticate]),
--- a/sparkth/plugins/googledrive/plugin.py
+++ b/sparkth/plugins/googledrive/plugin.py
@@ -3,6 +3,7 @@
 import sparkth.plugins.googledrive.models  # noqa: F401 - registers tables in SQLModel metadata
 from sparkth.lib.config.hooks import CONFIG_SCHEMAS
 from sparkth.lib.frontend.hooks import DISPLAY_INFO, FRONTEND_APPS, DisplayInfo, FrontendApp
+from sparkth.lib.i18n import gettext_noop
 from sparkth.lib.plugins import SparkthPlugin
 from sparkth.lib.routes import register_router
 from sparkth.plugins.googledrive.config import GoogleDriveConfig
@@ -20,6 +21,9 @@ class GoogleDrivePlugin(SparkthPlugin):
         super().__init__("google-drive")
         register_router(self, router)
         CONFIG_SCHEMAS.add_item(self, GoogleDriveConfig)
-        DISPLAY_INFO.add_item(self, DisplayInfo("Google Drive", "All imported files from your connected plugins"))
+        DISPLAY_INFO.add_item(
+            self,
+            DisplayInfo(gettext_noop("Google Drive"), gettext_noop("All imported files from your connected plugins")),
+        )
         # Ships a frontend page but no sidebar entry.
         FRONTEND_APPS.add_item(self, FrontendApp())
--- a/sparkth/plugins/googledrive/routes/oauth.py
+++ b/sparkth/plugins/googledrive/routes/oauth.py
@@ -6,6 +6,7 @@ from sqlmodel.ext.asyncio.session import AsyncSession
 
 from sparkth.lib.auth import get_current_user
 from sparkth.lib.db import get_async_session
+from sparkth.lib.i18n import _
 from sparkth.lib.log import get_logger
 from sparkth.lib.models import User
 from sparkth.plugins.googledrive.oauth import (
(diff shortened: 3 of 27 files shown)


hamza-56:
@claude-review

sparkth/plugins/canvas/plugin.py
```diff
@@ -32,7 +33,11 @@ def __init__(self) -> None:
         super().__init__("canvas")
         CONFIG_SCHEMAS.add_item(self, CanvasConfig)
         DISPLAY_INFO.add_item(
-            self, DisplayInfo("Canvas", "Canvas LMS integration with tools for courses, modules, pages, and quizzes")
+            self,
+            DisplayInfo(
+                gettext_noop("Canvas"),
+                gettext_noop("Canvas LMS integration with tools for courses, modules, pages, and quizzes"),
```
abdulrafey1, line 39: why are we not marking plugin descriptions for translation?

sparkth/plugins/googledrive/plugin.py
```diff
@@ -20,6 +21,9 @@ def __init__(self) -> None:
         super().__init__("google-drive")
         register_router(self, router)
         CONFIG_SCHEMAS.add_item(self, GoogleDriveConfig)
-        DISPLAY_INFO.add_item(self, DisplayInfo("Google Drive", "All imported files from your connected plugins"))
+        DISPLAY_INFO.add_item(
+            self,
+            DisplayInfo(gettext_noop("Google Drive"), gettext_noop("All imported files from your connected plugins")),
```
abdulrafey1, line 26: same question, why not translate the description

sparkth/plugins/googledrive/utils.py
```diff
@@ -214,15 +215,15 @@ async def _process_single_file(
         await session.commit()
     except ScannedPDFError:
         logger.warning("RAG processing rejected scanned PDF '%s'", log_name)
-        await update_document_status(session, document_id, DocumentStatus.FAILED, ScannedPDFError.USER_MESSAGE)
+        await update_document_status(session, document_id, DocumentStatus.FAILED, str(ScannedPDFError.USER_MESSAGE))
```
abdulrafey1, line 218: should this not be `_` instead of `str`

sparkth/plugins/slack/routes/__init__.py
```diff
@@ -279,7 +280,7 @@ async def _dispatch_event(
                 "Slack agentic RAG disabled for workspace %s: no AI key configured",
                 workspace_id,
             )
-            answer = NO_AI_KEY_MESSAGE
+            answer = str(NO_AI_KEY_MESSAGE)
```
abdulrafey1, line 283: why are none of the string constants being translated here?

sparkth/plugins/googledrive/routes/oauth.py
```diff
@@ -55,16 +56,16 @@ async def oauth_callback(
         state_data = decode_state(state)
         user_id = state_data["user_id"]
     except SignatureExpired:
-        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="OAuth state expired. Please try again.")
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=_("OAuth state expired. Please try again."))
     except BadSignature, KeyError, ValueError, TypeError:
```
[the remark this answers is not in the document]
hamza-56: This one is a false positive. Unparenthesized `except A, B, C:` is valid as of Python 3.14 ([PEP 758](https://peps.python.org/pep-0758/)), and the project requires 3.14. It parses and the suite collects and passes:

```
$ python -c "import ast; ast.parse(open(\"sparkth/plugins/googledrive/routes/oauth.py\").read())"   # no error
$ uv run pytest sparkth/plugins/googledrive -q  # collects, green
```

No change needed. The parenthesized form in `slack/routes/oauth.py` is there because that block binds the exception with `as exc`, which still requires parentheses.

sparkth/plugins/canvas/plugin.py
hamza-56: It is marked, `gettext_noop` is the marker. `DisplayInfo.description` is a plain `str` field on a dataclass that Pydantic serializes, so a `LazyString` cannot be stored in it, and translating at `__init__` would be wrong anyway: plugins register at import, long before any request locale exists, so it would freeze whatever locale was active at startup for every user.

So the string is marked for extraction here and translated at its rendering boundary, in `UserPluginResponse.for_plugin` (`sparkth/core/plugins/service.py`, added in #602):

```python
display = DisplayInfo(translate(display.display_name), translate(display.description), display.icon)
```

That is the `gettext_noop` pattern documented in `sparkth/lib/i18n.py` and `docs/guides/translations.md`. The description does get translated per request, just not at this line.

sparkth/plugins/googledrive/plugin.py
hamza-56: Same as the Canvas one, `gettext_noop(...)` marks it for extraction and `UserPluginResponse.for_plugin` translates it per request at serialization time. See the reply on `canvas/plugin.py` for the reasoning on why it cannot be translated at registration.

sparkth/plugins/googledrive/utils.py
hamza-56: `ScannedPDFError.USER_MESSAGE` is already marked, with `lazy_gettext` (`sparkth/rag/exceptions.py:30`) — it is a class-level constant, so it evaluates at import before any request locale exists. `str()` is what *renders* it: `LazyString.__str__` calls `gettext` at that point, in the active request locale.

Using `_` here instead would be wrong twice over — `_(LazyString)` is not a `str`, and `pybabel` would try to extract a non-literal. The `lazy_gettext` at the definition + `str()` at the boundary pair is the documented pattern for constants (`docs/guides/translations.md`). The `str()` is also load-bearing beyond translation: `update_document_status` takes a `str`, so a bare `LazyString` would land in the DB column as a repr.

sparkth/plugins/slack/routes/__init__.py
hamza-56: They are translated — the constants in `sparkth/plugins/slack/constants.py` are marked with `lazy_gettext` (e.g. `NO_AI_KEY_MESSAGE` on line 12), because they are module-level and evaluate at import, before any locale exists. This line is their rendering boundary, and `str()` is what triggers the lookup: `LazyString.__str__` calls `gettext` with the locale active for this request.

Marking at the constant and rendering with `str()` at dispatch is the documented pattern for module-level constants (`docs/guides/translations.md`). Wrapping the literal in `_` at the definition instead would translate once at import and serve that one locale to every workspace forever.

abdulrafey1 (approved):

PR #609, opened 2026-08-18
feat(frontend): add i18n foundations with next-intl

hamza-56:
## What
Sets up the frontend static-translation layer with next-intl in cookie-based mode, so the UI can render in any supported language and backend responses arrive in the same one. Part of #510.

## Changes
- feat(frontend): locale config, cookie helpers and catalog loader in `lib/i18n/`
- feat(frontend): `LocaleProvider` resolving the locale client-side (static export has no server to negotiate it) and providing the catalog app-wide
- feat(frontend): `Accept-Language` echoed on every API call via `localeMiddleware`, keeping backend responses in the UI language
- feat(frontend): home page converted to `useTranslations` as the first namespace, with en/es/fr catalogs
- test(frontend): unit tests for locale config, provider, API header and home page; `renderWithIntl` helper for component tests
- build(frontend): i18n-check catalog drift guard wired into `make test.frontend` (missing, invalid, unused and undefined keys)
- docs: translations guide restructured into backend and frontend sections

## How to Test
1. `make test.frontend` — lint, typecheck, vitest (247 tests, 21 new), format and the new `test.frontend.i18n` catalog check all pass.
2. `make frontend.up.dev`, open http://localhost:3000 — the landing page renders in English.
3. In the browser console run `document.cookie = "NEXT_LOCALE=es; path=/"` and reload — the landing copy renders in Spanish, `<html lang>` switches to `es`, and API calls carry `Accept-Language: es`.
4. `cd frontend && NODE_ENV=production bun run build` — the static export builds with next-intl in place.

## Notes
- Dependencies added: `next-intl` (runtime), `@lingual/i18n-check` (dev).
- The Spanish and French catalog strings are LLM-authored and pending native-speaker review, same status as the backend catalogs from #604.
- No migrations, no env vars, no breaking changes.

_This PR description and the code were written with the assistance of an LLM (Claude)._

### DIFF
--- a/frontend/app/LocaleProvider.tsx
+++ b/frontend/app/LocaleProvider.tsx
@@ -4,6 +4,7 @@ import { NextIntlClientProvider } from "next-intl";
 import { useEffect, useState } from "react";
 
 import en from "@/messages/en.json";
+import { setActiveLocale } from "@/lib/i18n/active-locale";
 import { defaultLocale, readLocaleCookie, type Locale } from "@/lib/i18n/config";
 import { loadMessages, type Messages } from "@/lib/i18n/messages";
 
@@ -21,6 +22,7 @@ export function LocaleProvider({ children }: { children: React.ReactNode }) {
       .then((messages) => {
         if (cancelled) return;
         document.documentElement.lang = locale;
+        setActiveLocale(locale);
         setState({ locale, messages });
       })
       .catch((error: unknown) => {
--- a/frontend/app/LocaleProvider.tsx
+++ b/frontend/app/LocaleProvider.tsx
@@ -7,10 +7,6 @@ import en from "@/messages/en.json";
 import { defaultLocale, readLocaleCookie, type Locale } from "@/lib/i18n/config";
 import { loadMessages, type Messages } from "@/lib/i18n/messages";
 
-// Production is a static export, so the locale is resolved client-side from the
-// cookie. Rendering starts synchronously on the bundled English catalog (keeping
-// the prerendered HTML non-empty) and swaps to the cookie locale once its catalog
-// chunk loads; a failed load logs and stays on English.
 export function LocaleProvider({ children }: { children: React.ReactNode }) {
   const [state, setState] = useState<{ locale: Locale; messages: Messages }>({
     locale: defaultLocale,
--- a/frontend/app/LocaleProvider.tsx
+++ b/frontend/app/LocaleProvider.tsx
@@ -3,29 +3,38 @@
 import { NextIntlClientProvider } from "next-intl";
 import { useEffect, useState } from "react";
 
-import { readLocaleCookie, type Locale } from "@/lib/i18n/config";
+import en from "@/messages/en.json";
+import { defaultLocale, readLocaleCookie, type Locale } from "@/lib/i18n/config";
 import { loadMessages, type Messages } from "@/lib/i18n/messages";
 
-// Resolves the UI locale on the client (production is a static export, so there
-// is no server to negotiate it) and provides the matching message catalog.
-// Renders nothing until the catalog is loaded, so no untranslated flash occurs.
+// Production is a static export, so the locale is resolved client-side from the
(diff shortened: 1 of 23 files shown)


hamza-56:
@claude-review

frontend/app/LocaleProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  useEffect(() => {
+    const locale = readLocaleCookie();
+    let cancelled = false;
+    loadMessages(locale).then((messages) => {
+      if (cancelled) return;
+      document.documentElement.lang = locale;
+      setState({ locale, messages });
+    });
+    return () => {
```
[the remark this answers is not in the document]
hamza-56: Fixed in f7ad197. With the provider now seeded synchronously from the bundled English catalog, a rejected chunk import can no longer blank the app: the `.catch` logs the failure and the UI simply stays on English. Covered by a new test (`stays on the default catalog and logs when the locale catalog fails to load`).

frontend/app/LocaleProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+      cancelled = true;
+    };
+  }, []);
+
+  if (state === null) return null;
+  return (
```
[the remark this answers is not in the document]
hamza-56: Adopted in f7ad197: initial state is now seeded synchronously with the statically imported `en.json`, so the prerendered static-export HTML keeps its content (verified: `out/index.html` contains the full landing copy) and the effect swaps to the cookie locale once its chunk loads. The brief English flash for es/fr users is the tradeoff the plan already accepted as inherent to client-side locale resolution on a static export. This also restores the immediate auth check/redirect noted in the nit, since children render on first paint now.

frontend/app/LocaleProvider.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const locale = readLocaleCookie();
+    if (locale === defaultLocale) return;
+    let cancelled = false;
+    loadMessages(locale)
+      .then((messages) => {
+        if (cancelled) return;
+        document.documentElement.lang = locale;
+        setState({ locale, messages });
+      })
+      .catch((error: unknown) => {
```
abdulrafey1, line 28: If `NEXT_LOCALE` is set to `es` and the `es` catalog chunk fails to load, `catch` only logs. State stays at `{ locale: "en", messages: en }` and the cookie is left holding `es`.

The UI renders English while the backend is asked for Spanish, which should not happen.

Something similar happens in `frontend/lib/api/middleware.ts:22,25` and `docs/guides/translations.md:163-171`

frontend/lib/api/middleware.ts
```diff
@@ -18,6 +19,14 @@ export const authMiddleware: Middleware = {
   },
 };
 
+// Echoes the UI locale so backend responses arrive in the language the UI renders.
+export const localeMiddleware: Middleware = {
```
abdulrafey1, line 24: `localeMiddleware` clobbers a caller-supplied Accept-Language; its sibling in the same file deliberately doesn't

`authMiddleware`, eleven lines up, guards against exactly this:

```
// A caller-supplied Authorization header (explicit token param) wins.
if (request.headers.has("Authorization")) return request;
```

frontend/app/LocaleProvider.tsx
hamza-56: Fixed in b7feb758. The provider now records the locale it actually renders in `lib/i18n/active-locale.ts`, and `localeMiddleware` echoes that instead of the cookie, so a failed catalog load keeps both the UI and `Accept-Language` on English while the cookie keeps the preference. Docs updated to match, plus new provider and client tests for the divergence case.

frontend/lib/api/middleware.ts
hamza-56: Fixed in b7feb758: `localeMiddleware` now returns early when the request already carries an `Accept-Language`, same guard as `authMiddleware`. Test added.

abdulrafey1 (approved):

PR #610, opened 2026-08-19
refactor(i18n): move plugin translations into per-plugin catalogs

hamza-56:
## What
Plugin translation catalogs now live inside each plugin instead of the core catalog, so a plugin's translations travel with it when it moves to its own repository.

## Changes
- refactor(plugins): each in-tree plugin owns the `.po` catalogs under its own `locale/` directory and registers it on `LOCALE_DIRS` from its `__init__`; the existing es/fr translations moved out of the core catalog with their translations preserved
- refactor(core): core extraction ignores `plugins/` (`babel.cfg`), and every `i18n.*` Make target loops over the core catalog plus each `sparkth/plugins/*/locale` that exists (a plugin opts in by creating that directory)
- build(docker): the `catalog-builder` stage compiles the core and per-plugin catalogs and hands `sparkth/` (with the compiled `.mo` files) to the runtime stage
- test(core): `tests/core/i18n/test_catalog_containment.py` enforces the boundary: core catalogs reference no plugin sources, plugin catalogs reference only their own plugin, and plugin catalogs cover the same language set as core
- test(plugins): each plugin asserts its `locale/` directory is registered on `LOCALE_DIRS` at import
- fix(plugins): add `sparkth/plugins/__init__.py` so pytest's prepend import mode resolves a plugin `tests/conftest.py` under the canonical package name; without it the plugin `__init__` is re-executed as a second module, which now breaks its import-time hook registration with a duplicate-key error
- docs: update the translations guide, the plugin guide's translations section, and `sparkth/locale/README.md`

## How to Test
1. `uv run pytest tests/core/i18n sparkth/plugins` passes (containment, registration, and the existing plugin translation tests).
2. `make i18n.update` regenerates every catalog; `git diff` shows only the `POT-Creation-Date` header lines changing, so no strings move between catalogs.
3. `make i18n.compile`, then translate a plugin string under the Spanish locale and confirm it is served from the plugin-owned catalog:
   ```bash
   uv run python -c "
   import sparkth.main
   from sparkth.core.i18n import gettext, locale_context
   with locale_context('es'):
       print(gettext(\"I'm not configured with an AI key yet. Please ask your instructor to add one in the bot settings.\"))
   "
   ```
   prints the Spanish translation from `sparkth/plugins/slack/locale/`.

## Notes
No migration, no new env vars. The core es/fr catalogs shrink accordingly: plugin entries moved to `sparkth/plugins/<name>/locale/` via `pybabel update --ignore-obsolete` with no fuzzy entries introduced. The Docker runtime stage now re-copies `sparkth/` from `catalog-builder` instead of only `sparkth/locale`, since the compiled plugin catalogs live inside the package tree.


hamza-56:
@claude-review

abdulrafey1 (approved):

PR #612, opened 2026-08-19
feat(chat): infer the reply and content language from the conversation

abdulrafey1:
Bottom of an 8-PR stack. Ships on its own and does not close the issue.

## What
The chat system prompt named an output language resolved from the signed-in user's stored preference; it now tells the model to write in the language of the user's most recent message, so the assistant follows whoever it is talking to.

## Changes
- feat(plugins): replace the OUTPUT LANGUAGE directive with an inference directive; `get_learning_design_system_prompt()` takes no argument
- feat(plugins): stop resolving a generation language for the chat system prompt
- test(plugins): pin the directive's wording, prove no language name is injected, and add a route-level test that a stored preference does **not** reach the prompt
- docs(core): correct the `SUPPORTED_LANGUAGES` comment and the user guide, both of which said the preference drives generated content

## How to Test
1. `uv run pytest sparkth/plugins/chat/tests/test_language_inference.py sparkth/plugins/chat/tests/test_prompt.py -v`
2. Confirm the negative test is load-bearing: pass a tag back into `get_learning_design_system_prompt` at `routes/completions.py:199` and re-run — `test_stored_language_does_not_reach_the_system_prompt` fails. Revert.
3. In the running app, write to the assistant in Spanish and confirm the reply and any generated outline come back in Spanish without configuring anything.
4. Switch to English mid-conversation and confirm the next turn follows.

## Notes

`SUPPORTED_LANGUAGES` no longer constrains generated output — it is the shipped-interface allowlist only. Its comment previously promised the opposite, so it is corrected here rather than left to contradict the code.

The directive keeps stating the full content scope (titles, lesson text, assessment questions, answer options, feedback) because partial translation of quiz options is a common model failure, and it carves out the case where a user writing in one language asks for the course in another.

**Not covered by any test in this stack:** every test mocks the LLM, so these prove the directive reaches the prompt, never that the output is good. See the stack's top PR for the manual scenarios.

### DIFF
--- a/sparkth/plugins/chat/prompt.py
+++ b/sparkth/plugins/chat/prompt.py
@@ -4,8 +4,6 @@ from datetime import datetime
 from pathlib import Path
 from typing import cast
 
-from sparkth.lib.language import SUPPORTED_LANGUAGES
-
 _ASSETS_DIR = Path(__file__).parent / "assets"
 _scope_cfg: dict[str, object] | None = None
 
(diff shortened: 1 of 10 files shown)


sparkth/plugins/chat/prompt.py
```diff
@@ -41,21 +39,20 @@ def get_current_datetime() -> datetime:
...
-    calling. The tag's English name is what reaches the model: it is the form an LLM
-    follows most reliably, and it keeps the directive readable in logs.
+    No language is injected. The template's OUTPUT LANGUAGE section instructs the model to
+    write in the language of the user's most recent message and to switch when the user
+    does, so the output language is inferred from the conversation rather than resolved
+    from stored state.
 
-    Rendered fresh on every request, so a preference changed mid-conversation takes
-    effect on the very next turn; earlier messages stay as they were.
+    Rendered fresh on every request, which is what lets a language change take effect on
```
hamza-56: The "which is what lets a language change take effect" part is stale now: the switch happens because the model reads the latest message, not because the prompt is re-rendered. The only per-request substitution left is `current_datetime` (the refusal is static).

abdulrafey1: Correct on both counts — fixed in `2e99b47`.

Confirmed the template has exactly two placeholders, `{current_datetime}` and `{refusal_message}`, and `REFUSAL_MESSAGE` is a module constant loaded at import, so the render carries nothing language-related at all. The switch is entirely the model reading the latest message under the template's static OUTPUT LANGUAGE section.

Worth noting how it got that way: on `main` the sentence was true, because the resolved language really was substituted per render (`language_name=SUPPORTED_LANGUAGES[language].name`). This PR removed that substitution and rewrote the sentence into a *stronger* causal claim, so the edit turned an inherited true statement false in the same commit that made it false.

The clause now states the real mechanism and names `current_datetime` as the only substitution that varies per request, so the next reader doesn't have to re-derive what the render depends on.

Also checked it stays accurate through the rest of the stack: #615 turns `REFUSAL_MESSAGE` into a `gettext_noop`, but the render call keeps passing it unwrapped all the way to #619, so "the refusal sentence is a constant" still holds at the tip.

hamza-56 (approved):

PR #613, opened 2026-08-19
feat(chat): infer the conversation title's language

abdulrafey1:
Second of an 8-PR stack. Does not close the issue.

## What
Generated conversation titles were built from a language tag resolved from the stored preference and threaded through two functions into a background task; the model now reads the language off the message it is already given.

## Changes
- feat(plugins): `generate_conversation_title` and `get_or_create_conversation` lose their `language` parameter; the title prompt asks for the language of the message
- feat(plugins): the completion handler no longer resolves a language at all
- test(plugins): the title prompt names no specific language, and the scheduled background task carries no language kwarg

## How to Test
1. `uv run pytest sparkth/plugins/chat/tests/test_conversation_title.py sparkth/plugins/chat/tests/test_language_inference.py -v`
2. `git grep -n "resolve_language" -- sparkth/plugins/` returns nothing — the chat plugin resolves no language anywhere after this PR.
3. Start a **new** conversation in Spanish and confirm the sidebar title is Spanish.

## Notes

A title now follows the language of the message the conversation opened with, which suits a label better than a stored preference did.

`background_tasks.add_task` is typed as a bare `Callable`, so mypy cannot check its kwargs against the task signature — a stale `language=` would fail only at runtime, inside a task whose exceptions are logged and swallowed. `test_scheduled_title_task_receives_no_language` is the only guard against that, and it was verified to fail when the kwarg is reintroduced.

### DIFF
--- a/sparkth/plugins/chat/tests/test_conversation_title.py
+++ b/sparkth/plugins/chat/tests/test_conversation_title.py
@@ -3,7 +3,6 @@ from unittest.mock import AsyncMock, MagicMock
 from sqlmodel import select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
-from sparkth.lib.language import SUPPORTED_LANGUAGES
 from sparkth.plugins.chat.conversation_title import (
     extract_title_from_messages,
     generate_conversation_title,
@@ -150,13 +149,11 @@ class TestGenerateConversationTitle:
         conversation_id: int,
         user_id: int = 42,
         first_user_message: str = "some message",
-        language: str = "en",
     ) -> None:
         await generate_conversation_title(
             conversation_id=conversation_id,
             user_id=user_id,
             first_user_message=first_user_message,
-            language=language,
             service=ChatService(),
             provider=provider,
         )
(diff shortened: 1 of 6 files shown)


sparkth/plugins/chat/tests/test_conversation_title.py
```diff
@@ -236,21 +233,21 @@ async def test_passed_provider_send_message_is_called(self, session: AsyncSessio
...
-        provider = self._provider("Cours de Confidentialité")
-        await self._generate(provider, conversation_id=11, language="fr")
+        provider = self._provider("Some Title")
+        await self._generate(provider, conversation_id=11)
 
-        prompt = provider.send_message.await_args.kwargs["messages"][0]["content"]
-        assert SUPPORTED_LANGUAGES["fr"].name in prompt
-        assert SUPPORTED_LANGUAGES["es"].name not in prompt
+        prompt = provider.send_message.call_args.kwargs["messages"][0]["content"]
+        for name in ("English", "Spanish", "French"):
```
hamza-56: The sibling guard in `test_language_inference.py` iterates `SUPPORTED_LANGUAGES` instead of hardcoding names, so it stays correct when a language is added. Worth doing the same here so this test doesn't silently miss future languages.

abdulrafey1: Done in `e02520f` — the guard now iterates `SUPPORTED_LANGUAGES`, matching the sibling.

Verified the gap was real rather than assuming it: added a fourth language to the allowlist and injected its name into the title prompt. The derived guard fails; the hardcoded tuple passed. That is the failure mode worth naming — it is a negative assertion, so a missing language produces no signal at all, it just quietly checks one language fewer than exists.

One deliberate difference from the sibling: no `!= "English"` exception here. That exclusion exists in `test_language_inference.py` because the system prompt legitimately names English in its phrasing guidance; the title prompt never names it, so English stays one more name this guard holds to. Copying the exclusion across would have removed an assertion that currently works.

Worth flagging the limit, since it is not closed by this: the guard is bounded by the allowlist, and after #614 generation accepts arbitrary tags. A regression injecting something outside the allowlist — `Portuguese (Brazil)`, say — would still pass both guards. Deriving from the allowlist is strictly better than a frozen list, but it does not make the negative assertion airtight.

hamza-56 (approved):

PR #614, opened 2026-08-19
feat(mcp): accept any language tag for course generation

abdulrafey1:
Third of an 8-PR stack. Does not close the issue.

## What
The MCP course-generation tool resolved its caller's language against `SUPPORTED_LANGUAGES`, so a tag outside the shipped-interface allowlist raised `KeyError` on a public tool boundary; any valid BCP 47 tag is now accepted, named via Babel.

## Changes
- feat(core): add `language_display_name`, which names an arbitrary tag and falls back to the platform default for an absent, unparseable, or display-name-less value
- feat(mcp): pass the agent's tag through unclamped, and widen the published field description to say any language is accepted
- fix(core): fall back rather than emitting the literal string `"None"` for a tag Babel parses but has no CLDR English name for
- test(core, mcp): supported and unsupported tags, hyphenated regional subtags, unusable values, and the tool's published schema

## How to Test
1. `uv run pytest tests/core/test_language.py tests/mcp/ -v`
2. Call the tool with `{"course_params": {"course_name": "...", "course_description": "...", "language": "de"}}` and confirm the prompt names German — this raised `KeyError` before.
3. Pass `"pt-BR"` and confirm "Portuguese (Brazil)", then `"klingon"` and confirm it falls back without erroring.
4. `make test.frontend.api` reports no drift: the MCP tool schema is not part of the REST OpenAPI document, so `frontend/lib/api/generated.ts` is untouched.

## Notes
No migration, no env var, no dependency. `babel>=2.17` was already required.

**Public MCP tool signature change** — additive and backward compatible: the accepted value range widens, and omitting `language` still uses the platform default.

Two things about Babel that are easy to get wrong and are load-bearing here. Its parser defaults to the POSIX underscore separator, so `sep="-"` must be passed or every regional tag such as `pt-BR` silently falls back. And `UnknownLocaleError` does **not** subclass `ValueError`, so both must be caught.

The second commit fixes a bug found in review: `Locale.get_display_name()` is typed `str | None` and returns `None` — without raising — for valid tags with no CLDR English name, so `str(...)` rendered the literal text `"None"` into the prompt ("Write the entire course … in None."). Affected real tags include `skr` (Saraiki, ~26M speakers), `lld` and `apc`. mypy cannot catch this, since `str(Optional[str])` type-checks cleanly. The `skr` regression test carries a comment explaining why it is not a duplicate of the `klingon` case — it exercises the `None` return, not the exception path.

### DIFF
--- a/sparkth/mcp/types.py
+++ b/sparkth/mcp/types.py
@@ -7,9 +7,10 @@ class CourseGenerationPromptRequest(BaseModel):
     language: str | None = Field(
         default=None,
         description=(
-            "BCP 47 language tag for the generated course, e.g. 'en', 'es', 'fr'. "
+            "BCP 47 language tag for the generated course, e.g. 'en', 'es', 'de', 'pt-BR'. "
             "The whole course — titles, descriptions, lesson text, assessment questions, "
-            "answer options and feedback — is written in it. Omit it, or pass a tag the "
-            "platform does not support, and the platform default language is used."
+            "answer options and feedback — is written in it. Any language is accepted. "
+            "Omit it, or pass a value that is not a valid language tag, and the platform "
+            "default language is used."
         ),
     )
--- a/sparkth/mcp/types.py
+++ b/sparkth/mcp/types.py
@@ -4,13 +4,19 @@ from pydantic import BaseModel, Field
 class CourseGenerationPromptRequest(BaseModel):
     course_name: str
     course_description: str
+    # 35 is the practical ceiling for a registered BCP 47 tag, the bound ``User.language``
+    # carries for the same reason. This field is read straight off an unauthenticated
+    # request, so without it the caller decides how long the value travelling into the
+    # resolver and the log is.
     language: str | None = Field(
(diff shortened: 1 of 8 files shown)


sparkth/mcp/types.py
```diff
@@ -7,9 +7,10 @@ class CourseGenerationPromptRequest(BaseModel):
     language: str | None = Field(
```
hamza-56, line 11: With the allowlist clamp gone, this field is unbounded at an unauthenticated boundary, and an unusable tag gets echoed twice per call into the logs by [`language_display_name`](https://github.com/edly-io/sparkth/blob/a69177b766af17e3f6d81b1913e7b70052e0cc05/sparkth/core/language.py#L51-L56): once via `%r`, once inside babel's exception message. `User.language` already caps at `max_length=35`; the same bound here would close that off. Also, since that except block swallows with a fallback, CLAUDE.md's exception table asks for `warning` rather than `info`.

PR #615, opened 2026-08-19
feat(chat): translate the chat out-of-scope refusal

abdulrafey1:
Fourth of an 8-PR stack. Does not close the issue.

## What
The chat out-of-scope refusal was a hardcoded English sentence — the last English-only string the assistant produces, and the one a non-English user is most likely to hit first.

## Changes
- feat(plugins): move the sentence out of `scope_keywords.json` into `prompt.py` as a `gettext_noop` constant, and render it with `gettext()` at the four boundaries where the backend emits it itself
- feat(i18n): add Spanish and French catalog entries
- test(plugins): the refusal translates under a non-default locale, the prompt still carries the English source, and both non-streaming render sites are covered

## How to Test
1. `uv run pytest sparkth/plugins/chat/tests/test_refusal_language.py sparkth/plugins/chat/tests/test_scope_validation.py -v`
2. Confirm each render site is load-bearing: revert any one of `completions.py:113`, `:175`, `:184` or `routes/utils/__init__.py:46` to a bare `REFUSAL_MESSAGE` and re-run — its covering test fails. Revert.
3. With `Accept-Language: es`, send an out-of-scope request phrased to trip the English keyword pre-filter (e.g. "write me a poem") and confirm the refusal is Spanish.
4. Send an out-of-scope request in German and confirm the model's own refusal comes back in German.

## Notes

**The two refusal paths are deliberately asymmetric.** The model receives the English source and the system prompt tells it to send that sentence in the conversation's language. The keyword pre-filter and classifier refusals are written by the backend with no model authoring the text, so they render under the interface locale instead. The guides explain why rather than just asserting it.

The sentence had to leave the JSON asset because `pybabel` extracts from Python source only. `gettext_noop` keeps it a plain `str`, which matters: a `LazyString` cannot be substituted into the prompt template, `json.dumps`'d onto the SSE stream, or assigned to a pydantic field.

**⚠️ The Spanish and French translations are machine-generated and need a native speaker's review before this merges.** Both entries carry a translator comment saying so. They are deliberately **not** marked `#, fuzzy` — `pybabel compile` skips fuzzy entries, which would ship English to es/fr users and lose the feature.

### DIFF
--- a/docs/guides/user-management.md
+++ b/docs/guides/user-management.md
@@ -73,11 +73,6 @@ model handles is available.
 - **Generated course content** — titles, descriptions, lesson text, assessment questions,
   answer options and feedback — follows the same language.
 - **Conversation titles** follow the language of the message the conversation opened with.
-- **API error messages and other fixed backend strings** follow the interface language
-  rather than the conversation, because no model is involved in producing them. The
-  out-of-scope refusal is a special case: the assistant sends it in the language of the
-  conversation, while the copy streamed directly by the backend on the same path follows
-  the interface language.
 
 A user writing in one language may ask for the course itself in another; the assistant
 honours that request for the course content and keeps replying in the language the user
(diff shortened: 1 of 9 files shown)


hamza-56 (approved):
LGTM.

PR #616, opened 2026-08-19
refactor(core): make the preferred language the interface locale

abdulrafey1:
Fifth of an 8-PR stack. Does not close the issue.

## What
Translated interface text followed `Accept-Language` alone, because the locale middleware runs before authentication and cannot see a signed-in user — so a user who chose a language in their profile still got whatever their browser advertised.

## Changes
- refactor(core): add `bind_locale` and call it from `get_current_user` once the user is resolved, giving the chain stored preference → `Accept-Language` → `DEFAULT_LANGUAGE`
- fix(core): bind on the cached-user path too, not only the full lookup
- docs(core): correct the middleware docstring, the `.env` comment, the configuration reference, and a stale paragraph in the translations guide

## How to Test
1. `uv run pytest tests/core/i18n/ -v`
2. Confirm precedence is genuinely tested: remove the `_bind_interface_locale` call before `return user` and re-run — `test_stored_language_overrides_accept_language` fails with the French detail. Revert. Then remove the call before `return cached_user` — the cached-path test fails. Revert.
3. `PATCH /api/v1/user/me` with `{"language": "es"}`, then send any request with `Accept-Language: fr` and confirm translated error details come back Spanish, not French.
4. Clear the preference and confirm the header is back in charge.

## Notes
No migration — `User.language` already exists and only its **meaning** changes, from generation language to interface language. That is the whole point of the earlier PRs in this stack.

`bind_locale` is a plain setter rather than a context manager, mirroring `bind_audit_actor`: the request's context is discarded with its task, so there is nothing to restore. It relies on `get_current_user` being `async` — a sync dependency would run in a threadpool with a copied context and lose the binding silently. The docstring says so.

**Binding on the cached path is not an optimisation, it is the feature.** `PluginAccessMiddleware` populates `request.state.user` on every authenticated plugin route, and every chat route is a plugin route — so binding only on the full lookup would have left the stored preference unapplied across the entire surface the previous PR built.

`is_supported_language` is used rather than `resolve_language` deliberately: the latter's fallback would overwrite a negotiated header with `DEFAULT_LANGUAGE`, which is worse than leaving the header in charge.

**@hamza.shafique — this changes what `User.language` means to the static-translation workstream.** The team accepted the semantics; please confirm the PR ownership does not overlap with work you already have in flight.

### DIFF
--- a/sparkth/lib/auth.py
+++ b/sparkth/lib/auth.py
@@ -62,6 +62,23 @@ async def get_user_by_username(username: str, session: AsyncSession) -> User | N
     return result.one_or_none()
 
 
+def _bind_interface_locale(user: User) -> None:
+    """Install ``user``'s stored language as the request's interface locale.
+
+    Called from both of ``get_current_user``'s exits. The cached path matters as much as
+    the full lookup: ``PluginAccessMiddleware`` populates ``request.state.user`` on every
+    authenticated plugin route, so binding only on the lookup path would leave the stored
+    preference unapplied on all of them.
+
+    The membership check is what keeps a language withdrawn from the allowlist from being
+    honoured. ``resolve_language`` is deliberately not used — its fallback would overwrite
+    a negotiated ``Accept-Language`` with ``DEFAULT_LANGUAGE``, which is worse than
+    leaving the header in charge.
+    """
+    if user.language and is_supported_language(user.language):
+        bind_locale(user.language)
+
+
 async def get_current_user(
     request: Request,
     credentials: HTTPAuthorizationCredentials = Depends(security_scheme),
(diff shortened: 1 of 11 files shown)


sparkth/lib/auth.py
```diff
@@ -61,6 +62,23 @@ async def get_user_by_username(username: str, session: AsyncSession) -> User | N
     return result.one_or_none()
 
 
+def _bind_interface_locale(user: User) -> None:
+    """Install ``user``'s stored language as the request's interface locale.
+
+    Called from both of ``get_current_user``'s exits. The cached path matters as much as
+    the full lookup: ``PluginAccessMiddleware`` populates ``request.state.user`` on every
+    authenticated plugin route, so binding only on the lookup path would leave the stored
+    preference unapplied on all of them.
```
hamza-56: This slightly overclaims: [`PluginAccessMiddleware` renders its own translated 403](https://github.com/edly-io/sparkth/blob/524be9db5f914d3ebb44618a59128ab93246fb7e/sparkth/core/plugins/middleware.py#L87-L105) after it has decoded the token and loaded the user, but before anything binds the locale, so that response still follows `Accept-Language` even though the stored preference is in hand. It's the only middleware that short-circuits with a rendered response, so binding there once the user is resolved would close the gap (it already imports from `sparkth.lib.auth`). The "gate that runs ahead of authentication" carve-out in the translations guide reads wrong for the same reason: that gate has authenticated by the time it renders.

hamza-56 (approved):

PR #617, opened 2026-08-19
docs: document how the output language is chosen

abdulrafey1:
Sixth of an 8-PR stack. Documentation only. Does not close the issue.

## What
Sparkth now resolves two unrelated languages — the interface language and the language the model generates in — and conflating them is the easiest mistake to make in this code; the distinction was written down nowhere.

## Changes
- docs: add a "Two languages, resolved separately" section to the translations guide, covering both mechanisms, the consequences of nothing recording a generated language, and why the out-of-scope refusal touches both
- docs: cross-link it from the user-management guide
- docs(plugins): correct a docstring that claimed no model participates in the deterministic refusal path

## How to Test
1. `make docs` — builds strict with no new warnings.
2. Open `site/guides/translations/index.html` and confirm the new section renders above "How the locale is resolved" and that the cross-link resolves.
3. Sweep for anything stale: `git grep -rn "in the user's language" -- sparkth/ docs/` returns only regression assertions, and `git grep -n "language_name" -- sparkth/plugins/chat/` returns only test helpers.

## Notes
No code behaviour changes.

Every earlier PR in this stack corrected the documentation it invalidated; this one adds only the cross-cutting explanation no single PR owned.

The corrected docstring is worth calling out because it is subtle: an LLM scope classifier **does** run on the deterministic refusal path — it decides whether the request is in scope. What the backend does itself is *write the sentence*, which is why that path follows the interface language. The earlier wording ("no model is in the loop") was wrong about the reason while right about the outcome.

`sparkth/mcp/prompts/course_generation_prompt.txt` legitimately still interpolates a language name: that path takes an explicit tag from its caller and has no conversation to infer from.

### DIFF
--- a/docs/guides/translations.md
+++ b/docs/guides/translations.md
@@ -23,8 +23,9 @@ In chat, the prompts instruct the model to write in the language of the user's m
 message and to switch when the user switches; a conversation title instead follows the
 language of the message the conversation opened with, and the MCP course-generation tool
 takes an explicit tag from its caller rather than reading one off a message. Either way,
-there is no tag to store, no setting to configure, and no allowlist to check. Any language
-the model handles works.
+there is no tag to store and no allowlist to check — any language the model handles works.
+The one setting that can decide a generated language is `DEFAULT_LANGUAGE`: the MCP tool
+falls back to it when its caller omits the tag or sends an unusable one.
 
 The practical consequences worth knowing:
 
--- a/docs/guides/translations.md
+++ b/docs/guides/translations.md
@@ -18,12 +18,13 @@ exists or it does not. Everything else in this guide is about this language.
 
 **Generated language** — what the model writes: chat replies, course titles, lesson text,
 assessment questions, answer options, feedback, and conversation titles. It is not
-resolved at all. In chat, the prompts instruct the model to write in the language of the
-user's most recent message and to switch when the user switches; a conversation title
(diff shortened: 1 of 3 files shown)


docs/guides/translations.md
```diff
@@ -5,6 +5,47 @@ messages, emails) with gettext. This guide covers how the locale is resolved,
...
+exists or it does not. Everything else in this guide is about this language.
+
+**Generated language** — what the model writes: chat replies, course titles, lesson text,
+assessment questions, answer options, feedback, and conversation titles. It is not
+resolved from a stored preference or a negotiated header, the way interface language is.
+In chat, the prompts instruct the model to write in the language of the user's most recent
+message and to switch when the user switches; a conversation title instead follows the
+language of the message the conversation opened with, and the MCP course-generation tool
+takes an explicit tag from its caller rather than reading one off a message. Either way,
+there is no tag to store, no setting to configure, and no allowlist to check. Any language
```
hamza-56: "no setting to configure, and no allowlist to check" isn't quite true for the MCP path: an omitted or unusable tag falls back to `DEFAULT_LANGUAGE` in `language_display_name`, and that setting is validated against the allowlist. user-management.md documents that fallback, so the two guides currently disagree; a short qualifier here would fix it.

hamza-56 (approved):

PR #618, opened 2026-08-19
feat(plugins): tag published open edx courses with their language

abdulrafey1:
Seventh of an 8-PR stack. Does not close the issue.

## What
A course published to Open edX carried whatever language the destination defaulted to, so a Spanish course was discoverable as an English one; the model now supplies the tag, because after this stack nothing server-side records which language a course was written in.

## Changes
- feat(plugins): optional BCP 47 `language` on `CreateCourseArgs`, described so the model sets it to the language it wrote the course in
- feat(plugins): `set_course_language` applies it via Studio's course-details endpoint after the run is created; `OpenEdxClient` gains the missing `put` wrapper
- test(plugins): the round trip preserves other detail fields, the tag is lowercased, an omitted tag skips the write, and a failed write still returns the created course
- docs: restore and correct the publishing-language note, and update the translations guide

## How to Test
1. `uv run pytest sparkth/plugins/openedx/ -v`
2. Confirm the silent-failure guard is load-bearing: add `"language": payload.language` to `course_data` in `openedx_create_course_run` and re-run — `test_language_is_not_sent_to_the_course_runs_endpoint` fails. Revert.
3. Against a real Studio, publish a course with `language: "es"` and confirm Settings → Schedule & Details shows Spanish.
4. Publish with the language omitted and confirm the destination default is left alone.

## Notes

**The obvious implementation is wrong, and wrong silently.** Studio's `api/v1/course_runs/` create endpoint accepts only `org`, `number`, `run`, `title`, `schedule`, `pacing_type`, `team` and `images` — there is no language field, and DRF drops unknown keys without complaint, so adding one to the create payload would look like it worked and do nothing. The language lives on `PUT /api/contentstore/v1/course_details/{course_id}` instead. That endpoint is a `PUT` and implements no `PATCH`, so the write reads the details first and puts the whole object back with only `language` replaced — a partial body would wipe every other detail field, and a test pins that.

**Failure semantics are deliberate:** a failed tagging call is logged and swallowed, so a successful course creation is never reported as an error. The course exists and `course_language` is discovery metadata rather than behaviour, so losing the tag must not lose the course. The except tuple includes `aiohttp.ClientError` and `TimeoutError` because the HTTP client converts only non-2xx responses and JSON decode errors — transport failures would otherwise escape and reach the model as a failed tool call while the course already existed, inviting a retry and a duplicate course run.

Canvas is untouched: its course `locale` is read-only over the REST API and settable only through the Canvas UI, so Canvas-published content is in the right language while the metadata is absent. There is nothing to assert, which is why there is no Canvas test.

One accepted hole: the model can name a language it did not actually generate in, and nothing validates the claim. A wrong tag mislabels a course rather than breaking it, and no cheaper check exists without reintroducing the detection this stack removed.

### DIFF
--- a/sparkth/plugins/openedx/tools.py
+++ b/sparkth/plugins/openedx/tools.py
@@ -231,7 +231,7 @@ async def openedx_get_user_info(payload: LMSAccess) -> dict[str, Any]:
             return {"error": {"message": str(err)}}
 
 
-async def set_course_language(auth: AccessTokenPayload, course_id: str, language: str) -> None:
+async def set_course_language(course_id: str, language: str, client: OpenEdxClient, studio_url: str) -> None:
     """Set ``language`` as an existing course's Open edX language.
 
     The course-run creation endpoint has no language field, so this is a second call. The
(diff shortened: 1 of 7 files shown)


sparkth/plugins/openedx/tools.py
```diff
@@ -228,6 +231,43 @@ async def openedx_get_user_info(payload: LMSAccess) -> dict[str, Any]:
...
+    course-details endpoint is a ``PUT`` rather than a ``PATCH``, so the current details are
+    read first and written back with only ``language`` replaced — sending a partial body
+    would clear every other detail field.
+
+    ``language`` is a BCP 47 tag; Open edX language codes are lowercase, so ``pt-BR`` is
+    sent as ``pt-br``. Raises on a failed request: the caller decides whether a missing
+    metadata tag is worth failing over.
+    """
+    endpoint = f"api/contentstore/v1/course_details/{course_id}"
+    async with OpenEdxClient(auth.lms_url, auth.access_token) as client:
```
hamza-56: This opens a second session while the caller's client in `openedx_create_course_run` is still open against the same host and token, making this the only spot in the file with two live sessions (the other helpers run sequentially). Since `set_course_language` is internal-only, it could take the open client plus `studio_url` instead of `auth` and reuse the pool.

hamza-56 (approved):

PR #619, opened 2026-08-19
test(core): guard that the chat refusal stays extractable

abdulrafey1:
Top of an 8-PR stack. Does not close the issue.

## What
An earlier PR in this stack marked the refusal sentence with `gettext_noop`, which is literally `return message` — so no runtime test can see the marking, and deleting the wrapper would silently stop `pybabel` finding the string while every test kept passing.

## Changes
- test(core): assert the refusal is extractable by `pybabel`, scoping the scan to the same tree `babel.cfg` sees
- docs: note that a `gettext_noop` constant is guarded by an extraction test because the marking has no runtime effect
- fix(i18n): cross-cutting corrections from the whole-branch review (see Notes)

## How to Test
1. `uv run pytest tests/core/i18n/ -v`
2. **This is the test that matters, so prove it can fail:** replace `REFUSAL_MESSAGE = gettext_noop(...)` in `sparkth/plugins/chat/prompt.py` with a bare string literal and re-run — `test_the_chat_refusal_is_extractable_by_pybabel` fails. Restore it and confirm it passes.
3. `uv run pytest -q` — 1882 passed, 3 skipped (the 3 need a TimescaleDB instance).
4. `make docs`, `uv run mypy sparkth`, `make lint.format.backend` all clean.

## Notes

**Why an extraction test rather than a runtime one.** `gettext()` is content-keyed, so it keeps translating from an existing catalog entry whether or not the constant is marked. Review proved by mutation that all six tests in `test_refusal_language.py` stay green with the wrapper removed. The damage is silent and delayed: extraction stops seeing the literal, the next `make i18n.update` marks the entry obsolete, and the shipped Spanish and French refusals rot with nothing failing. This test asserts the extractor's view, the only view that can see the marking.

The scan passes a `directory_filter` built from the existing skip-list so it sees exactly what `make i18n.extract` sees. Without it the test scanned a superset, and a membership assertion over a superset can pass where real extraction would fail.

**This PR also carries fixes from the final whole-branch review that touch files owned by earlier PRs in the stack** — a documentation criterion that contradicted the code, a stale rationale in the published API reference, the transport-error gap in the Open edX language write, a test docstring that reintroduced a corrected claim, and a prompt clause that pulled against the refusal directive. They were applied in one pass at the top rather than redistributed down the stack, so a reviewer of an intermediate PR may see something already fixed here. The stack merges bottom-up as one unit and the end state is verified.

**Before merging the stack:**
1. A Spanish and a French speaker should review the two refusal translations added in the fourth PR — they are machine-generated drafts.
2. Run the manual language scenarios. Every automated test mocks the LLM, so 1882 passing tests prove each directive reaches the prompt, never that the output is good.

### DIFF
--- a/tests/core/i18n/test_marking_enforcement.py
+++ b/tests/core/i18n/test_marking_enforcement.py
@@ -32,6 +32,10 @@ CHAT_PLUGIN_DIR = SOURCE_ROOT / "plugins" / "chat"
 # extra `-k` flags the Makefile's i18n.extract target adds on top of them. This is that
 # same effective set, kept in step so a divergence never guards something extraction doesn't.
 EXTRACT_KEYWORDS = DEFAULT_KEYWORDS | {"lazy_gettext": None, "gettext_noop": None}
+# Keywords alone aren't enough: real extraction also skips SKIPPED_DIR_NAMES per babel.cfg's
+# `[ignore: **/tests/**]` / `[ignore: migrations/**]`. _extraction_directory_filter below
+# applies that same skip, so both the markers and the tree scanned stay in step with the
+# Makefile's invocation.
 
 _LITERAL_MESSAGE = (
     f"HTTPException detail is an unmarked string literal: wrap it in _() or add `# {EXEMPT_MARKER}: <reason>`"
@@ -152,6 +156,12 @@ def test_collect_violations_skips_tests_and_migrations(tmp_path: Path) -> None:
     assert "routes.py" in violations[0]
 
 
+def _extraction_directory_filter(dirpath: str) -> bool:
+    """Keep this test's view of the tree identical to ``babel.cfg``'s, so the guard
+    cannot be satisfied by a file real extraction never reads."""
+    return Path(dirpath).name not in SKIPPED_DIR_NAMES
+
+
 def test_the_chat_refusal_is_extractable_by_pybabel() -> None:
     """The refusal must reach the catalogs, and only extraction can prove it does.
 
@@ -168,7 +178,7 @@ def test_the_chat_refusal_is_extractable_by_pybabel() -> None:
     messages = {
         message
         for _filename, _lineno, message, _comments, _context in extract_from_dir(
-            CHAT_PLUGIN_DIR, keywords=EXTRACT_KEYWORDS
+            CHAT_PLUGIN_DIR, keywords=EXTRACT_KEYWORDS, directory_filter=_extraction_directory_filter
         )
         if isinstance(message, str)
     }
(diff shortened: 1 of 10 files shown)


tests/core/i18n/test_marking_enforcement.py
```diff
@@ -19,11 +19,23 @@
...
+from sparkth.plugins.chat.prompt import REFUSAL_MESSAGE
 
 EXEMPT_MARKER = "i18n-exempt"
 SKIPPED_DIR_NAMES = {"tests", "migrations", "__pycache__"}
 SOURCE_ROOT = Path(sparkth.__file__).resolve().parent
+CHAT_PLUGIN_DIR = SOURCE_ROOT / "plugins" / "chat"
+# Babel's DEFAULT_KEYWORDS already cover gettext/_; lazy_gettext and gettext_noop are the
+# extra `-k` flags the Makefile's i18n.extract target adds on top of them. This is that
+# same effective set, kept in step so a divergence never guards something extraction doesn't.
+EXTRACT_KEYWORDS = DEFAULT_KEYWORDS | {"lazy_gettext": None, "gettext_noop": None}
```
hamza-56: Nothing enforces the "kept in step" claim: if `-k gettext_noop` ever drops out of `i18n.extract`, real extraction rots while this guard stays green, which is exactly the failure it exists to catch. A pointer next to the Makefile target back to this constant, or deriving both from one place, would make that drift visible.

tests/core/i18n/test_marking_enforcement.py
```diff
@@ -142,3 +154,32 @@ def test_collect_violations_skips_tests_and_migrations(tmp_path: Path) -> None:
...
     assert "routes.py" in violations[0]
+
+
+def _extraction_directory_filter(dirpath: str) -> bool:
+    """Keep this test's view of the tree identical to ``babel.cfg``'s, so the guard
+    cannot be satisfied by a file real extraction never reads."""
+    return Path(dirpath).name not in SKIPPED_DIR_NAMES
+
+
+def test_the_chat_refusal_is_extractable_by_pybabel() -> None:
```
hamza-56: This test is chat-owned (it imports `REFUSAL_MESSAGE` and hardcodes `plugins/chat`), so per the [test layout rules](https://github.com/edly-io/sparkth/blob/1261a9f4cdd9ae93380c4fffb5dac0335c5641c7/CLAUDE.md#test-layout) it belongs in `sparkth/plugins/chat/tests/`, next to `TestRefusalIsMarked` which already guards the sibling properties of the same constant. As placed, chat moving to its own repo would break this whole core module at import. If the point is the extraction mechanism itself, a synthetic `tmp_path` fixture would test it without the plugin coupling.

hamza-56 (approved):

PR #622, opened 2026-08-21
feat(frontend): let plugins ship their own message catalogs

hamza-56:
## What
Gives frontend plugins the same catalog ownership backend plugins got in #610: a plugin ships and registers its own next-intl messages instead of putting strings in the core catalog, and the chat input placeholder converts as the first plugin-owned string. Part of #510.

## Changes
- feat(frontend): optional `loadMessages` loader on `PluginDefinition`; `lib/i18n/messages.ts` merges every registered plugin's catalog into the core one, falling back to that plugin's English catalog when a chunk fails to load
- feat(frontend): chat plugin ships `plugins/chat/messages/{en,es,fr}.json` scoped to the `chat` namespace; `ChatInput`'s placeholder is the first consumer
- feat(frontend): `LocaleProvider` loads the merged catalog for the default locale too, so plugin messages reach English UIs
- build(frontend): `i18n:check` now runs once per catalog via `frontend/scripts/i18n-check.mjs` (core catalog against core sources, each `plugins/*/messages` against its own plugin), picking up new plugin catalogs automatically
- test(frontend): merge, namespace-containment, and failure-fallback tests in `tests/lib/i18n/messages.test.ts`; a `ChatInput` placeholder test; `renderWithIntl` accepts a plugin catalog
- docs: the translations guide and the frontend plugin guide document the mechanism step by step

## How to Test
1. `make test.frontend`: lint, react-doctor, api drift, typecheck, format and the per-catalog i18n check all pass; vitest is 258 passed with `AnalyticsPage.test.tsx` failing on this machine on `main` too (pre-existing local flake, unrelated).
2. `make frontend.up.dev`, open the chat plugin: the input placeholder comes from the catalog. Set `document.cookie = "NEXT_LOCALE=es; path=/"` and reload: it renders "Describe el curso que quieres crear...".
3. Delete a key from `frontend/plugins/chat/messages/fr.json` and run `make test.frontend.i18n`: the chat catalog check fails; restore it and the check passes.
4. `cd frontend && NODE_ENV=production bun run build`: the static export builds with per-locale plugin chunks.

## Notes
- The es/fr chat strings are LLM-authored and pending native-speaker review, same status as the catalogs from #609.
- Containment rules: core catalogs carry no plugin namespace, a plugin catalog is scoped to exactly its own name, and plugin components use only their own namespace. Enforced by the new tests and the per-catalog drift check.
- No migrations, no env vars, no new dependencies.

_This PR description and the code were written with the assistance of an LLM (Claude)._

### DIFF
--- a/docs/guides/translations.md
+++ b/docs/guides/translations.md
@@ -225,14 +225,15 @@ no locale URL prefix. The locale is resolved on the client.
    prerendered static-export HTML keeps its content. The provider then loads
    the full catalog for the cookie locale: `frontend/messages/<locale>.json`
    merged with the catalog of every registered plugin that ships one (see
-   [plugin catalogs](#plugin-catalogs) below). For a non-default locale it
-   also sets `document.documentElement.lang` and re-renders in that language
-   once the chunks arrive (a brief English flash is inherent to client-side
-   locale resolution); if the load fails it logs and stays on English.
+   [plugin catalogs](#plugin-catalogs) below). Once the chunks arrive it
+   sets `document.documentElement.lang` and re-renders with the merged
+   catalog (for a non-default locale, the brief English flash before that is
+   inherent to client-side locale resolution). A chunk that fails to load
+   logs and falls back to its English counterpart.
 3. The provider records the locale it actually renders in
    `frontend/lib/i18n/active-locale.ts`. The cookie stores the *preference*;
-   the active locale tracks what is on screen, and the two diverge exactly
-   when a catalog load fails and the UI stays on English.
+   the active locale tracks what is on screen, and the two diverge when the
+   catalog load fails outright and the UI stays on English.
 4. The API client echoes the active locale as an `Accept-Language` header on
    every request (`localeMiddleware` in `frontend/lib/api/middleware.ts`), so
    backend responses arrive in the language the frontend renders, cookie or
--- a/docs/guides/translations.md
+++ b/docs/guides/translations.md
@@ -169,11 +169,13 @@ no locale URL prefix. The locale is resolved on the client.
    the layout) reads the `NEXT_LOCALE` cookie via
    `readLocaleCookie()` (`frontend/lib/i18n/config.ts`). An absent or
    unsupported value falls back to `en`.
-2. Rendering starts immediately on the bundled English catalog, so the
-   prerendered static-export HTML keeps its content. For a non-default locale
-   the provider loads `frontend/messages/<locale>.json`, sets
-   `document.documentElement.lang`, and re-renders in that language once the
-   catalog chunk arrives (a brief English flash is inherent to client-side
+2. Rendering starts immediately on the bundled core English catalog, so the
+   prerendered static-export HTML keeps its content. The provider then loads
+   the full catalog for the cookie locale: `frontend/messages/<locale>.json`
+   merged with the catalog of every registered plugin that ships one (see
+   [plugin catalogs](#plugin-catalogs) below). For a non-default locale it
+   also sets `document.documentElement.lang` and re-renders in that language
+   once the chunks arrive (a brief English flash is inherent to client-side
    locale resolution); if the load fails it logs and stays on English.
 3. The provider records the locale it actually renders in
    `frontend/lib/i18n/active-locale.ts`. The cookie stores the *preference*;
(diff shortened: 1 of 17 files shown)


hamza-56:
@claude-review

frontend/lib/i18n/messages.ts
```diff
@@ -1,17 +1,65 @@
...
-export async function loadMessages(locale: Locale): Promise<Messages> {
-  const catalog = await catalogs[locale]();
-  return catalog.default;
+async function loadPluginMessages(locale: Locale): Promise<Record<string, unknown>[]> {
+  return Promise.all(
+    getAllPlugins()
+      .filter((plugin) => plugin.loadMessages !== undefined)
+      .map(async (plugin) => {
+        try {
+          return await plugin.loadMessages!(locale);
```
abdulrafey1: The `.filter()` narrows this by eye but not for TypeScript across the chained `.map()`, so both call sites need `!`.

This is the same shape flagged on #567 ("the `?.` is only there because TS can't narrow across the `.sort()`"), which was fixed there by turning the predicate into a type guard. The same remedy removes both assertions here:

```ts
.filter(
  (plugin): plugin is PluginDefinition & { loadMessages: NonNullable<PluginDefinition["loadMessages"]> } =>
    plugin.loadMessages !== undefined,
)
```

frontend/lib/i18n/messages.ts
```diff
@@ -1,17 +1,65 @@
...
+  );
+}
+
+// Merges the core catalog with every registered plugin's catalog. A plugin
+// catalog that fails to load falls back to that plugin's English catalog (or
+// is skipped entirely), so one broken chunk never takes down the others.
+export async function loadMessages(locale: Locale): Promise<LoadedMessages> {
+  const [core, pluginCatalogs] = await Promise.all([
+    catalogs[locale](),
+    loadPluginMessages(locale),
```
abdulrafey1: `loadPluginMessages` never rejects (every branch resolves), but `catalogs[locale]()` beside it is unguarded. Under `Promise.all`, a core-chunk failure therefore rejects the whole call and throws away plugin catalogs that loaded fine — so the comment above holds for plugin chunks but not for the core one, which is also a chunk.

Concretely, with `NEXT_LOCALE=es` and a stale `es` core chunk after a redeploy: `LocaleProvider`'s `.catch` leaves state at the synchronous seed, which carries no plugin namespaces at all, so `ChatInput`'s `t("inputPlaceholder")` renders the raw key `chat.inputPlaceholder`. That placeholder was a hardcoded literal before this PR, so it could never be key-shaped.

Falling back to the statically imported English core keeps the promise the comment makes:

```ts
const [core, pluginCatalogs] = await Promise.all([
  catalogs[locale]().then((c) => c.default).catch(() => en),
  loadPluginMessages(locale),
]);
return Object.assign({}, core, ...pluginCatalogs) as LoadedMessages;
```

frontend/tests/plugins/chat/components/input/ChatInput.test.tsx
```diff
@@ -0,0 +1,52 @@
+import { screen } from "@testing-library/react";
+import { describe, it, expect, vi } from "vitest";
+
+import { ChatInput } from "@/plugins/chat/components/input/ChatInput";
+import chatEn from "@/plugins/chat/messages/en.json";
+import { renderWithIntl } from "../../../../intl-test-utils";
```
abdulrafey1: CLAUDE.md's Test Layout says the only correct relative imports inside `frontend/tests/` are "to helpers that live in the mirror beside their callers (e.g. `./test-utils`)". This helper sits at the tests root, four levels up.

Every other relative helper import in `frontend/tests/` climbs at most one level (`./test-utils`, `../test-utils`, `../intl-test-utils`), so this is the only `../../../../` in the tree.

`@` maps to the frontend root in both `vitest.config.ts` and `tsconfig.json`, so `@/tests/intl-test-utils` works and stays correct if this file ever moves.

docs/guides/translations.md
```diff
@@ -169,11 +169,13 @@ no locale URL prefix. The locale is resolved on the client.
...
-   prerendered static-export HTML keeps its content. For a non-default locale
-   the provider loads `frontend/messages/<locale>.json`, sets
-   `document.documentElement.lang`, and re-renders in that language once the
-   catalog chunk arrives (a brief English flash is inherent to client-side
+2. Rendering starts immediately on the bundled core English catalog, so the
+   prerendered static-export HTML keeps its content. The provider then loads
+   the full catalog for the cookie locale: `frontend/messages/<locale>.json`
+   merged with the catalog of every registered plugin that ships one (see
+   [plugin catalogs](#plugin-catalogs) below). For a non-default locale it
+   also sets `document.documentElement.lang` and re-renders in that language
```
abdulrafey1: This keeps the "for a non-default locale" gating, but the same commit removes the `if (locale === defaultLocale) return;` early exit from `LocaleProvider`, so the `.then()` body — including the `document.documentElement.lang` assignment — now runs for every locale. The code comment added alongside it says exactly that ("Runs for the default locale too").

No behavioural impact, since setting `lang="en"` when it is already `en` is a no-op. Just worth dropping the qualifier so the guide matches the control flow.

frontend/lib/i18n/messages.ts
hamza-56: Fixed in f3e34818, same type-guard shape as #567; both `!` assertions are gone.

frontend/lib/i18n/messages.ts
hamza-56: Fixed in f3e34818, with a `console.error` on the fallback path and a regression test: a broken `es` core chunk now yields the English core merged with the intact `es` chat catalog. One consequence worth noting: since `loadMessages` no longer rejects for a broken core chunk, the active locale stays on the requested locale in that case (matching how the per-plugin fallback already behaved), so the divergence wording in `active-locale.ts` and the guide now describes only an outright load failure.

frontend/tests/plugins/chat/components/input/ChatInput.test.tsx
hamza-56: Fixed in f3e34818, switched to `@/tests/intl-test-utils`.

docs/guides/translations.md
hamza-56: Fixed in f3e34818. The sentence now describes the load running for every locale, and picks up the per-chunk English fallback from the other thread.

frontend/lib/i18n/messages.ts
```diff
@@ -1,17 +1,65 @@
...
 import type en from "@/messages/en.json";
+import type chatMessages from "@/plugins/chat/messages/en.json";
+
+type CoreMessages = typeof en;
+type PluginMessages = typeof chatMessages;
+
+// The full catalog shape next-intl's types are bound to: the core catalog
+// plus one namespace per plugin that ships messages. A plugin that adds
+// catalogs intersects its `en.json` type into PluginMessages above
+// (type-only, so the plugin stays a self-contained runtime unit).
```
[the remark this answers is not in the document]
hamza-56: Added the pointer in f3e34818: the comment now links to docs/guides/frontend-plugins.md.

frontend/lib/i18n/messages.ts
```diff
@@ -1,17 +1,65 @@
...
+}
+
+// Merges the core catalog with every registered plugin's catalog. A plugin
+// catalog that fails to load falls back to that plugin's English catalog (or
+// is skipped entirely), so one broken chunk never takes down the others.
+export async function loadMessages(locale: Locale): Promise<LoadedMessages> {
+  const [core, pluginCatalogs] = await Promise.all([
+    catalogs[locale](),
+    loadPluginMessages(locale),
+  ]);
```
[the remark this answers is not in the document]
hamza-56: Agreed, keeping it as is while only chat ships a catalog; scoping the merge to enabled plugins is a reasonable follow-up once more plugins add catalogs.

frontend/app/LocaleProvider.tsx
```diff
@@ -6,17 +6,18 @@ import { useEffect, useState } from "react";
...
-  const [state, setState] = useState<{ locale: Locale; messages: Messages }>({
+  const [state, setState] = useState<{ locale: Locale; messages: LoadedMessages }>({
     locale: defaultLocale,
     messages: en,
   });
 
   useEffect(() => {
+    // Runs for the default locale too: the synchronous seed above carries only
+    // the core catalog, and plugin catalogs arrive through this load.
     const locale = readLocaleCookie();
```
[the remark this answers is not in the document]
hamza-56: Agreed, no change: the chat UI is lazy-loaded and mounts after the root effect resolves, so the window is not visible in practice.

abdulrafey1 (approved):

PR #625, opened 2026-08-25
feat(plugins): log every chat scope refusal

abdulrafey1:
> **Base of a stack** — #626 (removing the keyword scope filter) is stacked on this branch.
> Merge this PR first, then retarget #626 to `main`. #626 deletes the keyword-filter log
> added here; the classifier log stays.

## What

The chat assistant's out-of-scope refusal could reach a user mid-conversation with no record of
it anywhere; both gates that can emit it now log the refusal, tagged with the conversation UUID
so it can be traced to the thread a user reports.

## Changes

- `feat(plugins)`: warn on a keyword pre-filter refusal, naming the out-of-scope keywords that
  matched and the absence of an in-scope one
- `feat(plugins)`: warn on a scope-classifier refusal, naming the deciding model, how many
  history turns were available, and the attachment count
- `feat(plugins)`: thread the conversation UUID through `classify_in_scope` into both logs
- `test(plugins)`: cover both logs, the UUID reaching each gate, and that no message text is
  ever logged

### Why it was invisible

`REFUSAL_MESSAGE` has three emitters, and only one of them involves the chat model:

| Emitter | History it sees | Logged before |
|---|---|---|
| `is_query_in_scope()` keyword filter | none — stateless | no |
| `ScopeClassifier.classify()` | last `_MAX_HISTORY_TURNS` messages | no |
| The chat model's own system-prompt refusal | full conversation | n/a |

`classify_in_scope` short-circuits on the keyword filter (`if not is_query_in_scope(...): return
False`), so a refusal there never reaches the LLM classifier, never reaches the chat model, and
left no log line — the request simply returned the refusal. The classifier tier was equally
silent. Neither could be correlated with a user report.

## How to Test

1. Run the suites covering both gates:
   ```bash
   python -m pytest sparkth/plugins/chat/tests/test_scope_validation.py \
     sparkth/plugins/chat/tests/test_classifier.py -q
   ```
2. Trigger the keyword filter directly — no API key or server needed:
   ```bash
   python -c "
   import asyncio, logging
   from uuid import uuid4
   logging.basicConfig(level=logging.INFO)
   from sparkth.plugins.chat.routes.utils import classify_in_scope
   asyncio.run(classify_in_scope('Add a Python code example', 'anthropic', 'k', [], None, uuid4()))
   "
   ```
   Expect one `WARNING sparkth.plugins.chat.prompt` line naming `['code', 'python']` and the UUID
   you passed. Re-run with `None` in place of `uuid4()` and confirm it reads `conversation_uuid=None`.
3. Exercise the classifier tier end to end: start the backend, open a chat, and send a
   mid-conversation reply that carries no out-of-scope keyword but reads as out of scope on its own
   (e.g. a bare `"the second one"` after several turns). If the classifier refuses it, the backend
   log shows `WARNING sparkth.plugins.chat.classifier` whose `conversation_uuid` matches the `?id=`
   in the browser URL.
4. Confirm nothing sensitive is logged: no refusal line should contain any part of the message
   text — only keywords, counts, and lengths.

## Notes

- No migration, no breaking change, no new env var, no dependency change.
- `conversation_uuid` is used for logging only — it is never sent to any model. Both new
  parameters default to `None`, so existing callers are unaffected.
- `conversation_uuid=None` is a real state, not a gap: the first scope check runs deliberately
  before any DB write (so an out-of-scope first message leaves no trace), meaning such a refusal
  has no conversation row to point at. A test pins that the log says so explicitly.
- Whether a stateless keyword filter should be able to refuse mid-conversation at all is a
  separate design question, deliberately left untouched here — these logs exist to inform it.

### DIFF
--- a/sparkth/plugins/chat/routes/completions.py
+++ b/sparkth/plugins/chat/routes/completions.py
@@ -103,7 +103,8 @@ async def chat_completion(
     # Runs before any DB writes so an out-of-scope first message leaves no trace.
     _skip_main_scope_check = False
     if not conversation_uuid:
-        if not await classify_in_scope(query_text, provider_name, api_key, [], None):
+        # No conversation exists yet on this path, so the scope logs carry no uuid.
+        if not await classify_in_scope(query_text, provider_name, api_key, [], None, None):
             if request.stream:
                 return StreamingResponse(
                     stream_out_of_scope_refusal(),
@@ -162,7 +163,12 @@ async def chat_completion(
                 if m is not db_messages[-1] or not (m.role == "user" and m.content == query_text)
             ]
             _in_scope = await classify_in_scope(
-                query_text, llm_config.provider, api_key, prior_history, attached_document_names or None
+                query_text,
+                llm_config.provider,
+                api_key,
+                prior_history,
+                attached_document_names or None,
+                conversation.uuid,
             )
         else:
             _in_scope = True
--- a/sparkth/plugins/chat/classifier.py
+++ b/sparkth/plugins/chat/classifier.py
@@ -1,6 +1,7 @@
 """LLM-based scope classifier for the learning design assistant."""
 
 from typing import Any, Literal, TypedDict
+from uuid import UUID
 
 from langchain_anthropic import ChatAnthropic
 from langchain_core.exceptions import LangChainException
@@ -74,6 +75,7 @@ class ScopeClassifier:
         classifier_model = _CLASSIFIER_MODELS.get(provider_name)
         if classifier_model is None:
             raise ValueError(f"Unsupported provider for classifier: {provider_name!r}")
+        self._model = classifier_model
 
         llm: Any
         match provider_name:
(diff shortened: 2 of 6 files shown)


abdulrafey1:
@claude-review

sparkth/plugins/chat/routes/completions.py
```diff
@@ -103,7 +103,8 @@ async def chat_completion(
     # Runs before any DB writes so an out-of-scope first message leaves no trace.
     _skip_main_scope_check = False
     if not conversation_uuid:
-        if not await classify_in_scope(query_text, provider_name, api_key, [], None):
+        # No conversation exists yet on this path, so the scope logs carry no uuid.
+        if not await classify_in_scope(query_text, provider_name, api_key, [], None, None):
```
hamza-56, line 107: This first-message path is the one the PR exists for (a refusal that leaves no trace), but the log here gets neither a conversation uuid nor a user id, so it still can't be matched to a user report. `user_id` is already in hand at the top of this function.

abdulrafey1: Good catch — and it is broader than this path: at the top of the stack none of the three classifier log lines carried a user id.

Fixed in #640 rather than here, because this log line does not survive the stack: #626 removes the keyword filter and its log entirely, and #629 moves the classifier logging into `MessageScopeClassifier.in_scope`. A `user_id` added here would be deleted before the stack lands, leaving the gap at the tip.

#640 gives both classifiers the user id as constructor state (it does not vary within a request, and never reaches a model) and logs it on a scope refusal, a scope classifier failing open, and a declined search. The fail-open line gains the conversation uuid too — it was naming neither.

PR #626, opened 2026-08-25
refactor(plugins): remove the keyword scope filter from chat

abdulrafey1:
> **Stacked on #625** — base is `refey/feat/chat-scope-refusal-logging`. Review and merge
> #625 first, then retarget this PR to `main`. The diff shown here is this PR's own changes
> only.

## What

Chat scope is now decided by the LLM classifier alone: the lexical keyword filter that ran
in front of it judged messages on wording, refused legitimate turns, and kept a second,
drifting definition of what "in scope" means.

## Changes

- `refactor(plugins)`: drop `is_query_in_scope` and the keyword pattern machinery from
  `sparkth/plugins/chat/prompt.py`; the module now only renders the system prompt and holds
  the refusal copy
- `refactor(plugins)`: delete `sparkth/plugins/chat/assets/scope_keywords.json` (the
  hand-maintained in-scope/out-of-scope word lists)
- `refactor(plugins)`: `classify_in_scope` calls the classifier for every non-empty query;
  the empty-query short-circuit is the only one left, and `conversation_uuid` is now carried
  for the classifier's refusal log alone
- `test(plugins)`: replace the keyword truth-table tests with four that pin the single gate —
  a query with no course vocabulary reaches the classifier, its verdict is returned
  unchanged, an empty query costs no model call, and the conversation UUID is forwarded
- `test(plugins)`: the 13 tests that patched the filter now drive the classifier mock
  directly (9 already mocked it and just dropped the redundant patch)
- `chore(i18n)`: refresh the refusal message's source reference in the es/fr catalogs

## How to Test

1. `make test.backend.pytest sparkth/plugins/chat/` — 328 pass.
   (1889 passed, 3 skipped; the 3 are the `pg` analytics lane).
3. Behaviour check against a running backend (`make services.up && make backend.up.dev`,
   plus `make frontend.up.dev`): in chat, send `translate this into Spanish`. Before, the
   keyword filter refused it outright; now the classifier decides, so it is answered.
4. Send `what is the capital of France?` — still refused, and no conversation appears in
   the sidebar (the gate still runs before any DB write).
5. Check the backend log for that refusal: the warning from #625 is still there, now written
   only by the classifier, carrying the conversation UUID.

## Notes

- Cost/latency: a message that the keyword filter would have refused now pays one
  cheap-model classifier call (`gpt-4o-mini` / `claude-haiku-4-5` / `gemini-2.0-flash`).
  Every message that passed the filter already paid for that call, so the change is
  confined to the refusal path.
- The classifier fails open (returns in-scope on any error), so a classifier outage now
  lets an out-of-scope message through to the chat model, where the system prompt's own
  refusal instruction is the remaining guard.
- This removes the keyword-filter half of #625's logging along with the filter. The
  classifier half — `classifier.py`, `completions.py` and `test_classifier.py` — is
  untouched by this PR.


abdulrafey1:
@claude-review

hamza-56 (approved):

PR #627, opened 2026-08-25
feat(plugins): add a shared base for the chat classifiers

abdulrafey1:
> **Stacked on #626** — base is `refey/refactor/remove-chat-keyword-scope-filter`, which is
> itself stacked on #625. Merge in order: #625 → #626 → this. Nothing depends on this PR yet;
> the two classifiers move onto it in the PRs stacked above it.

## What

The chat plugin's two single-call LLM classifiers each carry their own copy of the same
plumbing; `BaseClassifier` holds it once so a classifier module can be read for its
judgement instead of its wiring.

## Changes

- `feat(plugins)`: add `sparkth/plugins/chat/classifiers/base.py` — `BaseClassifier`, generic
  over an input and an output pydantic model, taking a system prompt and both schemas plus
  the user's provider and key
- `feat(plugins)`: `small_model_for` / `build_small_llm` — classification runs on the
  smallest model of whichever provider the user configured for chat (`gpt-4o-mini` /
  `claude-haiku-4-5` / `gemini-2.0-flash`, `temperature=0`), under that same key
- `feat(plugins)`: add `ClassifierError` and `ClassifierInputError` to
  `sparkth/plugins/chat/exceptions.py`
- `test(plugins)`: 20 tests in `tests/test_base_classifier.py` covering model selection,
  input validation, prompt assembly, output enforcement and failure translation against a
  throwaway subclass

### What the base does, and what it refuses to do

`classify(payload)` validates the payload against the declared input schema, prepends the
system prompt to whatever `_build_messages` returned, invokes the chain, and returns the
parsed output model. Four decisions worth flagging for review:

- **The output schema is the entire answer-format contract.** It goes to
  `with_structured_output`; no format instruction is written into any prompt. It is enforced
  again on the answer: structured output normally returns a parsed model, and a provider path
  that hands back the raw mapping is validated rather than passed through, so a caller always
  receives an instance of what it declared.
- **`_build_messages` is abstract, with no default.** How an input becomes messages is where
  the two classifiers differ most — one replays conversation turns, the other summarises
  attachment structure — so there is nothing to inherit by accident. A subclass that omits it
  cannot be constructed.
- **The base applies no fallback.** A failed call raises `ClassifierError` with the cause
  preserved. What a failed classification *means* differs per classifier (the scope
  classifier fails open, the RAG router fails loud), so that choice stays in the subclass.
- **`ClassifierInputError` is not a `ClassifierError`.** A classifier that falls back when the
  model fails must not also swallow a malformed call, which is a bug at the call site.

## How to Test

1. `make test.backend.pytest sparkth/plugins/chat/tests/test_base_classifier.py` — 20 pass.
   (1909 passed, 3 skipped; the 3 are the `pg` analytics lane).
3. Nothing to check at runtime: the module has no callers in this PR.

## Notes

- No migration, no new env var, no dependency change, no behaviour change — no code path
  reaches this module yet.
- `SMALL_MODELS` duplicates the `_CLASSIFIER_MODELS` map still in `message_scope.py`. The
  duplicate goes when that classifier moves onto the base in the next PR of the stack.
- The RAG intent router currently runs on the user's **full chat model**. Moving it onto this
  base (a later PR in the stack) switches it to the small model — cheaper and faster, and a
  behaviour change to weigh there rather than here.

### DIFF
--- a/sparkth/plugins/chat/classifiers/base.py
+++ b/sparkth/plugins/chat/classifiers/base.py
@@ -20,7 +20,7 @@ from langchain_core.language_models import BaseChatModel
 from langchain_core.messages import BaseMessage, SystemMessage
 from langchain_google_genai import ChatGoogleGenerativeAI
 from langchain_openai import ChatOpenAI
-from pydantic import BaseModel, ValidationError
+from pydantic import BaseModel, SecretStr, ValidationError
 
 from sparkth.plugins.chat.exceptions import ClassifierError, ClassifierInputError
 
@@ -50,17 +50,23 @@ def small_model_for(provider_name: str) -> str:
 def build_small_llm(provider_name: str, api_key: str) -> BaseChatModel:
     """Build the smallest classification model for ``provider_name``, at temperature 0.
 
+    Each client is constructed through its own field names — ``openai_api_key`` and ``model_name``
+    for OpenAI, ``anthropic_api_key`` and ``model`` for Anthropic — rather than the ``api_key`` and
+    ``model`` aliases their docs show. The aliases work at runtime but are absent from the
+    ``__init__`` signatures pydantic generates, so reaching for them costs a type-ignore per call.
+
     Raises:
         ValueError: the provider is unsupported, or has no client wired here.
     """
     model = small_model_for(provider_name)
+    key = SecretStr(api_key)
     match provider_name:
         case "openai":
-            return ChatOpenAI(api_key=api_key, model=model, temperature=0)  # type: ignore[call-arg]
+            return ChatOpenAI(openai_api_key=key, model_name=model, temperature=0)
         case "anthropic":
-            return ChatAnthropic(api_key=api_key, model=model, temperature=0)  # type: ignore[call-arg]
+            return ChatAnthropic(anthropic_api_key=key, model=model, temperature=0)
         case "google":
-            return ChatGoogleGenerativeAI(google_api_key=api_key, model=model, temperature=0)
+            return ChatGoogleGenerativeAI(google_api_key=key, model=model, temperature=0)
     # Only if SMALL_MODELS gains a provider with no client above — a gap here, not bad input.
     raise ValueError(f"No client wired for provider {provider_name!r}")
 
--- a/sparkth/plugins/chat/classifiers/base.py
+++ b/sparkth/plugins/chat/classifiers/base.py
@@ -1,10 +1,10 @@
 """Shared machinery for the chat plugin's classifiers.
 
 A classifier here is one structured-output model call, not an agent: no tools, no reasoning
-loop, no retries. What makes each one distinct is three declarations — the system prompt, the
-input schema a payload must satisfy, and the output schema the answer comes back in — plus how
-it turns a validated input into messages. Subclasses provide exactly those; this module owns
-the rest, so a classifier module can be read for its judgement instead of its plumbing.
+loop, no retries. What makes each one distinct is its system prompt, the two schemas it is
+generic over — what it is asked to judge and the shape of the answer — and how it turns an input
+into messages. Subclasses provide exactly those; this module owns the rest, so a classifier
+module can be read for its judgement instead of its plumbing.
 
 The model is never the user's chat model: classification is a few tokens and wants low
 latency, so it runs on the smallest model of whichever provider the user already configured
@@ -20,7 +20,7 @@ from langchain_core.messages import BaseMessage, SystemMessage
 from pydantic import BaseModel, ValidationError
 
 from sparkth.lib.llm import get_provider
-from sparkth.plugins.chat.exceptions import ClassifierError, ClassifierInputError
+from sparkth.plugins.chat.exceptions import ClassifierError
 
 InputT = TypeVar("InputT", bound=BaseModel)
 OutputT = TypeVar("OutputT", bound=BaseModel)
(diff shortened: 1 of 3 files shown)


abdulrafey1:
@claude-review

sparkth/plugins/chat/classifiers/base.py
```diff
@@ -0,0 +1,148 @@
...
+        ]
+
+        try:
+            answer = await self._chain.ainvoke(messages)
+            # Structured output usually parses the answer already; a raw mapping is validated,
+            # not trusted.
+            if isinstance(answer, self._output_schema):
+                return answer
+            return self._output_schema.model_validate(answer)
+        except (LangChainException, ValidationError) as exc:
```
hamza-56, line 119: Rate limits, auth failures and connection drops come out of the provider SDKs as their own exception types, not `LangChainException` subclasses, so they escape this catch raw instead of becoming `ClassifierError`. That breaks the one-exception-type contract the subclasses are meant to build their fallbacks on, for the most common real failures. `sparkth/llm/providers.py` already catches a wider set for exactly this reason.

sparkth/plugins/chat/classifiers/base.py
```diff
@@ -0,0 +1,148 @@
...
+    Each client is constructed through its own field names — ``openai_api_key`` and ``model_name``
+    for OpenAI, ``anthropic_api_key`` and ``model`` for Anthropic — rather than the ``api_key`` and
+    ``model`` aliases their docs show. The aliases work at runtime but are absent from the
+    ``__init__`` signatures pydantic generates, so reaching for them costs a type-ignore per call.
+
+    Raises:
+        ValueError: no client is wired here for that provider.
+    """
+    key = SecretStr(api_key)
+    match provider_name:
```
hamza-56: This hand-builds the three clients and keeps a plugin-local provider map, but the lib facade already owns provider-to-client construction (`get_provider(...).create_llm()`). That's two registries to keep in sync by hand: a provider added to core silently becomes a `ValueError` here.

sparkth/plugins/chat/classifiers/base.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @abstractmethod
+    def _build_messages(self, payload: InputT) -> list[BaseMessage]:
+        """Render a validated payload into the messages that follow the system prompt.
+
+        This is where classifiers genuinely differ — one replays conversation turns, another
+        summarises attachments — so the base supplies no default.
+        """
+
+    async def classify(self, payload: dict[str, object]) -> OutputT:
```
hamza-56: Taking a plain dict here means `InputT` never reaches the caller, so mypy can't check any call site and a malformed payload only surfaces at runtime as `ClassifierInputError`. From the outside the generic isn't buying anything.

sparkth/plugins/chat/classifiers/base.py
abdulrafey1: This should not be an issue since we would be able to see what specific exception caused the `ClassifierError` via the logs.

Having one clear exception raised here makes in easier to manage it in the caller.

sparkth/plugins/chat/classifiers/base.py
abdulrafey1: Agreed, and reused — thanks.

`build_classifier_llm` is gone. Construction is now `get_provider(provider, key, model, temperature=0).create_llm()`, which takes the match block, the `SecretStr` wrapping, three langchain imports and the per-provider field-name knowledge with it. That last part is the sharpest version of your point: an earlier commit on this branch had hand-written `openai_api_key`/`model_name` versus `anthropic_api_key`/`model` to drop two type-ignores, which is knowledge `providers.py` already carries once.

What stays plugin-local is which model is *smallest* per provider, since the facade has no notion of cheapest. So the second half of your point — two maps drifting — is now pinned by a test: `{entry['id'] for entry in get_provider_catalog()} == set(SMALL_MODELS)`. A provider added to the registry without a small model here fails that test instead of raising `ValueError` on every chat request from whoever selects it. Worth noting that path was worse than it looked: the scope classifier is constructed before the route's `try`, so it would have been an unhandled 500.

The construction tests followed: they assert what the facade is asked for rather than how each client spells its arguments, which removed a per-provider field-name table from the suite.

sparkth/plugins/chat/classifiers/base.py
abdulrafey1: Fixed — and `InputT` stays; it just reaches the caller now.

`classify` takes `InputT` instead of `dict[str, object]`. A misspelled or missing field is a type error at the call site, which I checked by deliberately passing a mapping: `Argument 1 to "classify" of "BaseClassifier" has incompatible type "dict[str, str]"; expected "_ColourInput"`. The two subclasses build their schemas (`MessageScopeInput(...)`, `RAGSearchInput(...)`) rather than dicts.

Two things existed only to serve the dict and went with it:

- the `input_schema` constructor argument — the class's own generic parameter already declares it, and it was used for nothing but `model_validate`
- `ClassifierInputError` — it had one raiser. Pydantic has already validated the values by the time an instance exists, so there was no runtime check left to translate.

Three tests went too: they covered a malformed payload, which is now the type checker's answer rather than a test case.

One incidental gain in the search classifier: it was doing `[h.model_dump() for h in headings]` so the base could validate the dicts back into the same models. It passes the models.

hamza-56 (approved):

PR #629, opened 2026-08-25
refactor(plugins): move the scope classifier onto the shared base

abdulrafey1:
> **Stacked on #627** — base is `rafe/chat-classifier-base`. The chain is
> #625 → #626 → #627 → this. Merge in that order.

## What

The scope classifier now subclasses `BaseClassifier` and keeps only what is specific to
judging scope; its provider-to-model map, client construction and failure handling move to
the base, where the other classifier will share them.

## Changes

- `refactor(plugins)`: `chat/classifier.py` → `chat/classifiers/message_scope.py`, named for
  what it classifies rather than for being the only one, as `MessageScopeClassifier`
- `refactor(plugins)`: its system prompt moves out of the source into
  `assets/message_scope_classifier_system_prompt.txt`, beside the plugin's other prompts,
  read via a `constants.py` entry
- `refactor(plugins)`: `MessageScopeInput`, `MessageScopeVerdict` and `HistoryTurn` move to
  `chat/schemas.py`, replacing the private `_ScopeResult`
- `refactor(plugins)`: `in_scope()` unwraps the verdict, so `classify_in_scope` stays a
  one-line hand-off and no route needs to know the output schema
- `refactor(plugins)`: delete the duplicate provider-to-model map and `match` block — the
  base owns both
- `refactor(plugins)`: delete `classify_in_scope` — with the keyword tier gone and the rest
  moved into `in_scope`, it was two lines forwarding five arguments; the route asks the
  classifier directly, building one instance for its two mutually exclusive checks
- `refactor(plugins)`: the history cap moves to `constants.py` as
  `MESSAGE_SCOPE_CLASSIFIER_CONVERSATION_HISTORY`, where the plugin keeps its tunables
- `feat(plugins)`: `MessageScopeVerdict` gains an optional `refusal_reason` the model fills in
  when it refuses, and the refusal log carries it — see below
- `refactor(plugins)`: `in_scope` judges scope and nothing else — it no longer decides whether
  it was handed a real turn; the pre-flight check reads the turn's attachment names off the
  request, and a request with no user turn is rejected at the boundary
- `test(plugins)`: `test_classifier.py` → `test_message_scope.py`, rewritten: the seven
  provider/model/temperature cases now belong to the base's own suite, and what is left
  asserts turn rendering, the empty-turn rule, the boolean, failing open, and refusal logging
- `test(plugins)`: `test_scope_validation.py` keeps only the route's own behaviour around a
  refusal, and the ~20 classifier patches across five files now target the route module

### `HistoryTurn.role` widens from a union to `str`

It was typed `Literal["user", "assistant"]`, and `completions.py` reached it by
`cast(Literal[...], m.role)` over stored messages that also carry `tool` and `system` roles —
a cast that told the type checker something untrue. The role is now the plain string it always
was, the cast is gone, and `_build_messages` drops the roles it has no turn type for. That
also keeps a stored `system` turn from being replayed to the classifier as a system
instruction.

### Why a refusal is logged, and why as a category

The refusal log carried who and how much — conversation, model, history and attachment counts,
message length — but never what the model took the message for, so a reviewer holding a user's
report could not tell which rule the turn fell under.

`refusal_reason` defaults to empty, so a model that omits it costs the refusal nothing but the
reason. Both the prompt and the field description specify a **category** — "general knowledge
question", "code help" — and forbid quoting the user's message: this log has never carried
message text, which can be confidential course material, and a free-form explanation would
have put it there. Two tests guard the pairing, because a schema field the prompt never
mentions comes back empty and nothing at runtime notices.

## How to Test

1. `make test.backend.pytest sparkth/plugins/chat/` — 343 pass.
   (1904 passed, 3 skipped; the 3 are the `pg` analytics lane).
3. With a backend running (`make services.up && make backend.up.dev`, `make frontend.up.dev`):
   send `what is the capital of France?` — refused, and the log carries
   `Scope classifier refused a message: conversation_uuid=… model=claude-haiku-4-5
   reason='general knowledge question' …`. Confirm the reason names a category and does not
   quote the message.
4. Send `Create a course on data privacy` — answered, and nothing is logged for it.
5. Attach a PDF, type nothing, send — as the very first message of a brand-new chat, and again in
   an existing one. Both are answered, and the classifier is told the attachment name in place of
   the words that were not typed. The second case is the one a reviewer caught: a locally uploaded
   file never becomes a `Document` row, so the conversation's own attachment list is empty and the
   name has to come off the request on that path too. Fixed, and covered by
   `TestAnExistingConversationSeesTheWholeTurnToo`.

## Notes

- **One behaviour change.** A request whose `messages` carry no `user` turn is now a `422`
  instead of being classified. `messages` already required `min_length=1`; what it did not
  require was that any of them came from the user, and the last user message is what the
  completion answers, judges for scope and retrieves against. Reaching this needs a direct API
  caller — the UI cannot produce it.
- An empty query is **not** treated as missing input when documents are attached: sending a
  document with no words is how a user asks the assistant to read it, the UI permits that send,
  and the attachment names are what the model judges instead. The pre-flight check for a new
  conversation now reads those names off the request, which is the only place they exist before
  the conversation row is written.
- No change to which model runs, which key is used, or the refusal a user sees.
- `classifiers/rag_intent.py` is deliberately absent; the RAG intent router moves onto the
  base in the next PR of the stack, which is also where it stops using the user's full chat
  model.


sparkth/plugins/chat/routes/completions.py
```diff
@@ -158,14 +166,12 @@ async def chat_completion(
 
         if not _skip_main_scope_check:
             prior_history: list[HistoryTurn] = [
-                {"role": cast(Literal["user", "assistant"], m.role), "content": m.content}
+                {"role": m.role, "content": m.content}
                 for m in db_messages
                 if m is not db_messages[-1] or not (m.role == "user" and m.content == query_text)
             ]
-            _in_scope = await classify_in_scope(
+            _in_scope = await scope_classifier.in_scope(
```
hamza-56, line 177: An attachment-only send in an *existing* conversation reaches this with an empty query, and for locally uploaded files (which never become `Document` rows) `attached_document_names` is empty too, so the classifier is asked to judge an empty message. Only the new-conversation path above reads the names off the request, though the PR notes say both paths do. Anthropic rejects the empty block outright (wasted call, fail-open warning per send); OpenAI/Gemini may refuse a legitimate send.

sparkth/plugins/chat/classifiers/message_scope.py
```diff
@@ -0,0 +1,110 @@
...
+
+        if not verdict.in_scope:
+            # The only record of a refusal: the chat model is never reached. Counts and lengths
+            # only, since the message may hold course content.
+            logger.warning(
+                "Scope classifier refused a message: conversation_uuid=%s model=%s reason=%r "
+                "history_turns=%d attachments=%d query_len=%d",
+                conversation_uuid,
+                self.model,
+                verdict.refusal_reason,
```
hamza-56, line 104: `refusal_reason` is free text the model writes and it lands in the log verbatim, so the only thing keeping message content out of here is a prompt instruction. The comment above says counts and lengths only because the message may hold course content; that guarantee is soft now.

sparkth/plugins/chat/routes/completions.py
abdulrafey1: Confirmed and fixed, and you were right that the PR notes were wrong — step 5 said both paths read the request when only the pre-flight one did. Notes corrected too.

Reproduced first: a route test posting an attachment-only turn into an existing conversation asserted `attachments == ["syllabus.pdf"]` and got `None`. Nothing in the plugin creates a `Document` row from a base64 block — only `drive_file` blocks map to ids — so `list_conversation_attachments` cannot see a local upload, and the classifier was handed an empty message.

Both paths read it now: the request names are collected once above the branch, and the main check sends them with the conversation's own, deduplicated so a file that is both ingested and re-sent is named once.

One correction to the consequence you described: it was not a wasted call plus a fail-open warning. Anthropic's rejection of the empty block is a `BadRequestError`, which is not a `LangChainException`, so `classify` never wraps it as `ClassifierError` and `in_scope` does not fail open — it propagated to the route's `LLM_PROVIDER_API_ERRORS` handler as a **502 with an error persisted to the conversation**, on a send the UI allows. Which makes your other point about the exception types the load-bearing one; it is noted and parked rather than dismissed.

hamza-56 (approved):

PR #630, opened 2026-08-26
refactor(plugins): move the RAG search decision onto the shared base

abdulrafey1:
> **Stacked on #629** — base is `rafe/chat-scope-classifier-refactor`. The chain is
> #625 → #626 → #627 → #629 → this. Merge in that order.

## What

The retrieval decision moves onto `BaseClassifier` as `RAGSearchClassifier`, keeping only what
deciding retrieval needs, and `resolve_rag_intent` goes with the wrapper it had become.

## Changes

- `refactor(plugins)`: `chat/intent_router.py` → `chat/classifiers/rag_search.py` as
  `RAGSearchClassifier`; prompt asset renamed to `rag_search_classifier_system_prompt.txt` and
  read through `constants.py`
- `refactor(plugins)`: `requires_search(query, documents, conversation_uuid) -> bool` — the one
  boolean the route acts on; the async section-heading lookup happens here, since the base
  defines `_build_messages` as sync
- `refactor(plugins)`: `ClassifierVerdict` base schema holds `refusal_reason`;
  `MessageScopeVerdict` and the new `RAGSearchVerdict` inherit it. `RAGRoutingDecision` is gone
- `refactor(plugins)`: delete `resolve_rag_intent`; the route asks the classifier directly, and
  only when there is something to search
- `refactor(plugins)`: `reason` dropped from the `skipping_rag` SSE event and
  `rag_routing_reason` from `ChatStreamProcessor`, replaced by `rag_search_declined`
- `refactor(plugins)`: `RAGIntentRouterError` → `RAGSearchError`
- `docs`: add [Chat document retrieval](docs/guides/chat-retrieval.md) — when retrieval runs,
  what gets searched, and what happens when the decision fails
- `test(plugins)`: 13 tests in `tests/test_rag_search.py`; `test_intent_router.py` deleted (its
  provider/model cases belong to the base's suite) and `test_intent_router_integration.py`
  renamed to `test_rag_search_integration.py`

### Where the guards went

`resolve_rag_intent` returned `False` when there were no attachments or no query. Asking a model
whether to search nothing is not a question, so the route decides that:

```python
rag_search_required = False
if attached_documents and query_text:
    rag_search_required = await search_classifier.requires_search(
        query_text, attached_documents, conversation.uuid
    )
```

The classifier holds no input guards, matching `MessageScopeClassifier` after the last PR.
Behaviour is unchanged: an attachment-only turn still skips retrieval, since there is no question
to search for.

### The skipping_rag event loses its reason

`RAGRoutingDecision.reason` was emitted on that event. Nothing in `frontend/` handles
`skipping_rag`, so the reason went nowhere; it is now logged with the conversation UUID and the
deciding model. `ChatStreamProcessor` takes `rag_search_declined: bool` instead, which keeps the event
firing only when the classifier actually weighed the turn — not for every message in a
conversation that has no attachments. `should_run_rag` is renamed `rag_search_required`
throughout, since it holds the classifier's answer rather than an instruction.

## How to Test

1. `make test.backend.pytest sparkth/plugins/chat/` — 347 pass.
   (1908 passed, 3 skipped; the 3 are the `pg` analytics lane).
3. `make docs` — the new guide builds and appears under Guides in the nav.
4. With a backend running, attach a document to a conversation and ask about its content —
   retrieval runs (`scanning_attachments`, then section events). Then say "thanks!" in the same
   conversation: the stream carries `skipping_rag` with no reason, and the log shows
   `Search classifier skipped retrieval: conversation_uuid=… model=claude-haiku-4-5
   reason='chit-chat' documents=1`.
5. Send a message in a conversation with no attachments — no `skipping_rag` event, and no
   classifier call at all.

## Notes

- **Behaviour change: the decision runs on the small model.** It used to use the user's full chat
  model via `provider.create_llm()`; the base builds the cheapest capable model for their
  provider instead. Cheaper and faster on every message that has attachments, and less capable
  at judging retrieval intent — the trade we flagged when the base landed.
- **SSE payload change:** `skipping_rag` no longer carries `reason`. No client reads that event.
- Retrieval still fails loud: a classifier failure raises `RAGSearchError`, the route persists an
  error message and the user sees it. Unlike scope, there is no second line of defence that would
  make guessing safe.


sparkth/plugins/chat/classifiers/rag_search.py
```diff
@@ -0,0 +1,109 @@
+"""The RAG search classifier: does answering this message need the attached documents?
```
hamza-56, line 1: Anchoring here since the doc isn't in the diff: `.claude/docs/architectural_patterns.md` (section 9) still lists `sparkth/plugins/chat/intent_router.py` and describes `RAGIntentRouter`, both deleted in this PR.

abdulrafey1: Fixed — and thanks for anchoring it, since nothing in the diff would have pointed there.

Section 9 now lists `sparkth/plugins/chat/classifiers/` instead of the deleted `intent_router.py`, and the `RAGIntentRouter` line is replaced by what the two classifiers actually decide: whether a turn is in scope, and whether answering it needs the attached documents — one structured-output call each on the cheapest model of the user's provider, sharing `BaseClassifier` for everything but the judgement.

I checked the rest of the section and the other docs for references this stack invalidated: those two lines were the only ones.

Also in this push: `RAGSearchClassifier` is exported from the classifiers package, matching what #629 did for `MessageScopeClassifier`, so the route imports both on one line rather than reaching into two modules.

hamza-56 (approved):

PR #631, opened 2026-08-26
refactor(plugins): register the LLM config adapter without a chat subclass

abdulrafey1:
> **Stacked on #630** — base is `rafe/chat-rag-search-classifier`. The chain is
> #625 → #626 → #627 → #629 → #630 → this. Merge in that order.

## What

`chat/adapter.py` held a subclass whose body was `pass`; the chat plugin now registers
`LLMConfigAdapter` itself, so the file and its import go without changing what runs.

## Changes

- `refactor(plugins)`: delete `sparkth/plugins/chat/adapter.py`; `plugin.py` registers
  `LLMConfigAdapter()` on `CONFIG_ADAPTERS` directly
- `test(plugins)`: delete `test_chat_config_adaptor.py` — its three cases re-tested base
  `preprocess_config` / `postprocess_config` behaviour that `tests/llm/test_plugin_adapter.py`
  covers in sixteen
- `test(plugins)`: `test_plugin_adapter.py` now asserts the wiring, not the type — it runs a
  config through `PluginService.apply_preprocess("chat", …)` and requires an unowned
  `llm_config_id` to be rejected
- `docs`: the plugin guide taught the empty subclass as the default path; it now shows
  registering `LLMConfigAdapter` directly, and keeps subclassing for changing the behaviour

### The subclass was pointless; the registration is not

Worth saying plainly, because `class ChatConfigAdapter(LLMConfigAdapter): pass` reads as dead
code. It is reached on every plugin-config request:

`chat/plugin.py` → `CONFIG_ADAPTERS` → `get_plugin_adapter("chat")` in
`core/plugins/service.py:154/171/187` → `api/v1/user_plugins.py` (POST, PUT, and every read).

Dropping the *registration* would have silently disabled the ownership check that stops a user
referencing another user's `LLMConfig`, plus deactivated-key and model-override validation, and
the resolution of `llm_config_name` / `llm_provider` / `llm_model` for the frontend. Only the
subclass is removed here; the behaviour is identical, and the new test fails if the registration
is ever dropped.

### A doc bug fixed alongside

The guide's sample imported `LLMConfigAdapter` from `sparkth.llm.adapter` — an import plugins
are not allowed to make (`CLAUDE.md`, "Public Library"), and one the same section's prose already
gets right two paragraphs above. Both adapter snippets now import from `sparkth.lib.llm`.

## How to Test

1. `make test.backend` — full suite, ruff, ruff format and mypy --strict all pass
   (1905 passed, 3 skipped; the 3 are the `pg` analytics lane). Three duplicate tests removed,
   the registration test rewritten.
2. `make docs` — the plugin guide builds; the new cross-reference resolves to the existing
   "Overriding adapter methods" heading.
3. With a backend running, PUT a chat plugin config with an `llm_config_id` belonging to another
   user — still rejected. GET it back and confirm `llm_config_name`, `llm_provider` and
   `llm_model` are still resolved.

## Notes

- No migration, no new env var, no dependency change, no behaviour change.
- `sparkth/plugins/slack/adapter.py` is the same pass-through subclass. I left it alone to keep
  this PR to the chat stack; happy to make it the next PR rather than leaving the codebase
  teaching two patterns.


sparkth/plugins/chat/tests/test_plugin_adapter.py
```diff
@@ -1,14 +1,30 @@
...
+what is asserted here is that the config pipeline reaches it for this plugin.
+"""
 
+from unittest.mock import AsyncMock, MagicMock
 
-def test_chat_adapter_extends_llm_config_adapter() -> None:
-    assert issubclass(ChatConfigAdapter, LLMConfigAdapter)
+import pytest
 
+from sparkth.core.plugins.service import PluginService
```
hamza-56: This is the same facade rule the PR fixes in the docs: `PluginService` should come from `sparkth.lib.plugins`, not `sparkth.core.*`. Matters extra in a plugin test, since these are meant to travel with the plugin.

hamza-56 (approved):

PR #633, opened 2026-08-26
refactor(plugins): give ChatService the conversation work it owns

abdulrafey1:
> **Stacked on #631** — base is `rafe/chat-drop-config-adapter-subclass`. The chain is
> #625 → #626 → #627 → #629 → #630 → #631 → this. Merge in that order.
>
> First of three PRs relocating what sits in `chat/routes/utils/`. This one takes the four
> functions that read and write conversation rows; the message helpers and the SSE generator
> follow.

## What

Four functions under `routes/utils` did conversation persistence while taking `ChatService` as an
argument. They are service methods now, and the route reads as four service calls.

## Changes

- `refactor(plugins)`: `get_or_create_conversation`, `attach_request_documents`,
  `persist_incoming_messages` and `persist_pre_stream_error` become
  `ChatService.get_or_create_conversation`, `.attach_owned_documents`, `.add_incoming_messages`
  and `.record_error_message`
- `refactor(plugins)`: `ConversationNotFound` replaces the route's `HTTPException(404)`, mapped in
  `plugin.py` via `register_exception_handler`
- `refactor(plugins)`: title-generation scheduling moves to
  `conversation_title.schedule_title_generation`, keeping `BackgroundTasks` out of the service
- `test(plugins)`: 14 tests in `tests/test_service_conversation_flow.py` — these four had no
  unit tests at all

`routes/utils/__init__.py`: 349 → 191 lines, and ten imports went with the moved code.

### Two things that could not move as they were

**The 404.** `get_or_create_conversation` raised `HTTPException(404)`, which the service layer has
no business doing. It raises `ConversationNotFound`, registered to 404 per `CLAUDE.md`'s
"Domain exceptions → HTTP responses". The registration sits at module level in `plugin.py`, not in
`ChatPlugin.__init__`, because a second registration of the same type raises and the plugin class
is instantiated more than once across a test session.

**The background task.** It also took `BackgroundTasks` and built an LLM provider to queue title
generation. That is a route concern, so it moved to `conversation_title.py` beside
`generate_conversation_title` and `extract_title_from_messages`, and the route calls it when it
created the conversation.

### What the new tests cover that nothing did

All four were reached only through route integration tests, so their edges were unasserted: an
unowned document id skipped rather than attached (and logged), another user's conversation
reported as absent rather than resolved, an attachment-only turn stored with a readable body
instead of an empty one, and a database failure while recording an error not replacing the error
already on its way to the user.

## How to Test

1. `make test.backend.pytest sparkth/plugins/chat/` — 358 pass.
   (1919 passed, 3 skipped; the 3 are the `pg` analytics lane).
3. With a backend running: POST a completion with a `conversation_id` that does not exist — still
   404, now with detail `Conversation not found`. Start a new conversation and confirm its title
   is replaced by the generated one a moment later. Send a request with `document_ids` containing
   one id you own and one you do not — only yours attaches, and the skip is logged.

## Notes

- **Response text change:** the completions 404 detail was `Conversation {uuid} not found` and is
  now `Conversation not found`, matching `routes/dependencies.py` and `routes/conversations.py`.
  The registry sends `str(exc)` to the client, and its contract asks for client-safe messages
  without internal identifiers. No test asserted the old text.
- `ConversationNotFound` is now available to the plugin's three other conversation 404s
  (`dependencies.py:28`, `conversations.py:92` and `:151`), which still raise `HTTPException`
  inline. Left alone here to keep this PR to the move.

### DIFF
--- a/sparkth/plugins/chat/routes/completions.py
+++ b/sparkth/plugins/chat/routes/completions.py
@@ -124,26 +124,26 @@ async def chat_completion(
         # Already judged, so the check below would spend a second call on the same message.
         _skip_main_scope_check = True
 
-    conversation = await service.get_or_create_conversation(
+    conversation, conversation_was_created = await service.get_or_create_conversation(
         session,
-        conversation_uuid,
-        user_id,
-        request.llm_config_id,
-        provider_name,
-        model,
-        extract_title_from_messages(request.messages, max_length=config.title_max_length),
+        conversation_uuid=conversation_uuid,
+        user_id=user_id,
+        llm_config_id=request.llm_config_id,
+        provider=provider_name,
+        model=model,
+        title=extract_title_from_messages(request.messages, max_length=config.title_max_length),
     )
-    if conversation_uuid is None:
+    if conversation_was_created:
         schedule_title_generation(
             background_tasks,
             service,
-            cast(int, conversation.id),
-            user_id,
-            request.messages,
-            provider_name,
-            api_key,
-            model,
-            config,
+            conversation_id=cast(int, conversation.id),
+            user_id=user_id,
+            messages=request.messages,
+            provider_name=provider_name,
+            api_key=api_key,
+            model=model,
+            config=config,
         )
 
     if request.document_ids:
--- a/sparkth/plugins/chat/routes/completions.py
+++ b/sparkth/plugins/chat/routes/completions.py
@@ -22,15 +22,12 @@ from sparkth.lib.models import User
 from sparkth.plugins.chat.classifiers import MessageScopeClassifier, RAGSearchClassifier
 from sparkth.plugins.chat.config import ChatSettings, get_chat_settings
 from sparkth.plugins.chat.constants import LLM_PROVIDER_API_ERRORS
+from sparkth.plugins.chat.conversation_title import extract_title_from_messages, schedule_title_generation
 from sparkth.plugins.chat.exceptions import RAGSearchError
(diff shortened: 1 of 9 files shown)


sparkth/plugins/chat/service.py
```diff
@@ -38,6 +56,35 @@ async def create_conversation(
         logger.info("Created conversation %s for user %s", conversation.id, user_id)
         return conversation
 
+    async def get_or_create_conversation(
```
hamza-56, line 60: The helper this replaces was keyword-only on purpose. This and `schedule_title_generation` went fully positional, and the route now passes long runs of same-typed args (`provider, model` here, `provider_name, api_key, model` in the title call). A transposition there type-checks clean and fails ugly.

sparkth/plugins/chat/routes/completions.py
```diff
@@ -126,29 +123,31 @@ async def chat_completion(
...
+        request.llm_config_id,
+        provider_name,
+        model,
+        extract_title_from_messages(request.messages, max_length=config.title_max_length),
     )
-
-    if request.document_ids:
-        await attach_request_documents(
-            session,
+    if conversation_uuid is None:
```
hamza-56: The service just decided create-vs-resolve internally, and the route re-derives that same decision here, so the creation rule now lives in two places. It also means `extract_title_from_messages` runs on every request only to be thrown away whenever the conversation already exists.

hamza-56 (approved):

PR #638, opened 2026-08-28
refactor(plugins): let the chat services decide what is missing

abdulrafey1:
> **Stacked on #636** — base is `rafe/chat-drop-type-checking-tests`. Bottom of the chain is #625;
> merge in stack order.

## What

Four chat routes looked a record up, found `None`, and raised their own `HTTPException(404)`. The
services now answer with an exception and the registry maps it, so the status is decided once.

## Changes

- `refactor(plugins)`: `ChatService.get_user_owned_document` → `require_owned_document`, returning
  the document or raising `DocumentNotFound`
- `refactor(plugins)`: add `ChatService.require_owned_conversation`, raising the
  `ConversationNotFound` already registered in #633
- `refactor(plugins)`: `DocumentNotFound` added to `chat/exceptions.py` and registered to 404 in
  `plugin.py`
- `refactor(plugins)`: `routes/attachments.py` (2 raises), `routes/dependencies.py` (1) and
  `routes/conversations.py` (2) drop their inline raises; `attachments.py` and `dependencies.py` no
  longer import `HTTPException`
- `test(plugins)`: 7 tests in `tests/test_service_ownership.py`
- `refactor(plugins)`: `get_or_create_conversation` resolves its uuid through
  `require_owned_conversation` rather than repeating the lookup, warning and raise
- `chore(i18n)`: regenerate the chat catalogs with `make i18n.update`

### Why the nullable return was the problem

`get_user_owned_document` returned `Document | None` and had exactly two callers, both of which
turned the `None` into the same 404 with the same message. Same for `get_conversation_by_uuid` at
three sites. The status decision was copied five times, which is what `CLAUDE.md`'s
"Domain exceptions → HTTP responses" rule is written to prevent: *"HTTP status decisions live in one
API-layer mapping, not scattered through business-logic routes."*

### One exception for three cases

`DocumentNotFound` covers absent, deleted, and belongs-to-another-user. Distinguishing them would
confirm to a caller that a document exists which they are not allowed to see. Its docstring says so,
and a test pins each of the three.

### What the tests add

The routes only ever asserted a status code, so the cases underneath were unasserted. The service
tests now cover another user's document, a soft-deleted one, an unknown UUID and another user's
conversation directly.

## How to Test

1. `make test.backend.pytest sparkth/plugins/chat/` — 364 pass.
   (1953 passed, 3 skipped; the 3 are the `pg` analytics lane).
3. With a backend running: `POST /api/v1/chat/conversations/{uuid}/attachments` with a
   `document_id` you do not own → still `404 {"detail": "Document not found or not accessible"}`.
   `GET /api/v1/chat/conversations/{unknown-uuid}` → still `404 {"detail": "Conversation not found"}`.
4. `make i18n.update` — re-run it and `git diff sparkth/plugins/chat/locale/` is clean apart
   from the `POT-Creation-Date` header, so the committed location comments are the ones pybabel
   produces. `Conversation not found` now extracts from a single site.

## Notes

- No migration, no new env var, no dependency change, **no change to any response** — same statuses,
  same detail strings, same translations.
- Ten `raise HTTPException` remain in `routes/completions.py`. Those are the legitimate case from the
  same rule: the status is context-dependent (`LLMConfigNotFoundError` → 404 vs
  `LLMConfigInactiveError` → 409 vs `RAGSearchError` → 502), and each is caught and translated at the
  route boundary with a persisted error message. Left alone deliberately.
- `get_conversation_by_uuid` keeps its nullable contract for `record_error_message`, which reads
  the `None` as an answer rather than an error.
- The catalog regeneration reaches further than the two messages this PR adds: the committed
  location comments were hand-written and every `service.py` reference was off by 15 lines, and the
  `routes/utils/__init__.py` entries still named a file #634 had split. `make i18n.update` rewrites
  a catalog whole, so those are corrected here too, and the msgid #633 removed retires to an
  obsolete `#~` entry. `test_catalog_containment.py` compares file paths only, stripping the line
  number, so none of this was failing CI.

### DIFF
--- a/sparkth/plugins/chat/service.py
+++ b/sparkth/plugins/chat/service.py
@@ -9,7 +9,7 @@ from sqlmodel.ext.asyncio.session import AsyncSession
 from sparkth.lib.documents import Document, DocumentStatus
 from sparkth.lib.i18n import _
 from sparkth.lib.log import get_logger
-from sparkth.plugins.chat.exceptions import ConversationNotFound
+from sparkth.plugins.chat.exceptions import ConversationNotFound, DocumentNotFound
 from sparkth.plugins.chat.messages import text_of
 from sparkth.plugins.chat.models import Conversation, ConversationAttachment, Message, MessageType
 from sparkth.plugins.chat.schemas import ChatMessage
@@ -93,6 +93,23 @@ class ChatService:
         )
         result = await session.exec(statement)
         return result.first()
+
+    async def require_owned_conversation(
+        self,
+        session: AsyncSession,
+        conversation_uuid: UUID,
+        user_id: int,
+    ) -> Conversation:
+        """Return the caller's conversation.
+
+        Raises:
+            ConversationNotFound: no such conversation, or it belongs to someone else.
+        """
+        conversation = await self.get_conversation_by_uuid(session, conversation_uuid, user_id)
+        if conversation is None:
+            logger.warning("Conversation %s not found for user %s", conversation_uuid, user_id)
+            raise ConversationNotFound(_("Conversation not found"))
+        return conversation
 
     async def list_conversations(
         self, session: AsyncSession, user_id: int, limit: int = 50, offset: int = 0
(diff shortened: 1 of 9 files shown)


sparkth/plugins/chat/service.py
```diff
@@ -94,6 +94,23 @@ async def get_conversation_by_uuid(self, session: AsyncSession, uuid: UUID, user
         result = await session.exec(statement)
         return result.first()
 
+    async def require_owned_conversation(
```
hamza-56, line 94: `get_or_create_conversation` just above still carries this exact body inlined: same lookup, same warning, same raise. Now that this exists, that branch should call it.

hamza-56 (commented):

PR #639, opened 2026-08-28
refactor(plugins): thin out the chat completion handler

abdulrafey1:
> **Stacked on #638** — base is `rafe/chat-document-not-found-exception`. Bottom of the chain is
> #625; merge in stack order. Last PR in the chat refactor stack.

## What

`chat_completion` was 300 lines, and four of the things it spent them on were written more than
once. It is 218 now, with the duplication gone and one 45-line block extracted and tested.

## Changes

- `refactor(plugins)`: the three `except` blocks for an unusable LLM config become one, sharing the
  log, the persisted message and the raise — `LLMConfigNotFoundError` → 404,
  `LLMConfigModelNotSetError` → 422, `LLMConfigInactiveError` → 409
- `refactor(plugins)`: `_refusal_response` replaces the out-of-scope refusal built twice
- `refactor(plugins)`: message assembly moves to `routes/utils/message_assembly.py`
- `refactor(plugins)`: the two `except` blocks that persist an error and answer 502 become one
- `refactor(plugins)`: `cast(int, conversation.id)` resolved once instead of eight times
- `test(plugins)`: 9 tests in `tests/test_message_assembly.py`, 6 in `tests/test_unusable_llm_config.py`
- `chore(i18n)`: regenerate the chat catalogs with `make i18n.update`

### The unreachable 500 that went away

The old code built the retrieval turn in one branch and then asserted it existed in another:

```python
if request.stream and rag_search_required:
    if unresolved_messages is None:
        raise HTTPException(500, detail=_("Chat completion failed"))
```

Unreachable — `rag_search_required` is only set when there are attached documents, which is the
same condition that builds the turn. `assemble_provider_messages` returns the turn alongside the
messages, so the invariant is structural and there is nothing left to assert. Same shape as the
`cast` that came out of `base.classify` earlier in this stack.

### Why the config failure is decided inline

An earlier revision paired each exception with its status in a module-level dict and matched it with
an `isinstance` scan, to avoid a `KeyError` if one of the three were ever subclassed. The `except`
clause already matches by `isinstance`, so the scan could not fail and its re-raise fallback was
unreachable — machinery guarding a case the clause had handled, and one more hop between the status
and the handler that returns it.

An `if`/`elif` over a literal `except` tuple puts each status and its sentence where the handler
answers, keeping only the tail that really was written three times. `gettext_noop` plus a `gettext`
at raise time collapses back to `_()`, so the catalog references name the handler.

### What message assembly now gets tested for

It was 45 lines inline that nothing exercised directly. The tests cover which end of the history is
replaced (the request's turns win, because storage flattens content blocks and retrieval needs
them), and all three shapes: streaming leaves the references for the stream to resolve,
non-streaming resolves them first, and no-retrieval builds nothing synthetic.

## How to Test

1. `make test.backend.pytest sparkth/plugins/chat/` — 379 pass.
   (1968 passed, 3 skipped; the 3 are the `pg` analytics lane).
3. `tests/test_unusable_llm_config.py` now covers each of the three failures — a `llm_config_id`
   that does not exist (404), one with no model set (422), a deactivated one (409) — asserting the
   status and that the message is persisted, so a reload shows it. No manual pass needed.
4. Send a message with an attached document, streaming and non-streaming, and confirm retrieval
   still runs in both.

## Notes

- **One reply changes.** A refusal in an existing conversation reported `llm_config.model` where
  the pre-conversation path reported the model actually in use. Both now report the model in use —
  the one `model_override` selected. Nothing asserted the old value.
- Two suggestions from the plan are **not** here, deliberately:
  - Extracting the non-streaming tail needs 10 parameters to say what it does. A 10-parameter
    helper is worse than the 30 cohesive lines it would replace.
  - Folding the two scope checks into one changes *when* a classifier call happens, which is
    behaviour. This PR is a restructure with the existing suite as its proof, and that belongs in
    its own change.
- The `LLMConfigNotFoundError`/`ModelNotSet`/`Inactive` trio maps 1-to-1 to statuses, which is the
  exception registry's ideal case. They stay inline because each also persists a message to the
  conversation, which the registry cannot do. Moving them would mean putting client-safe copy on
  the exceptions in `sparkth/llm/` — worth doing, but not as part of a restructure.

### DIFF
--- a/sparkth/plugins/chat/routes/completions.py
+++ b/sparkth/plugins/chat/routes/completions.py
@@ -1,4 +1,5 @@
 from typing import Any, cast
+from uuid import UUID
 
 from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, status
 from fastapi.responses import StreamingResponse
@@ -8,7 +9,7 @@ from sqlmodel.ext.asyncio.session import AsyncSession
 
 from sparkth.lib.auth import get_current_user
 from sparkth.lib.db import get_async_session
-from sparkth.lib.i18n import _, gettext
+from sparkth.lib.i18n import _, gettext, gettext_noop
 from sparkth.lib.llm import (
     LLMConfigInactiveError,
     LLMConfigModelNotSetError,
@@ -28,7 +29,7 @@ from sparkth.plugins.chat.lms_credentials import build_lms_credentials_message
 from sparkth.plugins.chat.messages import get_last_user_text
 from sparkth.plugins.chat.prompt import get_course_design_system_prompt
 from sparkth.plugins.chat.routes.utils import resolve_tools
-from sparkth.plugins.chat.routes.utils.rag_search import resolve_document_blocks
+from sparkth.plugins.chat.routes.utils.message_assembly import assemble_provider_messages
 from sparkth.plugins.chat.routes.utils.stream_processor import (
     ChatStreamProcessor,
     stream_out_of_scope_refusal,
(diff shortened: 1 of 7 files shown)


sparkth/plugins/chat/routes/completions.py
```diff
@@ -47,6 +48,57 @@
...
+        status.HTTP_409_CONFLICT,
+        gettext_noop(
+            "The selected AI key is deactivated. Go to AI Keys to reactivate it, "
+            "or choose a different one in chat settings."
+        ),
+    ),
+}
+
+
+def _unusable_config_failure(exc: Exception) -> tuple[int, str]:
```
hamza-56: More machinery than the three excepts it replaced: a dict, a linear isinstance scan, and a `raise exc` that can never run because the except tuple is built from the same dict. The three targeted excepts were easier to read and grep.

sparkth/plugins/chat/routes/utils/message_assembly.py
```diff
@@ -0,0 +1,66 @@
...
+    if query_text:
+        blocks.append({"type": "text", "text": query_text})
+    return ChatMessage(role="user", content=blocks)
+
+
+def _as_provider_turns(messages: list[ChatMessage]) -> list[dict[str, Any]]:
+    return [{"role": message.role, "content": message.content} for message in messages]
+
+
+async def assemble_provider_messages(
```
hamza-56, line 33: Six positionals with a bare bool and str in the middle, and the route passes them all positionally. Same concern as on #633: keyword-only please. Also only `messages` and `stream` are read off `request`, so taking the whole request widens the coupling for nothing.

hamza-56 (commented):
Few comments:

sparkth/plugins/chat/routes/utils/message_assembly.py
abdulrafey1: We should only define keyword arguments if necessary. i dont think its necessary here.

PR #640, opened 2026-08-28
feat(plugins): name the user in every classifier decision logged

abdulrafey1:
> **Stacked on #639** — base is `rafe/chat-completions-handler`. Bottom of the chain is #625; merge
> in stack order.
>
> Addresses [hamza-56's review comment on #625](https://github.com/edly-io/sparkth/pull/625#discussion_r1)
> — see "Why this is not a change to #625" below.

## What

A refusal on the first message of a chat is judged before any conversation row exists, so its log
line read `conversation_uuid=None` and carried nothing else. The one case the logging was added for
was the one it could not trace.

## Changes

- `feat(plugins)`: `MessageScopeClassifier` and `RAGSearchClassifier` take the user id and log it
- `feat(plugins)`: three log lines gain `user_id` — a scope refusal, a scope classifier failing
  open, and a declined search
- `feat(plugins)`: the fail-open line gains `conversation_uuid` too
- `test(plugins)`: 3 tests, one per line

```
Scope classifier refused a message: user_id=42 conversation_uuid=None model=claude-haiku-4-5
reason='general knowledge question' history_turns=0 attachments=0 query_len=27
```

### Why the user id is constructor state

It does not vary within a request, and both classifiers are already built per request from that
user's provider and key — so it belongs beside them rather than as a fourth argument on every call.
It never reaches a model, which the constructor docstrings say. `BaseClassifier` is untouched: this
is logging context, not classification input, and only the two subclasses that log need it.

### The fail-open line

`"Message scope classifier failed, defaulting to in_scope=True: %s"` named neither the user nor the
thread. A turn let through *unjudged* is at least as worth finding later as one that was refused, so
it now carries both.

### Why this is not a change to #625

The comment is on `completions.py:107` in #625, and that code is deleted twice over by the stack
above it: #626 removes the keyword filter and its log entirely, and #629 moves the classifier
logging into `MessageScopeClassifier.in_scope`. Patching #625 would add `user_id` to a log line that
no longer exists by the time the stack lands, and the gap would still be at the tip. It is also
broader than the original comment: at the tip *none* of the three classifier log lines carried a
user id, not only the first-message one.

## How to Test

1. `make test.backend.pytest sparkth/plugins/chat/` — 378 pass.
   (1939 passed, 3 skipped; the 3 are the `pg` analytics lane).
3. With a backend running, send `what is the capital of France?` as the **first** message of a new
   chat. The refusal log now names the user, where before it carried only
   `conversation_uuid=None`.

## Notes

- Logging only: no behaviour change, no API change, no new dependency.
- A user id is an internal identifier and is already logged elsewhere in the plugin
  (`attach_owned_documents`). The existing discipline holds: no message text, counts and lengths
  only, and the model-written reason is constrained to a category.

### DIFF
--- a/sparkth/plugins/chat/classifiers/message_scope.py
+++ b/sparkth/plugins/chat/classifiers/message_scope.py
@@ -23,13 +23,17 @@ logger = get_logger(__name__)
 class MessageScopeClassifier(BaseClassifier[MessageScopeInput, MessageScopeVerdict]):
     """Decides whether a chat turn falls within the assistant's learning-design scope."""
 
-    def __init__(self, provider_name: str, api_key: str) -> None:
+    def __init__(self, provider_name: str, api_key: str, user_id: int) -> None:
+        """``user_id`` never reaches the model. It is logged on every decision this classifier
+        records, because the first message of a chat is judged before a conversation exists and the
+        user is then the only thing a refusal can be traced by."""
         super().__init__(
             MESSAGE_SCOPE_CLASSIFIER_SYSTEM_PROMPT,
             MessageScopeVerdict,
             provider_name,
             api_key,
         )
+        self._user_id = user_id
 
     def _build_messages(self, payload: MessageScopeInput) -> list[BaseMessage]:
         """Replay the recent conversation as real turns, then the current message.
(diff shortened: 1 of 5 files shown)


sparkth/plugins/chat/classifiers/message_scope.py
```diff
@@ -23,13 +23,17 @@
 class MessageScopeClassifier(BaseClassifier[MessageScopeInput, MessageScopeVerdict]):
     """Decides whether a chat turn falls within the assistant's learning-design scope."""
 
-    def __init__(self, provider_name: str, api_key: str) -> None:
+    def __init__(self, provider_name: str, api_key: str, user_id: int) -> None:
```
hamza-56, line 26: `conversation_uuid` rides on each `in_scope` call but `user_id` is baked into the constructor, so the two correlation ids on the same log line come in through different doors. Pick one. And if it stays on the instance, `_user_id` belongs on `BaseClassifier` rather than copy-pasted into both subclasses.

hamza-56 (commented):
Few comments:

PR #641, opened 2026-09-08
feat(lib): add the LMS content contributor hook

abdulrafey1:
## What

Adds a named *content contributor* hook in `sparkth/lib`, so a plugin that owns content can hand it to a plugin that publishes to an LMS without either importing the other.

This is PR 1 of 8 implementing the PXC integration's first slice. The design doc's D3 calls for insertion to happen through a contributor hook in `sparkth/lib`, fired by a new Open edX tool, so that neither plugin imports the other — `pxc` can be removed or replaced without touching `openedx`, and a future Canvas plugin can consume the same hook unchanged. It is the pattern issue #372 points at, applied to one concrete case rather than designed in the abstract.

Nothing consumes the hook in this PR; PR 2 adds the first consumer.

## Changes

- `feat(lib)`: add `ContentBlock`, `ContentContributor` and the `LMS_CONTENT_CONTRIBUTORS` hook in `sparkth/lib/content/hooks.py`
- `feat(lib)`: relax `HasName` on `SingleNamedItemHook` to a read-only property so a frozen dataclass satisfies it
- `docs`: document contributors in `docs/guides/plugins.md` and add the module to `docs/reference/plugins.md`

`ContentContributor.build` is a callable field rather than a method to override, mirroring `Tool(handler)` on the `MCP_TOOLS` hook.

The `HasName` change is the one edit outside `sparkth/lib/content/`. It is required, not cosmetic: `SingleNamedItemHook`'s type bound declared `name: str` as a mutable attribute, which a frozen dataclass does not satisfy under `mypy --strict` (`Type argument … must be a subtype of HasNameAttr`). A read-only property is satisfied by frozen dataclasses *and* by the existing mutable-attribute items, so `PERMISSIONS` and `PERMISSION_SCOPES` are unaffected — verified against both.

## How to Test

1. `make test.backend` — the four new tests in `tests/lib/content/test_hooks.py` cover resolution by name, a miss, the block a contributor builds, and the duplicate-name rejection.
2. `make docs` and confirm `sparkth.lib.content.hooks` appears in the generated reference:
   `grep -c LMS_CONTENT_CONTRIBUTORS site/reference/plugins/index.html`
   A green `make docs` alone is not evidence — `mkdocs --strict` does not fail on a module simply absent from the reference.
3. Confirm the permission hooks still typecheck under the widened bound: `make mypy`.

## Notes

Registration must happen at **module level in a plugin's `plugin.py`** — not in the package `__init__.py` and not in `SparkthPlugin.__init__`. `SingleNamedItemHook.add_item` raises on a duplicate name, and both of those locations run more than once per pytest session. This is documented in the module docstring and in the guide.

Stacked PR 1 of 8. Merge in order; PR 2 targets this branch.

---

### DIFF
--- a/sparkth/lib/content/hooks.py
+++ b/sparkth/lib/content/hooks.py
@@ -1,0 +1,56 @@
+"""Hook through which a plugin contributes content another plugin publishes to an LMS.
+
+A *content contributor* is a named producer of one LMS content block. A plugin that owns
+content registers one; a plugin that publishes to an LMS resolves it **by name** and awaits
+its ``build``. Neither plugin imports the other, so the producer can be removed or replaced
+without touching the publisher, and a second publisher (Canvas) can consume the same hook
+with no change to the producer.
+
+A plugin registers at module level in its ``plugin.py``::
+
+    LMS_CONTENT_CONTRIBUTORS.add_item(ContentContributor("pxc", "...", build_pxc_block))
+
+Module level, not the package ``__init__`` and not ``SparkthPlugin.__init__``: this is a flat
+hook keyed by name, and both of those run more than once in a pytest session, which would
+raise on the second registration.
+"""
+
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass, field
+
+from sparkth.lib.hooks import SingleNamedItemHook
+
+
+@dataclass(frozen=True)
+class ContentBlock:
+    """One content block to create in an LMS course.
+
+    ``category`` is the LMS's block type (Open edX calls it the XBlock category, and it must
+    appear in the course's Advanced Module List for anything but the built-in types).
+    ``settings`` is applied to the block after creation; every value is carried as a string
+    because that is what an XBlock's metadata holds.
+    """
+
+    category: str
+    display_name: str
+    settings: dict[str, str] = field(default_factory=dict)
+
+
+@dataclass(frozen=True)
+class ContentContributor:
+    """A named producer of one :class:`ContentBlock`.
+
+    ``build`` is awaited with the destination course id and returns the block to create. It is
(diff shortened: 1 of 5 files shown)


sparkth/lib/content/hooks.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    name: str
+    description: str
+    build: Callable[[str], Awaitable[ContentBlock]]
+
+
+# Content contributors, keyed by name. Consumed by LMS-publishing plugins
+# (sparkth/plugins/openedx/tools.py).
+LMS_CONTENT_CONTRIBUTORS: SingleNamedItemHook[ContentContributor] = SingleNamedItemHook()
```
hamza-56, line 56: Having to register at module level (and the tests poking `_items` for teardown) is a workaround for the hook shape, and it's now leaking into the lib docs as a rule. Every other hook is registered from `SparkthPlugin.__init__` because it's keyed by plugin instance. A `PluginCollectionHook[ContentContributor]` resolved by name, or a `remove()` / idempotent add on `SingleNamedItemHook`, would drop the caveat entirely.

hamza-56 (commented):
One design point on the hook shape, otherwise fine.

sparkth/lib/content/hooks.py
abdulrafey1: Added a `register_content_contributor` function just like `register_event_schema`, which checks the duplicate entry.

Also added a `remove` method to the `SingleNamedItemHook`.

abdulrafey1:
@claude-review

hamza-56 (approved):
Thanks, the helper plus `remove()` is exactly what I was after. LGTM.

PR #642, opened 2026-09-08
feat(openedx): publish plugin-contributed content into a unit

abdulrafey1:
## What

Adds the two Open edX MCP tools that consume the content contributor hook, so an agent publishing a course can insert a block contributed by another plugin without `openedx` knowing anything about that plugin.

This is PR 2 of 8 implementing the PXC integration's first slice. It is the first consumer of PR 1's hook, and the proof that the boundary works: the tests exercise it against a *fake* contributor with no PXC knowledge at all.

## Changes

- `feat(openedx)`: add `openedx_add_plugin_content`, which resolves a contributor by name, awaits its block, creates it via the existing Studio path and patches its settings on
- `feat(openedx)`: add `openedx_list_content_contributors` so an agent can discover what is publishable
- `feat(openedx)`: add `AddPluginContentArgs` to `schemas.py` and register both tools in a new `openedx-content` category
- `feat(lib)`: add `ContentBuildError`, in `sparkth/lib/content/exceptions.py`
- `docs`: document the raise contract in `docs/guides/plugins.md`

Both tools reuse `openedx_create_basic_component` and `openedx_update_xblock_content` rather than adding a third Studio call surface.

Failure is reported by **returning** `{"error": {...}}` via the module's `_lms_error` helper, never by raising — matching every other tool in that file. `CLAUDE.md`'s raise-and-map-to-HTTP rule governs REST routes, not MCP tools.

`ContentBuildError` exists because `openedx_add_plugin_content` cannot know what an arbitrary contributor's storage layer raises, and the project forbids a bare `except Exception`. A contributor reports build failure by raising exactly that type; the tool converts it to its own error contract. It lands in this PR rather than PR 1 because the contract was discovered by writing the consumer.

## How to Test

1. `make test.backend` — `sparkth/plugins/openedx/tests/test_content_tools.py` covers the happy path, an unknown contributor name, a Studio failure, a contributor whose `build` raises, and a block with no settings (which must skip the patch call).
2. Confirm the boundary is real, not nominal: neither plugin imports the other.
   `grep -rn "plugins.pxc" sparkth/plugins/openedx/` and `grep -rn "plugins.openedx" sparkth/plugins/pxc/` — both return nothing.
3. Confirm `build` is awaited with the destination course id, not some other identifier — asserted in `test_creates_the_block_the_contributor_declares`.

## Notes

The published block's category must appear in the destination course's Advanced Module List, and an XBlock providing it must be installed in the Open edX instance. Publishing to an instance without it fails. That XBlock arrives in PR 8.

Three error branches are reachable but untested — `ValueError` from the create helper, and both failure types from the update helper. All three return the identical `_lms_error` shape that `test_a_studio_failure_is_reported_as_an_error_dict` already exercises, so the final review judged them a coverage gap rather than a risk.

Stacked PR 2 of 8, targeting `rafe/pxc-content-hook`. Merge in order.

---


hamza-56 (commented):
One thing on partial failure.

sparkth/plugins/openedx/tools.py
```diff
@@ -703,3 +706,78 @@ async def openedx_get_block_contentstore(payload: BlockContentArgs) -> dict[str,
...
+    except LMSRequestError as err:
+        return _lms_error(err, method=err.method, endpoint=err.url)
+    except ValueError as err:
+        return {"error": {"message": str(err)}}
+
+    if not block.settings:
+        return {"response": {"locator": locator, "category": block.category}}
+
+    try:
+        await openedx_update_xblock_content(payload.auth, payload.course_id, locator, None, block.settings)
```
hamza-56, line 777: If this patch fails the block already exists in Studio, but the caller only gets an error: no locator to clean up, and a block with no `placement` that will never render (plus seeded state nobody references). Worth a partial-failure story here. Also the two identical except pairs could be one try.

abdulrafey1:
@claude-code

abdulrafey1:
@claude-review

hamza-56 (approved):
Locator on the partial failure and the single try both look good. LGTM.

PR #643, opened 2026-09-08
feat(pxc): add the plugin skeleton and its sandbox build lanes

abdulrafey1:
## What

Adds the `pxc` plugin's foundations: the pinned `pxc-lib` dependency, the bundled sample activity, the WebAssembly build lanes for its sandbox, and the plugin's constants, domain exceptions and activity index.

This is PR 3 of 8 implementing the PXC integration's first slice. Nothing serves a learner yet — this PR makes the plugin exist, loadable, and buildable.

## Changes

- `build`: add `pxc-lib` as a git dependency pinned to an exact rev, plus the `[tool.hatch.metadata] allow-direct-references` that a direct reference requires
- `build(pxc)`: bundle the `mcq` sample activity from the pinned PXC rev and add `make pxc.activities.build`, the npm toolchain manifests beside it, and the `.gitignore` entries for its build products
- `feat(pxc)`: add `constants.py`, `exceptions.py` and `activities.py`, which resolves an activity type by its manifest `name` to a source directory and a state file
- `feat(pxc)`: add the plugin class, register it in `PLUGINS`, and map its six domain exceptions to statuses at module level
- `fix(pxc)`: reject a path-escaping activity name in `state_file`
- `ci`: build the sample activity before the backend tests, and add a Docker stage that builds it for the image
- `chore(pxc)`: register the `wasm` pytest marker and skip marked tests when the binary is absent

The dependency rev is pinned deliberately: `edly-io/pxc` moves, and an unpinned git dependency would break CI at random. Bumping it should be a deliberate commit.

`state_file` resolves its name through the activity index before building a path, so an unknown or traversal-shaped name raises `PxcActivityNotFound` rather than yielding a path outside the data directory. Two later PRs call it with a name derived from a launch token or an environment variable, so the check belongs in the module that owns the invariant rather than in each caller.

Bundled activities live under `sparkth/plugins/pxc/activities/`, one directory per type, and the index is scoped to that directory rather than globbing the whole plugin package. The npm toolchain sits inside the activity it builds rather than at the repository root. Plugins are meant to be self-contained and to eventually move to their own repositories, and `frontend/` already owns its manifests the same way. The activity's own Makefile gates `npm ci` on the lockfile, which is what makes step 1 below a genuine no-op on a second run: `npm ci` deletes and reinstalls `node_modules` unconditionally, so an ungated install leaves the target doing work forever.

The Docker builder stage gains `git`: `uv sync --frozen` has to clone the pinned dependency, and that image ships none.

## How to Test

1. `make pxc.activities.build` then `test -s sparkth/plugins/pxc/activities/mcq/sandbox.wasm`. Run it twice — the second run must be a no-op.
2. `git status --porcelain` — no build products appear as untracked, proving the ignore entries work rather than merely existing.
3. `make test.backend`. Note the skip count *drops* by one: `test_a_plugin_marking_strings_owns_catalogs` stops skipping the moment this plugin's `__init__` contains `sparkth.lib.i18n`, and then asserts the catalogs exist.
4. `make docker.build` — must succeed. This is the check that proves both the `git` addition and that the new locale catalogs satisfy the image's `pybabel compile` loop; an empty `locale/` directory fails that loop.
5. Confirm the plugin actually loads: `uv run python -c "import sparkth.main"` and look for `pxc` in the "Loaded N plugin(s)" line. Plugins are not auto-discovered — a missing `PLUGINS` entry is silent, and the loader swallows construction failures with a log.

## Notes

**New env vars.** `PXC_DATA_DIR` (default `./data/pxc`) and `PXC_LAUNCH_TOKEN_TTL_SECONDS` (default `300`) are in `.env`. `PXC_LAUNCH_SECRET` is **sensitive** and has no value in `.env` — only a reference in the must-change block. It belongs in `.env.local` and must match the value configured on the Open edX side.

**New dependency.** `pxc-lib` pulls in `jsonschema`, `pydantic` and `wasmtime`. Verified installable and importable on this project's Python 3.14: `wasmtime` publishes `py3-none-*` wheels, so there is no ABI-tag problem.

**Known limits carried by this PR.** State files are node-local, so multiple replicas need shared storage (L3), and every learner of one activity type writes to one SQLite file, so it needs WAL and short transactions (L4). One bundled activity is served for every request by design; there is no authoring UI (L5).

`make i18n.init` must never be used to add a single plugin's catalogs — it loops `pybabel init` over every catalog directory in the repo and silently overwrites existing translations. Use a scoped `pybabel init`, and note the `mkdir` must precede `make i18n.extract`, because `I18N_CATALOG_DIRS` is a wildcard over directories that already exist.

Stacked PR 3 of 8, targeting `rafe/pxc-openedx-tools`. Merge in order.

---


sparkth/plugins/pxc/constants.py
```diff
@@ -0,0 +1,18 @@
+import os
+from pathlib import Path
+
+# Where the per-activity-type SQLite state files and the activities' file storage live.
+# One file per activity type holds every course and every learner for that type (D1).
+PXC_DATA_DIR = Path(os.getenv("PXC_DATA_DIR", "./data/pxc"))
+
+# The activity sources bundled inside this plugin. One sample ships today and is served for
+# every request; an authoring module comes later.
+PXC_ACTIVITY_ROOT = Path(__file__).parent
```
hamza-56, line 10: This makes the whole plugin package the activity root, so `index_activities` globs `*/manifest.json` across `tests/`, `locale/`, `static/` too, and any stray manifest.json becomes a bundled activity. A dedicated `activities/<name>/` root would be tighter, and would fix the naming too (a dir called `activity` holding an activity called `mcq`).

Makefile
```diff
@@ -133,6 +133,11 @@ cli: ## Run CLI tool (make cli -- users --help)
 mypy: ## Run mypy type checking
 	uv run mypy --strict sparkth/ tests/ scripts/
 
+.PHONY: pxc.activity.build
+pxc.activity.build: ## Build the bundled PXC sample activity's WASM sandbox
+	npm ci
```
hamza-56, line 138: A root `package.json` and `node_modules` for one plugin's build cuts against plugins being self-contained (they're meant to move to their own repos). This belongs under `sparkth/plugins/pxc/activity/`. Also `npm ci` wipes and reinstalls node_modules every time, so the second-run-is-a-no-op claim only holds for the wasm step.

hamza-56 (commented):
Two structural things, both easier to change now than later.

abdulrafey1:
@claude-review

sparkth/plugins/pxc/constants.py
```diff
@@ -0,0 +1,18 @@
...
+PXC_DATA_DIR = Path(os.getenv("PXC_DATA_DIR", "./data/pxc"))
+
+# Where the bundled activity sources live, one directory per activity type. One sample ships
+# today and is served for every request; an authoring module comes later.
+PXC_ACTIVITY_ROOT = Path(__file__).parent / "activities"
+
+# Shared with the Open edX XBlock. The XBlock signs a launch token with it and this plugin
+# verifies that signature, which is what makes the learner's identity trustworthy rather than
+# client-asserted. Sensitive: set it in .env.local, never in .env.
+PXC_LAUNCH_SECRET = os.getenv("PXC_LAUNCH_SECRET", "")
```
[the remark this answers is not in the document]
abdulrafey1: This is taken care of in PR #650

hamza-56 (approved):
LGTM.

PR #644, opened 2026-09-08
feat(pxc): store activity fields in the activity type's sqlite file

abdulrafey1:
## What

Implements `SqliteFieldStore`, the plugin's version of `pxc-lib`'s `FieldStore` interface, over one SQLite file per activity type.

This is PR 4 of 8 implementing the PXC integration's first slice. It is the storage half of "Sparkth owns the activity's data"; PR 5 hands it to the PXC runtime.

## Changes

- `feat(pxc)`: implement all 11 `FieldStore` methods — four scalar, seven log — in `sparkth/plugins/pxc/field_store/sqlite.py`
- `refactor(pxc)`: make `field_store` a package, one module per storage engine, so a second backend has somewhere to go
- `refactor(pxc)`: type-check the store against the interface it implements, instead of an inline ignore
- `perf(pxc)`: set the journal mode once in `__init__` rather than on every connection
- `test(pxc)`: 18 tests, including the two branches a mutation test proved unguarded

The file is the activity type's boundary (design doc D1): one file holds every course the type appears in and every learner who answered, keyed the way PXC's runtime already keys state — by course, placement and learner. There is no Alembic migration, because these tables are not part of the application schema.

Three decisions worth naming, each with a test that pins it:

**Values are JSON-encoded rather than stored in typed columns.** It keeps arrays and objects in one column alongside the scalars, which `test_structured_values_survive_a_round_trip` pins.

**Connections are wrapped in `contextlib.closing`.** `sqlite3.Connection.__exit__` commits or rolls back but never closes, and the runtime calls `get_field`/`set_field` on every sandbox host call — so a per-operation helper that never closes leaks a connection per call.

**Journal mode is set once, in `__init__`.** WAL is a property of the file rather than of a connection, so running the pragma on every connect repeated work the first connection had already done for good. `test_the_file_is_opened_in_wal_mode` asserts the pragma through an independent connection rather than looking for a `-wal` sidecar: the store closes every connection it opens, SQLite removes that file on a clean last close, and a file check would be a race.

`SqliteFieldStore` is checked against the interface it implements rather than carrying an inline ignore. `pxc-lib` is fully annotated and only missing a `py.typed` marker, so a `follow_untyped_imports` override on its two field modules turns the check on for the same number of lines a `disallow_subclassing_any` override would have turned it off. The override names those two modules rather than all of `pxc.*`, because following the whole package surfaces an unrelated duck-typed socket in a later PR that would itself need an ignore.

The connection is still opened per operation. The store is built per request and handed to a runtime the routes drive through two separate `asyncio.to_thread` calls, which the default executor does not pin to one thread, so a cached connection would raise `sqlite3.ProgrammingError` as soon as two learners act at once.

## How to Test

1. `make test.backend` — 18 tests in `sparkth/plugins/pxc/tests/test_field_store_sqlite.py`.
2. Confirm the interface is fully implemented, not stubbed:
   `uv run python -c "from sparkth.plugins.pxc.field_store import SqliteFieldStore; from pxc.lib.field_store import FieldStore; print(sorted(FieldStore.__abstractmethods__ - set(dir(SqliteFieldStore))))"` prints `[]`, and `grep -rn NotImplementedError sparkth/plugins/pxc/field_store/` prints nothing.
3. Behaviour should match `pxc-lib`'s own `MemoryKVStore` reference implementation — ascending `log_get_after`, descending `log_get_before`, the `{"id", "value"}` entry shape, and the composite `keys()` string format. That equivalence is what lets PXC's runtime consume this store unmodified.

## Notes

Log entry ids come from `AUTOINCREMENT`, so they are strictly increasing but sparse within any one log. Nothing compares them for adjacency — the two cursor methods only order by them — which is the same guarantee PXC's own notebook store gives.

`keys()` is implemented for interface parity and returns composite key strings matching `MemoryKVStore`'s format. `ActivityRuntime` never calls it.

**Known limits carried by this PR.** Globally-scoped fields end up isolated per activity type, because the file is the activity type's boundary (L2) — the bundled sample uses activity scope only. State files are node-local (L3), and one file per activity type under concurrent write needs WAL and short transactions (L4).

Stacked PR 4 of 8, targeting `rafe/pxc-plugin-skeleton`. Merge in order.

---


sparkth/plugins/pxc/field_store.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def _log_scope(
+    course_id: str, activity_name: str, activity_id: str, user_id: str, key: str
+) -> tuple[str, str, str, str, str]:
+    """The five-column log key, as a parameter tuple."""
+    return (course_id, activity_name, activity_id, user_id, key)
+
+
+class SqliteFieldStore(FieldStore):  # type: ignore[misc]  # pxc-lib ships no py.typed marker
```
hamza-56, line 63: New `type: ignore`, which reviewers have been strict about on other PRs. A per-module mypy override (`disallow_subclassing_any = false` for this file) says the same thing without the inline ignore.

sparkth/plugins/pxc/field_store.py
```diff
@@ -0,0 +1,277 @@
...
+    """Field persistence for one activity type, in one SQLite file."""
+
+    def __init__(self, path: Path) -> None:
+        self._path = path
+        path.parent.mkdir(parents=True, exist_ok=True)
+        with self._connect() as connection:
+            connection.executescript(_SCHEMA)
+
+    def _connect(self) -> closing[sqlite3.Connection]:
```
hamza-56, line 72: This opens a connection, runs two pragmas and closes for every `get`/`set`, and `__init__` reruns the DDL on every construction. A store is built per request (runtime, contributor, asset route) and used from one thread, so one action ends up doing a dozen `sqlite3.connect` calls. One connection per store, opened lazily and closed with the runtime, does the same job. WAL is persistent in the file so it only needs setting once anyway.

hamza-56 (commented):
Store logic looks right, two comments on how it's wired.

sparkth/plugins/pxc/field_store/sqlite.py
```diff
@@ -0,0 +1,277 @@
...
+class SqliteFieldStore(FieldStore):  # type: ignore[misc]  # pxc-lib ships no py.typed marker
+    """Field persistence for one activity type, in one SQLite file."""
+
+    def __init__(self, path: Path) -> None:
+        self._path = path
+        path.parent.mkdir(parents=True, exist_ok=True)
+        with self._connect() as connection:
+            connection.executescript(_SCHEMA)
+
+    def _connect(self) -> closing[sqlite3.Connection]:
```
abdulrafey1: The per-operation connect and the repeated DDL are real, but "used from one thread" does not hold. `routes.py` drives each request through two separate `asyncio.to_thread` calls — `build_runtime` constructs the store, `run_action` / `read_state` uses it — and the default executor does not pin those to one thread. Four concurrent requests:

```
r0: SAME thread -> ['asyncio_0']
r1: SAME thread -> ['asyncio_0']
r2: DIFFERENT threads -> ['asyncio_1', 'asyncio_2']
r3: SAME thread -> ['asyncio_0']
```

A connection cached in `__init__` then fails on the second call:

```
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same
thread. The object was created in thread id 8529616256 and this is thread id 6173552640.
```

That looks fine under sequential load and breaks the moment two learners act at once. Making it safe needs `check_same_thread=False`, a lock, and a close hook the runtime does not have — more machinery than the per-operation connect it replaces, for a `sqlite3.connect` on a local file that is noise next to the WASM sandbox call it serves.

Took the part that stands on its own: `PRAGMA journal_mode=WAL` is a property of the file, not the connection, so it now runs once in `__init__` instead of on every connect. `busy_timeout` stays per-connection because it has to. `_connect`'s docstring now records why a connection is not cached, rather than only that one is not shared.

Left `CREATE TABLE IF NOT EXISTS` per construction alone — it is a schema lookup, and the same cost argument applies.

abdulrafey1:
@claude-review

hamza-56 (approved):
LGTM.

sparkth/plugins/pxc/field_store/sqlite.py
hamza-56: Fair, the executor threads kill the cached-connection idea. WAL once in __init__ is the right split.

PR #645, opened 2026-09-08
feat(pxc): verify launch tokens and assemble the activity runtime

abdulrafey1:
## What

Adds launch-token verification and runtime assembly: how a learner's identity reaches Sparkth from Open edX, and how a PXC `ActivityRuntime` is built for one verified launch.

This is PR 5 of 8 implementing the PXC integration's first slice, and it is the security boundary of the whole integration. Identity comes entirely from the token — Sparkth never generates or looks up a user.

## Changes

- `feat(pxc)`: verify the launch token in `tokens.py`, and mint one for tests
- `refactor(pxc)`: sign launch tokens with pyjwt instead of by hand
- `feat(pxc)`: assemble the runtime in `runtime.py` — `build_runtime`, `read_state`, `run_action`
- `docs(pxc)`: document the dormant `SandboxRuntimeError` clause in `run_action`
- `fix(pxc)`: log a warning on a signature mismatch

The token is an HS256 JWT. `pyjwt` was already a direct dependency, so the signing, the timing-safe comparison and the expiry check come from a library audited for them rather than from thirty lines here and thirty more in the XBlock. The claim names are the plugin's own — `act`, `plc`, `cid`, `uid` — because none of the registered JWT claims describe a placement or an activity type.

Four properties, each with a test that fails if it breaks:

- **An empty secret fails closed.** With `PXC_LAUNCH_SECRET` unset, every token is refused and logged at `error`. It never verifies against `""`.
- **`exp` is required, not merely honoured when present.** A launch token without an expiry would be a permanent credential sitting in a URL, so a token that omits the claim is refused rather than treated as having no deadline.
- **The algorithm is pinned to HS256.** Introducing a JWT header also introduces the `alg: none` attack, where the header asks the verifier to skip the signature. Passing an explicit allowlist to the decode is what closes it, and the same allowlist refuses an asymmetric algorithm that would verify the shared secret as a public key.
- **Permission is not a claim.** The token carries activity, placement, course and user only; permission is fixed to `play` where the runtime is built, so a caller can never ask for `edit` by editing a token.

`pyjwt` does not cover every case it looks like it covers. It coerces `exp` with `int()`, so a validly signed `"exp": 1e400` — an IEEE-754 double overflowing to infinity — raises `OverflowError`, which is not an `InvalidTokenError` and would escape as a 500 where every other bad token is a 401. `read_launch_token` catches it alongside pyjwt's own errors, and a test pins it.

A bad signature is logged at `warning` with a best-effort, explicitly unverified activity/placement hint. The most likely production cause is a shared-secret mismatch between Sparkth and the XBlock, which otherwise leaves an operator with nothing but learner-reported 401s. The hint is never used to authenticate anything, and one test asserts it appears for a decodable payload while another asserts the message is exactly the bare prefix when the payload cannot be decoded — a substring check would pass either way.

`pxc.lib.signing` is deliberately unused. It reads `PXC_SIGNING_SECRET` from the environment, warns when unset, then **unconditionally** overwrites it with `b"dev-insecure-default"` — so every token it signs is signed with a public constant. This plugin signs its own.

## How to Test

1. `make test.backend` — 13 token tests and 7 runtime tests.
2. Exercise both `wasm`-marker states, which this PR is the first to use:
   `uv run pytest sparkth/plugins/pxc/tests/test_runtime.py -v` → 7 passed with the binary built; move `sparkth/plugins/pxc/activities/mcq/sandbox.wasm` aside with `mv` and rerun → 4 passed, 3 **skipped**, not errored; then restore it.
3. Confirm the fail-closed path: with `PXC_LAUNCH_SECRET` empty, `read_launch_token` refuses every token.

## Notes

No migration, no breaking change, no new dependency — `pyjwt[crypto]` was already declared and is already used by `sparkth/core/security.py` and `sparkth/lib/auth.py`. `PXC_LAUNCH_SECRET` and `PXC_LAUNCH_TOKEN_TTL_SECONDS` were added to `.env` in PR 3.

**The wire format is shared.** The XBlock in PR 8 mints what this verifies, and the two are separate distributions on separate servers that cannot import each other. They must change together; that half is committed on `rafe/pxc-xblock`.

Everything in `runtime.py` is synchronous and CPU-bound, because the sandbox executes WebAssembly on the calling thread. PR 6 calls it from a worker thread; nothing here should become `async`.

**New known limit (L9).** A sandbox crash *during an action* is invisible to the learner: `pxc-lib`'s `ActivityRuntime.on_action` catches `SandboxRuntimeError` and only logs it, carrying the upstream comment "It's OK to ignore on-action errors, but we should report errors to the frontend". So a crashed action yields HTTP 200 with no events and no error rather than a 502. `get_state` does re-raise, so `read_state`'s equivalent clause is live. `run_action`'s `except` clause is therefore **dormant, not dead** — kept because it becomes correct the moment upstream acts on that comment, and documented as such so nobody deletes it.

Worth reporting upstream alongside L8 (`pxc.lib.signing`).

Non-string token claims are coerced by `str()`, so a `"uid": null` becomes the literal `"None"` rather than being rejected. It is secret-gated, never crashes, and fails safely downstream into a 404 — the final review judged it correctly deferrable.

Stacked PR 5 of 8, targeting `rafe/pxc-field-store`. Merge in order.

---


sparkth/plugins/pxc/tokens.py
```diff
@@ -0,0 +1,139 @@
...
+        return ""
+    if not isinstance(claims, dict):
+        return ""
+    activity, placement = claims.get("act"), claims.get("plc")
+    if activity is None and placement is None:
+        return ""
+    return " (unverified activity=%s, placement=%s)" % (activity, placement)
+
+
+def mint_launch_token(
```
hamza-56, line 74: This is a hand-rolled HS256 JWT with custom claim names. `pyjwt` is already a dependency and covers exp handling, timing-safe compare and the non-ASCII edge case for free, and since Open edX ships pyjwt too the XBlock side would stop needing its own copy and the wire-format pinning tests. Separately, `LaunchClaims` exists but this takes the same fields as 5 positionals (7 by #651).

hamza-56 (commented):
The security properties look right. One comment on the token format itself.

hamza-56 (approved):
LGTM.

PR #646, opened 2026-09-08
feat(pxc): serve activities to learners over a launch-token surface

abdulrafey1:
## What

Adds the learner-facing HTTP surface — the routes a learner's browser actually hits from inside an Open edX course — and the transport client that drives them.

This is PR 6 of 8 implementing the PXC integration's first slice. It is the PR that makes the slice reachable by a human.

## Changes

- `feat(pxc)`: add five routes in `routes.py` — the embed shell, the activity configuration, client scripts, declared assets, and action submission
- `feat(pxc)`: add `schemas.py` and `static/sparkth-pxc.js`, an HTTP-transport subclass of PXC's own `PXC` element
- `fix(pxc)`: unmark a non-sandbox test and cover the embed token guard
- `fix(pxc)`: carry the launch token through the action and asset URLs
- `fix(pxc)`: reject a malformed action body as 422 rather than 500
- `chore(frontend)`: regenerate `generated.ts` for the new routes

**Five routes, not the design doc's four.** Its list omits a URL for `pxc.js` itself, which the browser must fetch. Serving it through the asset route would make that route both serve package files and enforce the manifest's declared-asset allowlist — two jobs. A dedicated `/client/{file_name}` route keeps each to one. Worth confirming against the doc's intent.

These requests carry no Sparkth session. They arrive from a learner's browser inside another LMS's course page, and the launch token authenticates them — which is why they pass the plugin access gate untouched: it fails open for anonymous callers by design, as Slack's OAuth callback already relies on.

Every call into the PXC runtime goes through `asyncio.to_thread`. The sandbox executes WebAssembly synchronously on the calling thread, so an inline call would block the event loop for every other request in the process — and no test in this suite would catch it.

`pxc.js` and `pxc-iframe.js` ship inside the installed `pxc-lib` distribution, so the client route serves them from there rather than vendoring a copy. `_CLIENT_FILES` is an explicit two-entry dict rather than a directory, so traversal is impossible by construction.

## How to Test

1. `make test.backend` — 12 route tests.
2. **Drive the URL the shell actually emits**, not a hand-built one. `test_the_shells_own_action_url_can_submit_an_answer` extracts `data-action-url` and `data-pxc-token` from the embed response and POSTs the composed URL exactly as the browser does. This test exists because an earlier revision emitted an action URL with no token: every other test injected the token by hand, so a 422-on-every-submission defect survived until someone traced the real flow.
3. Confirm the anonymous path: `GET /api/v1/pxc/embed?token=<valid>` returns **200, not 403**. A 403 would mean the access gate's fail-open behaviour changed and the entire learner surface is unreachable.
4. Confirm the asset route serves nothing the manifest does not declare — `GET /api/v1/pxc/assets/manifest.json?token=<valid>` is a 404.

## Notes

No migration, no breaking change, no new env vars, no dependency bumps. `openapi.json` is gitignored; the committed artifact is `frontend/lib/api/generated.ts`, and `make test.frontend.api` fails if it goes stale.

The launch token travels as a `?token=` query parameter, so it reaches access logs, the rendered page HTML and browser history. It is TTL-bounded (300s default) and scoped to one placement, but it is a bearer credential in a URL — worth revisiting in the route contract.

`submit_action` reads its body as a raw JSON value rather than a typed model, because an action's value is a manifest-defined `FieldType` — a JSON union no Pydantic model expresses generically. The generated contract therefore carries `requestBody?: never` by design, noted in the docstring so a reader of `generated.ts` alone does not mistake it for a bug.

Per L9 (see PR 5): a sandbox crash during an action currently yields **200 with an empty event list**, not a 502, because `pxc-lib` swallows that failure. The route docstring says so rather than promising an error that cannot arrive.

Stacked PR 6 of 8, targeting `rafe/pxc-tokens-runtime`. Merge in order.

---


sparkth/plugins/pxc/routes.py
```diff
@@ -0,0 +1,135 @@
...
+_CLIENT_FILES = {
+    "pxc.js": Path(pxc.lib.__file__).parent / "static" / "js" / "pxc.js",
+    "sparkth-pxc.js": Path(__file__).parent / "static" / "sparkth-pxc.js",
+}
+
+
+@router.get("/embed", response_class=HTMLResponse)
+async def embed_activity(request: Request, token: str = Query()) -> HTMLResponse:
+    """The document an LMS iframes to show one activity to one learner."""
+    read_launch_token(token)
```
hamza-56, line 44: Every route repeats `read_launch_token(token)` and three of them repeat the `to_thread(build_runtime)` line. A `Depends()` yielding `LaunchClaims` would make the auth boundary structural instead of a line each handler has to remember. The embed test exists precisely because this line is droppable.

sparkth/plugins/pxc/routes.py
```diff
@@ -0,0 +1,135 @@
...
+
+@router.get("/assets/{file_path:path}")
+async def activity_asset(file_path: str, token: str = Query()) -> FileResponse:
+    """Serve the activity's UI script or one of its declared assets.
+
+    Raises:
+        PxcAssetNotFound: if the manifest does not declare the file, or it is missing.
+    """
+    claims = read_launch_token(token)
+    runtime = await asyncio.to_thread(build_runtime, claims)
```
hamza-56, line 101: Building a full runtime here opens the SQLite file, runs the DDL and mkdirs storage just to resolve a static path, and it fires for ui.js plus every asset on each page load. Only the manifest is needed for this.

hamza-56 (commented):
Two things on the route layer.

hamza-56 (approved):
LGTM.

PR #647, opened 2026-09-08
feat(pxc): contribute a seeded activity placement to LMS publishing

abdulrafey1:
## What

Registers the `"pxc"` content contributor, closing the loop this stack was built for: an Open edX publishing tool can now insert a real PXC activity into a course without either plugin importing the other.

This is PR 7 of 8 implementing the PXC integration's first slice.

## Changes

- `feat(pxc)`: add `contributor.py` — mint a placement, seed the sample's configuration, return the block to create — and register it on the hook at module level
- `feat(pxc)`: add `PXC_BLOCK_CATEGORY` and `PXC_DEFAULT_ACTIVITY` to `constants.py`
- `chore(pxc)`: document `PXC_DEFAULT_ACTIVITY` in `.env`
- `fix(pxc)`: wrap `PxcActivityNotFound` as `ContentBuildError`
- `chore(pxc)`: refresh the `es`/`fr` catalog line references

A placement id is minted per call and nothing about it is persisted beyond the seeded fields, so there is no new table and no migration (design doc D4). The id in the block's settings and the id the fields are seeded under are the same binding, not two computations that could drift.

Two details that are load-bearing rather than incidental:

**The seed key blanks the learner segment.** The sample's three configuration fields are `activity`-scoped, and `ActivityRuntime._scope_key_segments` maps `Scope.activity` to `(activity_id, course_id, "")` — only the learner blanks, the course does not. Verified against the bundled `manifest.json`, which declares all three as `activity` scope.

**Storage failures are wrapped in `ContentBuildError`.** `openedx_add_plugin_content` catches exactly that type — deliberately, since the project forbids a bare `except Exception` and a publishing tool cannot know what an arbitrary contributor's storage raises. `SqliteFieldStore` is constructed *inside* the `try` because its `__init__` performs the `mkdir`, which is the most likely failure. A separate clause catches `PxcActivityNotFound` so a misconfigured `PXC_DEFAULT_ACTIVITY` names the setting rather than sending an operator to hunt a disk problem.

## How to Test

1. `make test.backend` — 7 tests in `sparkth/plugins/pxc/tests/test_contributor.py`.
2. **The test this whole architecture exists for** is `test_the_openedx_tool_publishes_this_contributor`: it drives the real `openedx_add_plugin_content`, mocking only the two outbound Studio calls. Remove the `LMS_CONTENT_CONTRIBUTORS.add_item(...)` block from `plugin.py` and it fails with `KeyError: 'response'` — the tool's real `contributor is None` short-circuit — proving the test traverses the actual hook resolution rather than a stub.
3. Confirm the seeded configuration is readable back under the placement the block carries, which `test_the_sample_configuration_is_seeded_under_the_placement` asserts by reconstructing the key from `block.settings["placement"]`.
4. Confirm the wrapping holds: point `PXC_DATA_DIR` at a path occupied by a file, and `build_pxc_block` raises `ContentBuildError` with the cause preserved — not a raw `OSError`.

## Notes

**New env var.** `PXC_DEFAULT_ACTIVITY` (default `mcq`) is in `.env`. Setting it to a non-bundled activity now fails loudly with a message naming the setting.

No migration, no breaking change, no dependency bumps.

Registration is at **module level** in `plugin.py`, not in `SparkthPlugin.__init__` or the package `__init__.py`. `SingleNamedItemHook.add_item` raises on a duplicate name, and both of those run more than once per pytest session.

**Known limits carried by this PR.** One bundled activity is seeded for every request, by design — there is no authoring UI (L5). Scores computed in the sandbox stay in Sparkth and never reach the Open edX gradebook (L1); the XBlock in PR 8 is positioned to change that later.

Stacked PR 7 of 8, targeting `rafe/pxc-routes`. Merge in order.

---


sparkth/plugins/pxc/contributor.py
```diff
@@ -0,0 +1,63 @@
...
+
+logger = get_logger(__name__)
+
+# The bundled sample's configuration. Activity-scoped fields, which is all the sample uses.
+_SAMPLE_QUESTION = "What is 2 + 2?"
+_SAMPLE_ANSWERS = ["3", "4", "5"]
+_SAMPLE_CORRECT_ANSWERS = [1]
+
+
+async def build_pxc_block(course_id: str) -> ContentBlock:
```
hamza-56, line 32: `async`, but everything inside is blocking sqlite (mkdir, DDL, three writes) on the event loop, while #646 goes out of its way with `to_thread` for the very same store.

hamza-56 (commented):
One thing.

sparkth/plugins/pxc/contributor.py
abdulrafey1: The activity now supplies its own starting configuration, so building a placement writes nothing and there is nothing left to block on.

hamza-56 (approved):
LGTM.

PR #648, opened 2026-09-08
feat(xblock): render a Sparkth-hosted PXC activity in an Open edX unit

abdulrafey1:
## What

Adds the Open edX XBlock that holds a reference to a Sparkth-hosted PXC activity and renders it — the missing half that makes the slice demoable end to end.

This is PR 8 of 8 implementing the PXC integration's first slice. Without it a course contains a `pxc` block no XBlock class can resolve.

## Changes

- `feat(xblock)`: add `xblock/`, a separate Django distribution — `sparkth_pxc.xblock` with its `Scope.settings` fields and `student_view`, plus a minting-only copy of the launch token
- `fix(xblock)`: `allow-same-origin` so the embedded activity can fetch its config
- `docs(xblock)`: name the bearer-token exposure behind `allow-same-origin`
- `docs(xblock)`: note the anonymous-learner identity collision at its call site
- `chore(xblock)`: ignore the `uv.lock` its own test run regenerates
- `refactor(xblock)`: mint launch tokens with pyjwt instead of by hand

**This package is not part of the Sparkth application.** Sparkth never imports it, and its dependencies (`XBlock`, `Django`, `web-fragments`, `PyJWT`) are not Sparkth dependencies. A root `collect_ignore` keeps its tests out of the main suite, and its own `pyproject.toml` carries pytest and mypy anchors so a run from inside `xblock/` does not climb to the outer repo's config and drag Sparkth in.

**The token format is shared, not imported.** PR 5 verifies these tokens; this package mints them. They are separate distributions on different servers and cannot import each other, so `xblock/tests/test_tokens.py` pins what both sides must agree on: the claim names, HS256, and an `exp` that is an absolute deadline rather than a duration. If the two halves ever disagree, every learner launch fails with a 401 that no test in either package alone would catch.

Both halves are HS256 JWTs signed by `pyjwt`, so the signing itself is no longer duplicated — only the three things above, which are this project's own choice and have nowhere else to live. Open edX already installs `pyjwt`, but a dependency that merely happens to be present in the host environment is not a declared one, so it goes in this package's own `dependencies`.

**`allow-same-origin` is required, not permissive by accident.** Without it the iframe gets an opaque origin, so every `fetch()` the shell makes — including back to the host that served it — is cross-origin with `Origin: null`, and Sparkth registers no CORS middleware. The activity would silently never render. The attribute carries an inline comment saying so, because the next reader would otherwise tighten it and break every launch. It does not re-open the classic `allow-scripts` + `allow-same-origin` hole, which requires the frame to be same-origin with its *embedder*: Sparkth is cross-origin to Open edX either way.

## How to Test

1. `uv run --with pytest --directory xblock pytest -v` — 7 tests. Only `pytest` needs adding; everything the package imports is in its own `dependencies`.
2. `make test.backend` — unchanged, because this package is outside the application.
3. Confirm the isolation holds:
   `uv run python -c "import sys, sparkth.main; print('django' in sys.modules, 'sparkth_pxc' in sys.modules)"` prints `False False`, and `grep -rn "sparkth_pxc" sparkth/` returns nothing.
4. Confirm the entry-point name matches the category the contributor publishes: `pxc` in `xblock/pyproject.toml` equals `PXC_BLOCK_CATEGORY` in `sparkth/plugins/pxc/constants.py`. A mismatch means Studio cannot resolve the block.

### Manual verification, which no test covers

1. `make pxc.activities.build && make services.up && make backend.up.dev`
2. Set `PXC_LAUNCH_SECRET` in `.env.local` and the same value as `SPARKTH_PXC_LAUNCH_SECRET` in the Open edX instance's Django settings.
3. `pip install ./xblock` into the Open edX instance; add `pxc` to the course's Advanced Module List.
4. Through the chat agent, create a unit and call `openedx_add_plugin_content` with `contributor: "pxc"`.
5. Open the unit as a learner. The activity renders, an answer submits, and feedback appears.
6. `sqlite3 data/pxc/mcq.sqlite3 "SELECT * FROM fields"` shows the state landed.

Step 5 is the one nothing automated exercises — a real browser, a real cross-origin iframe, a real Open edX identity.

## Notes

**New Django settings**, read by this package inside Open edX: `SPARKTH_PXC_BASE_URL`, `SPARKTH_PXC_LAUNCH_SECRET`, `SPARKTH_PXC_LAUNCH_TOKEN_TTL_SECONDS`. The secret must match Sparkth's `PXC_LAUNCH_SECRET`. Documented in `xblock/README.md`.

No migration, no breaking change to Sparkth, no Sparkth dependency bumps. This package declares one new dependency of its own, `PyJWT>=2.10.1`, which Open edX already installs.

**Known limits.** The XBlock must be installed in Open edX and enabled per course in the Advanced Module List (L6). This package lives in the Sparkth repository temporarily; moving it into the PXC repository as a remote-backend mode of the existing `pxc-xblock` is the obvious alternative and that decision is open (L7). Scores stay in Sparkth and never reach the gradebook (L1).

**L10, and it is worse than a one-line note suggests.** `allow-same-origin` means the shell runs at Sparkth's origin, and `pxc.js`'s `_loadScript` does `await import(url)` into *document* scope — a closed shadow root isolates markup, not script. The frontend writes the user's bearer token to `localStorage` on that origin, so an untrusted activity's `ui.js` could read it. Bounded today by browser storage partitioning for third-party frames and by L5 (only Sparkth's own sample ships). The follow-up is `pxc.js`'s `_initIframe` path — a nested sandboxed iframe talking by `postMessage` — or serving the embed from a separate origin. `SparkthPXC` deliberately opts out of that path today. Documented in `xblock/README.md`.

An anonymous Open edX learner yields `"None"` as the identity, so all anonymous visitors to a placement would share one. Latent, because the bundled sample uses activity scope only; closing it needs a product decision on what an anonymous visitor should see. Noted at the call site.

Stacked PR 8 of 8, targeting `rafe/pxc-contributor`. Merge in order.

---


hamza-56 (commented):
Only a cross-reference; the XBlock itself reads fine.

xblock/sparkth_pxc/tokens.py
```diff
@@ -0,0 +1,43 @@
...
+
+def _b64encode(data: bytes) -> str:
+    return urlsafe_b64encode(data).rstrip(b"=").decode("ascii")
+
+
+def _sign(payload: str, secret: str) -> str:
+    return _b64encode(hmac.new(secret.encode(), payload.encode(), hashlib.sha256).digest())
+
+
+def mint_launch_token(
```
hamza-56, line 26: See the note on #645's `mint_launch_token`: with pyjwt on both ends this module becomes one `jwt.encode` call and the duplicated wire format goes away.

hamza-56 (approved):
LGTM.

PR #650, opened 2026-09-09
fix(pxc): read the plugin's settings through pydantic-settings

abdulrafey1:
## What

The PXC plugin read its environment-backed values with `os.getenv` at import time, so nothing set in `.env` or `.env.local` ever reached it. For `PXC_LAUNCH_SECRET` that meant the secret stayed empty however it was configured, and Sparkth refused every learner's launch with a 401.

This is a follow-up to the PXC integration slice (#641–#648), stacked on #648.

## Changes

- `fix(pxc)`: add `config.py` — a `PxcSettings(BaseSettings)` with `env_prefix="PXC_"` and `env_file=ENV_FILES`, plus a cached `get_pxc_settings()`
- `fix(pxc)`: read `data_dir`, `launch_secret`, `launch_token_ttl_seconds` and `default_activity` from it in `activities.py`, `runtime.py`, `tokens.py` and `contributor.py`
- `fix(pxc)`: resolve `mint_launch_token`'s token lifetime per call instead of as an argument default
- `test(pxc)`: add `test_config.py`, and pin every PXC setting in an autouse fixture so tests cannot read the developer's env files
- `docs(xblock)`: the README's launch-secret section, now that `.env.local` works

**Why the values silently vanished.** `.env` and `.env.local` are parsed by pydantic-settings into the `Settings` object, which never writes to `os.environ`; `uv run` does not load them either. So every `os.getenv` read fell back to its in-code default. Three of the four defaults match `.env`, which is why this went unnoticed — it only bites a value whose default is deliberately unusable, and a fail-closed secret is exactly that.

**This is the pattern the project already mandates.** [`sparkth/lib/settings.py`](sparkth/lib/settings.py)'s docstring requires a plugin defining its own `BaseSettings` to pass `ENV_FILES`, and `chat`, `googledrive` and `slack` each do. PXC was the only plugin using bare `os.getenv`; after this there are none left in the package.

**`PXC_ACTIVITY_ROOT` and `PXC_BLOCK_CATEGORY` stay in `constants.py`.** Neither is configuration. One is a path inside the package, and the category must equal the name of the entry point the XBlock registers — making it settable would let the two drift apart and leave Studio unable to instantiate the block.

**The test fixture is not incidental.** Once `PxcSettings` reads the env files, a developer with a real `PXC_LAUNCH_SECRET` in `.env.local` would have it become the secret under test, and a real `PXC_DATA_DIR` would have tests writing into the working checkout's own state files. The autouse fixture in the plugin's conftest pins all four values, and environment variables win over both files.

## How to Test

1. `make test.backend` — 2048 passed, 3 skipped (4 new tests in `sparkth/plugins/pxc/tests/test_config.py`).
2. **The regression test for the actual defect** is `test_a_secret_from_configuration_is_what_verifies_a_token`: it configures the secret through the environment rather than by patching a module attribute, which is what no previous test did — every token test patched `tokens.PXC_LAUNCH_SECRET` directly, so none of them could observe that configuration never arrived.
3. Confirm the prefix mapping, which is the whole contract with `.env`: `test_the_pxc_prefix_maps_onto_the_settings_field_names` fails if a field is renamed without its variable.
4. Confirm the lifetime is no longer frozen at import: `test_the_token_ttl_default_comes_from_configuration` sets a negative TTL and expects an expired token. Zero would not work — expiry is a strict `<`.
5. Against a running instance, with `PXC_LAUNCH_SECRET` set **only** in `.env.local` and not exported:
   `curl "http://localhost:7727/api/v1/pxc/embed?token=bogus.signature"` returns `Bad launch token signature`. Before this change the same setup returned `Launch tokens are not configured`, and every real launch 401'd.

### Verified end to end

Against a local Tutor-hosted Open edX (Tutor 22.0.2) with the XBlock installed, secret only in `.env.local`: the LMS mints a launch token, `/config` returns the seeded activity state, and `POST /actions/answer.submit` with `[1]` returns `answer.result {"correct":true}`.

## Notes

No migration, no new env vars, no dependency bumps.

**Behaviour change worth noting for deployments:** the four `PXC_*` variables are now read from `.env`/`.env.local` in addition to the process environment, where previously only the process environment worked. An environment variable still overrides both, so a containerised deployment injecting real variables is unaffected.

`get_pxc_settings` is `lru_cache`d like its siblings, so a test changing one of these values must call `get_pxc_settings.cache_clear()` — the autouse fixture does this on both sides of every test.

---


hamza-56 (approved):
LGTM, this is the right fix and the autouse fixture pinning the env is a good call.

PR #651, opened 2026-09-09
feat(pxc): let a course author edit a PXC activity from Studio

abdulrafey1:
## What

Lets a course author edit a PXC activity from Studio's **Edit** button. Clicking it previously reported "We're having trouble rendering your component", which was misleading twice over — nothing was broken and students could access the component. The block simply had no `studio_view`, and that is the view Studio's edit modal renders.

Stacked on #650.

## Changes

- `feat(pxc)`: carry the launch permission as a signed `prm` token claim, resolved on the verifying side
- `feat(pxc)`: build the runtime at the permission the token asked for, replacing a hardcoded `Permission.play`
- `feat(xblock)`: add `studio_view`, minting `edit`, with a notice explaining that Studio's own Save and Cancel do not apply
- `chore(pxc)`: run the `xblock/` package in CI, and five fixes from the whole-branch review

**The activity side already worked.** `activity/sandbox.js` refuses `config.save` unless the permission is `edit` and exposes `correct_answers` only in `edit`; `activity/ui.js` already had a complete author view with a Save button. This PR is the plumbing between them, which is why it adds no activity code.

**A consequence worth stating: the hardcoded `play` was safe by enforcement, not by accident.** A learner could not read the answer key or rewrite the question before this change, and still cannot.

## Security

#645 documented the property *"permission is not a claim, so a caller can never ask for `edit` by editing a token."* **That property survives.** `prm` sits inside the HMAC-signed payload, so altering it invalidates the signature — pinned by `test_escalating_the_permission_claim_breaks_the_signature`, which forges a payload that is valid base64 JSON so its refusal is provably about the signature rather than about malformed input.

What changes is *which party decides*, and it moves to Open edX — the only party that knows who holds authoring rights on a course.

**The cost, stated plainly: `PXC_LAUNCH_SECRET` now grants `edit`, not only `play`.** Compromising it previously allowed impersonating a learner; it now also allows rewriting activity content.

**One deployment-dependent caveat.** Open edX's LMS `xblock_view` endpoint renders an arbitrary `view_name` for any authenticated user with access to the block, gated only by `FEATURES['ENABLE_XBLOCK_VIEW_ENDPOINT']`. It is off by default (and off on the instance this was verified against), but with it on an enrolled learner can request `studio_view` and receive an `edit` token. Named in `xblock/README.md` and in `sparkth/plugins/pxc/tokens.py`'s module docstring. Deliberately not gated in code: a check would need the user service and would break the test runtime that makes the view-level tests possible.

## How to Test

1. `make test.backend` — 2056 passed, 3 skipped.
2. `make test.xblock` — **new target**, 11 tests plus mypy and ruff. `conftest.py`'s `collect_ignore = ["xblock"]` keeps that separate distribution out of the root suite, and until this PR nothing automated ran it at all.
3. **The escalation pair is the test to read.** `test_escalating_the_permission_claim_breaks_the_signature` (forged claim rejected on signature) and `test_the_runtime_takes_its_permission_from_the_claims` (honest claim reaches the runtime, asserted in both directions). Either alone would pass an implementation that ignores the claim entirely.
4. Both compatibility paths: an absent `prm` claim and an unrecognised value each resolve to `play`, the unrecognised one logging a warning that now names the activity and placement so it can be correlated to a launch.
5. `test_configuration_saves_with_edit_and_is_ignored_with_play` (`wasm`-marked) drives the real sandbox: a play-mode `config.save` is *silently ignored* rather than raising, because the sandbox is the enforcement point.
6. Both views are rendered through a real XBlock runtime, so a permission swap between them fails the tests. That matters more than it sounds — see below.

### Verified end to end against a live Open edX

Tutor 22.0.2 with the XBlock installed, after rebuilding the derived image: `studio_view` exists on a real block and its token carries `prm=edit` with the real course id; `student_view` carries `play`; Sparkth returns `permission: edit` with `correct_answers` exposed for the edit token and `play` without them for the play token; `config.save` rewrites the question with an edit token and returns 200 while changing nothing with a play token.

## Notes

No migration, no new env vars, no dependency bumps.

### Studio's Save and Cancel are inert for this component, by design

The author's edits persist the moment they click Save *inside the activity* — `ui.js` sends `config.save`, Sparkth applies it through the sandbox, the sandbox writes to the field store. By the time Studio's modal buttons are reachable the change is already durable, so **Cancel reverts nothing**.

Studio's buttons operate on XBlock fields, and no XBlock field holds activity content: the course stores only a reference (design doc D4). Making Cancel behave would mean copying activity content into XBlock fields and holding the same state in two places. Disclosed in three places instead of fixed — a notice in the editing view itself, a README section, and here.

### What the whole-branch review found, and what it changed

The review returned "fix first" on something worth highlighting: **`xblock/`'s tests were run by nothing automated**, while this PR makes that directory the only automated guard on a privilege decision. Changing `student_view`'s literal `"play"` to `"edit"` would hand every learner the answer key, and the one test that catches it never ran in CI. Hence the `make test.xblock` target and its CI step.

It also found that a rejected action reported success to the author: `sendAction` did `console.error(...); return` on a non-OK response, so `ui.js`'s `catch` never fired and it wrote "Configuration saved!" over a discarded edit — reachable once the 300-second launch token lapses mid-authoring. `sendAction` now throws. The learner path already try/catches and renders an error, so this improves both.

And it identified two mutations the whole suite missed: reversing the notice/iframe concatenation (leaving the notice below a 24em iframe where an author never sees it), and `_user_id` returning any constant. Both now have guards.

### Two things accepted rather than fixed

**Per-event permission filtering.** `submit_action` returns events without filtering on each event's declared minimum permission, which `pxc-lib`'s `send_event` records. Not exploitable with the bundled activity — only `config.save` emits an `edit`-scoped event and it is refused in play mode — but a future activity emitting `edit`-scoped events would leak them to play-mode clients. Enforcing it is a decision about the event contract rather than a one-liner.

**Fix 2's test is a source-level guard**, not an executed JS test: it regex-checks the served JavaScript for the `throw`. This repo has no JS runner for a backend-served static asset, and building one for a three-line change was disproportionate. Verified independently to fail if the fix is reverted, but it is a source guard and worth knowing as one.

### Rollout

Deploy Sparkth before the XBlock. Sparkth reads an absent `prm` claim as `play`, so an older XBlock keeps launching correctly; the reverse renders the editing notice above a learner view whose Save button does not exist.

---


sparkth/plugins/pxc/tests/test_routes.py
```diff
@@ -70,6 +70,24 @@ async def test_the_client_route_serves_the_bundled_pxc_component(client: AsyncCl
     assert "class PXC" in response.text
 
 
+async def test_a_rejected_action_throws_rather_than_returning(client: AsyncClient) -> None:
```
hamza-56, line 73: Regex over the served JS source isn't a behaviour test: it breaks on any reformat and passes with a `throw` that never runs. Either a small vitest for the static file, or drop it.

Makefile
```diff
@@ -218,6 +218,16 @@ test.backend.analytics: services.up ## Run the TimescaleDB test lane (starts bac
 	ANALYTICS_TEST_PG_URL="$${ANALYTICS_TEST_PG_URL:-postgresql://sparkth:sparkth_password@localhost:5432/sparkth_analytics_test}" \
 		uv run pytest -m pg $(ARGS)
 
+# xblock/ is a separate Django distribution (see conftest.py's collect_ignore) with its own
+# pyproject.toml, so it is not covered by `make test.backend` — everything it needs (pytest,
+# mypy, ruff) is run explicitly here instead, on the same footing as the main package.
+.PHONY: test.xblock
+test.xblock: ## Run the Open edX XBlock package's tests, mypy and ruff (separate distribution)
```
hamza-56, line 225: `--with django --with XBlock --with web-fragments` with no pins and `xblock/uv.lock` gitignored means CI installs the latest of each on every run. Same random-breakage argument the stack made for pinning pxc-lib.

hamza-56 (commented):
Permission-as-signed-claim is sound. Two comments on the test side.

hamza-56 (approved):
LGTM.

PR #653, opened 2026-09-10
fix(api): report a plugin's config schema keys when its config is empty

regisb:
## What

A plugin whose config was never saved reported no config keys at all, so its settings form rendered zero fields — Open edX and Canvas were impossible to configure.

## Changes

- fix(api): merge the schema keys under the stored config at every user-plugin read site, via the new `PluginService.config_with_schema_keys`
- test(api): cover the empty-config and partially-filled-config cases on the list, get, and update endpoints

## How to Test

1. `pytest tests/api/v1/test_list_user_plugins.py tests/api/v1/test_get_user_plugin.py tests/api/v1/test_update_user_plugin.py`
2. Enable the Open edX plugin in the settings UI, reopen its config modal, and check that the credential fields are listed.

## Notes

Enabling a plugin stores an empty config `{}`, and the endpoints previously fell back to the schema keys only when the stored config was `None`. Merging instead of choosing also restores keys that a partially filled config lacks. No migration, no breaking change.

### DIFF
--- a/tests/api/v1/conftest.py
+++ b/tests/api/v1/conftest.py
@@ -1,4 +1,4 @@
-from typing import cast
+from typing import Any, cast
 
 import pytest
 from sqlmodel.ext.asyncio.session import AsyncSession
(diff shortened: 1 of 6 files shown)


tests/api/v1/conftest.py
```diff
@@ -35,3 +35,41 @@ async def user_plugins(current_user: User, session: AsyncSession) -> User:
     await session.flush()
 
     return current_user
+
+
+# A plugin config schema in the shape Pydantic's ``model_json_schema()`` emits, so the
+# responses exercise the same key extraction as a real plugin (see
+# ``PluginService.initial_config``).
```
hamza-56: `initial_config` got renamed to `config_with_schema_keys` in this PR, so this comment now points at nothing.

abdulrafey1 (approved):

PR #654, opened 2026-09-10
feat(plugins): render the settings form from the declared config schema

regisb:
## What

The generic settings modal rendered one untyped text input per config key, so any plugin needing a richer control had to ship a bespoke React modal — chat and Slack each carried one, which is why they alone had usable forms.

## Changes

- feat(api): expose the plugin's JSON config schema on the user-plugins endpoints
- feat(plugins): add `ConfigWidget` / `widget()` so a config field can name the control that renders it
- feat(frontend): build the settings form from the schema — types, labels, descriptions, defaults, required-ness and numeric bounds all come from the Pydantic config class
- feat(frontend): resolve `widget` hints against a registry core populates for shared controls and a plugin extends for its own, the way plugin icons already work
- refactor(frontend): delete `ChatConfigModal` and `SlackConfigModal` along with `PluginDefinition.loadSettingsComponent`
- feat(plugins): give Open edX and Canvas masked credential fields
- test: cover the schema builder, the widget registry, the generic modal, and each plugin's config

## How to Test

1. `pytest tests/core/plugins/test_config_widgets.py sparkth/plugins/*/tests/test_config.py`
2. `cd frontend && npm test -- tests/lib/plugin-schema.test.ts tests/components/settings/ConfigModal.test.tsx`
3. In the settings UI, open the config modal for chat, Slack, Open edX and Canvas: each renders its declared fields, with the secret fields masked and the LLM config field rendered as a select.

## Notes

One behaviour change: Slack's synthesis switch becomes a "None" option on the LLM config select, which is what disabled synthesis before. No migration, no new env var.


frontend/lib/plugins/schema.ts
```diff
@@ -0,0 +1,208 @@
...
+  return variantsOf(property).find((variant) => variant.type !== "null") ?? property;
+}
+
+function widgetOf(property: SchemaObject, type: ConfigFieldType): string {
+  const declared = asString(property.widget);
+  if (declared) return declared;
+
+  if (type === "boolean") return "checkbox";
+  if (type === "number" || type === "integer") return "number";
+  if (type === "array") return "tags";
```
hamza-56, line 87: `"tags"` is never registered in widgets.tsx, so a `list[str]` field with no hint falls through to TextWidget, and then `payloadOf` throws the typed string away and sends `[]`. Nothing breaks today since the only list field (Slack `allowed_sources`) has the doc-sources hint, but the first plugin that declares a plain list gets a form that silently saves nothing. Either ship a small tags / comma-separated widget or fall back to `text` and split on commas in `payloadOf`.

frontend/components/settings/widgets.tsx
```diff
@@ -0,0 +1,232 @@
...
+  useEffect(() => {
+    if (!token) return;
+    let ignore = false;
+    Promise.all([fetchLLMConfigs(token, { includeInactive: true }), fetchProviderCatalog(token)])
+      .then(([configResult, catalog]) => {
+        if (ignore) return;
+        setConfigs(configResult.configs);
+        setProviders(catalog.providers);
+      })
+      .catch(() => {
```
hamza-56, line 190: If the catalog fetch fails we swallow it and the select stays disabled saying "Loading available models…" forever. Worth surfacing it (even just a "could not load models" helper) so the user is not left waiting. Small nit on the same lines: this refetches the LLM configs that LLMConfigSelect just fetched one field above.

frontend/components/settings/widgets.tsx
```diff
@@ -0,0 +1,232 @@
...
+const asText = (value: unknown) => (value === null || value === undefined ? "" : String(value));
+
+function TextWidget({ field, value, error, disabled, onChange }: ConfigWidgetProps) {
+  const isUrl = field.widget === "url";
+  return (
+    <Input
+      id={`config-${field.name}`}
+      name={field.name}
+      label={field.label}
+      type={field.widget === "password" ? "password" : isUrl ? "url" : "text"}
```
hamza-56, line 75: Password inputs invite browsers to autofill saved logins into API key fields and to offer to "save" the key as a password. Passing `autoComplete="new-password"` (or `"off"`) when the widget is `password` avoids that, and Input already spreads the rest of the props.

sparkth/core/plugins/service.py
```diff
@@ -61,18 +61,31 @@ class UserPluginResponse(pydantic.BaseModel):
...
 
     @classmethod
-    def for_plugin(cls, plugin_name: str, enabled: bool, config: dict[str, Any], is_core: bool) -> "UserPluginResponse":
+    def for_plugin(
+        cls,
+        plugin_name: str,
+        enabled: bool,
+        config: dict[str, Any],
+        is_core: bool,
+        config_schema: dict[str, Any] | None = None,
```
hamza-56, line 87: All five call sites pass `config_schema` now, so making it optional mostly leaves room for a future call site to forget it and ship a schema-less response, which is the same "zero fields" failure #653 just fixed. I would make it required.

sparkth/api/v1/user_plugins.py
```diff
@@ -179,11 +184,19 @@ async def get_user_plugin(
             config,
         )
         return UserPluginResponse.for_plugin(
```
hamza-56, line 186: The two branches of this `if` now differ only in `enabled` and whether postprocess runs. The list endpoint already runs `apply_postprocess` on the schema-key config whether or not a user_plugin row exists, so this could do the same: postprocess `config` once, `enabled=user_plugin.enabled if user_plugin else True`, and a single `for_plugin` call.

Issue #657, opened 2026-09-14
A PXC activity stops accepting a learner's actions 300 seconds after page load, until the unit is reloaded


abdulrafey1:
## Summary

A PXC activity's WebSocket session is bounded by the launch token's 300-second
lifetime, not by how long the learner is actually using the activity. The socket
itself stays open, but 300 seconds after the unit rendered, the next action the
learner sends closes it — and the only way to send actions again is to reload the
browser tab.

Any activity a learner spends more than five minutes in is affected. A chess game,
a multi-step exercise, a group activity — all of them stop accepting input
mid-session.

> **Where this code lives.** The socket transport is on a local, unpushed branch
> `rafe/pxc-websocket-transport` (tip `339901a8`), stacked above the open PXC PRs
> (#641–#648, #650, #651). The relevant commits are `74ae9999`
> (*feat(pxc): move the embed client onto the websocket transport*) and `910bf7e2`
> (*docs(pxc): document the socket transport and its deployment limits*). Nothing
> cited below is on `main` or reachable from an open PR yet.

## Mechanism

1. The launch token is minted **once, at page render**, by the XBlock, server-side,
   with the shared secret — `xblock/sparkth_pxc/xblock.py:95-104` (`student_view`)
   and `:117-126` (`studio_view`), both passing
   `settings.SPARKTH_PXC_LAUNCH_TOKEN_TTL_SECONDS`.
2. That TTL defaults to 300 seconds —
   `sparkth/plugins/pxc/config.py:37` (`launch_token_ttl_seconds: int = 300`).
3. The socket re-verifies the token on **every action frame**, not just at the
   handshake — `sparkth/plugins/pxc/websocket.py:162`
   (`claims = read_launch_token(token)` inside `_run_one_action`).
4. Once it has lapsed, the frame loop closes the socket with
   `WS_1008_POLICY_VIOLATION` — `sparkth/plugins/pxc/websocket.py:54-62`.
5. The browser cannot obtain a replacement. The client holds one token, read from
   `data-pxc-token` at construction, and never refreshes it
   (`sparkth/plugins/pxc/static/sparkth-pxc.js`). **And the configuration route is
   itself token-gated** — `activity_config` takes `token: str = Query()` and calls
   `read_launch_token(token)` (`sparkth/plugins/pxc/routes.py:84-86`) — so an
   expired client cannot fetch a fresh config to get a fresh token either.

The XBlock is the only party that can mint a token, and it only does so when it
[...]

PR #658, opened 2026-09-14
feat(pxc): move activity communication onto a websocket

abdulrafey1:
## What

Moves PXC activity communication onto a WebSocket so an activity can receive events it did not itself cause — the property that makes a shared, multi-player activity possible at all.

Stacked on #651.

## Changes

- `feat(pxc)`: add the activity event bus and a fan-out that survives a dead subscriber
- `feat(pxc)`: add the launch-authenticated activity websocket route
- `feat(pxc)`: run action frames from the socket and fan out their events
- `feat(pxc)`: publish action events instead of returning them, and carry the socket URL in the config
- `feat(pxc)`: move the embed client onto the websocket transport
- `fix(pxc)`: refuse binary, deeply nested and over-long-integer frames instead of crashing
- `fix(pxc)`: stop forcing every activity socket URL to `ws://`
- `docs(pxc)`: document the socket transport and its deployment limits

**Most of this transport already existed and is adopted rather than invented.** `pxc-lib`'s client ships the socket, reconnect backoff, IndexedDB action queue and 512 KiB HTTP fallback; its `EventBus` ships context and permission filtering. What did not exist anywhere — upstream or here — is a server handler: `event_bus.py` is the only file in the library that imports WebSocket types, and Sparkth had no WebSocket route at all. This PR is that handler, plus the wiring.

Adopting the bus also closes a finding accepted rather than fixed in #651: events are now filtered by each event's declared minimum permission, so an `edit`-scoped event does not reach a `play` client.

## Security

**The property most likely to be lost here is that a learner cannot escalate their own permission, and it survives.** `pxc.js` puts a `permission` field in every action frame, and the client belongs to the learner. The handler never reads it: `_run_one_action` takes exactly two things off the frame, `action` and `value`, and the permission the sandbox runs at comes from the signed `prm` claim resolved at `build_runtime`. A frame saying `"permission": "edit"` is inert.

Pinned by `test_a_frame_cannot_escalate_its_own_permission`, which asserts on the *effect* rather than an error — the sandbox silently ignores a `play`-mode `config.save`, so a test expecting a rejection message would pass against a handler that trusts the frame.

**The token is re-verified on every action frame**, not only at the handshake, so an action submitted after it lapses does not take effect — the property `xblock/README.md` states for the HTTP routes now holds identically for a socket session.

## How to Test

1. `make test.backend` — 2083 passed, 3 skipped (the pre-existing TimescaleDB lane).
2. `make pxc.activity.build` first, then `uv run pytest sparkth/plugins/pxc/ -rs`. **The skip list must be empty.** Four of these tests are `wasm`-marked, including the permission-escalation guard, and a skipped run silently leaves the central security property unverified.
3. `make test.xblock` — 11 passed. This branch changes no `xblock/` code, only its README, so it is a regression check.
4. `make test.frontend.api` — exit 0. The route's response shape and docstring both changed, so the generated client was regenerated; a stale one fails CI.
5. **The test to read is the two-learner one** (`test_two_learners_on_one_placement_both_receive_one_learners_event`): one client acts, the other sends nothing, both receive. That is the whole point of the branch and was impossible before it.

### Verified end to end against a live Open edX

Two real learner accounts, two browsers, through Open edX, over the socket, driving the chess sample activity on its real wasm sandbox: one player moves and the other board updates without them touching it.

## Notes

No migration, no new env vars, no dependency bumps. `ActionResult` is deleted and `POST /actions/{name}` now returns 204 — no external consumer exists, since the route is only reachable from this unmerged stack.

### Known limitations, documented rather than fixed

Recorded in `xblock/README.md` with the condition that triggers each:

- **Fan-out is per-process.** The bus is an in-memory dict, so both parties to a shared activity must be served by the same worker. Above one uvicorn worker or replica, two learners see divergent state **with no error at either end**.
- **The socket's lifetime is the launch token's.** 300 seconds by default, after which actions are refused and a dropped socket cannot reconnect — the client's only credential has expired and only a page render can mint a new one. Filed as #657, with options; it wants a design decision rather than a patch.
- **An action is acknowledged as queued, not as saved.** `sendAction` resolves once the record reaches IndexedDB, before any network round-trip, so a refused action is not surfaced to the learner.
- **The socket URL's scheme depends on `scope["scheme"]`.** Nothing here trusts a forwarded-proto header, so a TLS deployment behind an untrusted proxy yields `ws://` and the browser blocks it as mixed content. Pre-existing — `ui_url`, `asset_base_url` and `action_base_url` share the dependency — but it kills the activity outright rather than one asset.

### Found upstream while building this

Three defects in `edly-io/pxc` worth reporting alongside the existing ones: `EventBus.publish` catches only `WebSocketDisconnect` (an application-side close raises an uncaught `RuntimeError` that aborts the fan-out mid-loop); `publish` iterates its live subscriber list with an `await` inside, so a concurrent unsubscribe skips a subscriber; and `pxc.js`'s own `_postAction` has no `try/catch` around `fetch`, so a network throw strands the queue. The first is worked around here by the subscriber adapter; the others are not.

---


xblock/README.md
```diff
@@ -73,8 +73,18 @@ token.
 
 The launch token's lifetime also bounds how long an editing session can run: it defaults to
 300 seconds (`SPARKTH_PXC_LAUNCH_TOKEN_TTL_SECONDS`), and Sparkth refuses a save submitted
-after it lapses. The activity reports that refusal to the author as an error, rather than
-reporting success.
+after it lapses. The socket re-verifies the token on every action, so this bound applies to a
+socket session exactly as it applies to the HTTP routes.
+
+A refused save is not reported to the author as an error, though. `sendAction` resolves as soon
```
hamza-56, line 79: This is a regression on #651, which specifically made a rejected save throw so the author stops seeing "Configuration saved!" on a discarded edit. Moving to the socket brings that false success back, and writing it up as a known limitation doesn't make silent data loss on the editing surface okay to ship. The edit path needs some visible failure before this lands.

xblock/README.md
```diff
@@ -107,6 +117,55 @@ a learner reports a launch failure.
...
+matters more here, because a blocked WebSocket kills the activity outright rather than one
+asset.
+
+### A socket that drops after the token expires does not reconnect on its own
+
+`pxc.js`'s `_getWebsocketUrl` returns the cached socket URL unconditionally once one is set,
+and its reconnect loop simply reopens that same URL, so every reconnect presents the same
+launch token. Once that token has expired, the server refuses the reconnect's handshake
+outright — the client never reaches an open socket to receive a proper close code — and
+`pxc.js` retries anyway, on a backoff that caps at 2000 ms. That is a refused handshake every
```
hamza-56, line 155: Every stale tab hammering the handshake every 2 s forever, from the moment any blip happens after the TTL, is a self-inflicted load problem as much as a UX one. At minimum the client should give up (or the server should tell it to) rather than retry an expired token indefinitely.

hamza-56 (commented):
The transport itself reads well (frame loop, fan-out adapter, unsubscribe ordering). Three things inline, two of them about behaviour this PR documents rather than fixes.

sparkth/plugins/pxc/tests/test_routes.py
```diff
@@ -70,22 +74,51 @@ async def test_the_client_route_serves_the_bundled_pxc_component(client: AsyncCl
     assert "class PXC" in response.text
 
 
-async def test_a_rejected_action_throws_rather_than_returning(client: AsyncClient) -> None:
-    """Guards ``sparkth-pxc.js``'s ``sendAction()`` against reverting to a silent ``return``.
+async def test_the_client_connects_a_socket_from_the_configured_url(client: AsyncClient) -> None:
```
hamza-56, line 77: The regex-over-JS-source pattern I flagged on #651 went from one test to three here. These pin formatting, not behaviour, and every reformat of `sparkth-pxc.js` breaks them while a broken client still passes.

hamza-56 (approved):
LGTM.

Issue #507, opened 2026-07-06
Audit trail has no retention policy and no GDPR-compatible erasure model


hamza-56:
# Summary

Part of epic #435.

The audit trail has no retention policy and no erasure model. Nothing defines how long audit events of a given kind must be kept or when they may be removed, and there is no way to honor a GDPR Art. 17 erasure request for a user whose personal data (actor identifiers, IPs, payload fields) is embedded in records that are, by design, append-only and immutable.

# Where it happens

- No retention configuration, expiry mechanism, or per-category retention classification exists for audit data.
- Actor identifiers and event payloads are stored as plaintext, so removing one user's data from an immutable trail has no non-destructive path.

# Why it matters

The regimes anchoring this epic conflict on retention and the conflict must be resolved by design, not policy: the EU AI Act requires at least 6 months of logs (Art. 19, 26(6)), Title IV financial-aid audits need award-year-anchored 3+ year windows with legal holds (34 CFR 668.24), NIST AU-11 expects organization-defined retention, and GDPR requires erasure of personal data on request. An append-only trail built over plaintext PII cannot satisfy the last one retroactively; retrofitting an erasure-compatible design after the trail (and any integrity mechanism over it) exists is effectively a re-architecture.

# How to observe

1. Inspect the audit implementation: no retention window, expiry job, or per-category classification exists.
2. Attempt to answer "delete everything you hold about user X" while preserving the audit trail's integrity guarantees: there is no supported way to do both.

_This issue description was written with the assistance of an LLM (Claude)._

PR #660, opened 2026-09-16
feat(core): reject update, delete, and truncate on audit_events

### LINKED ISSUE #504: Audit records are not protected from update or delete at the database level
# Summary

Part of epic #435.

Audit records are append-only by application convention only: nothing at the database level prevents the runtime application role (or anyone with its credentials) from issuing `UPDATE`, `DELETE`, or `TRUNCATE` against audit data. A compromised application role, a buggy migration, or an ordinary ORM mistake can rewrite or destroy history without hitting any barrier.

# Where it happens

- The application connects with a single database role that owns all tables and holds full DML privileges (see the engine setup in `app/core/db.py` and the deployment defaults in `.env`).
- No database triggers, grants, or role separation protect audit data from modification.

# Why it matters

The regulatory floor identified in the epic's compliance anchoring (21 CFR Part 11 §11.10(e) "record changes shall not obscure previously recorded information", NIST 800-53 AU-9) requires that audit records cannot be silently overwritten through the normal application path. Application-level discipline alone does not satisfy this: any code path with the app's DB credentials can bypass it.

# How to observe

1. With the audit store in place, connect as the application role (`make db-shell`).
2. `UPDATE` or `DELETE` an audit row: it succeeds with no error and no trace.

hamza-56:
## What
`audit_events` is now append-only at the database level: triggers reject `UPDATE`, `DELETE`, and `TRUNCATE`, so nothing holding the application's database credentials can rewrite or erase history through the normal path.

## Changes
- feat(core): define the per-dialect append-only trigger DDL once on the `AuditEvent` model and run it from the table's `after_create` event, so `create_all` (SQLite tests) and the migration install the same guard
- feat(migrations): `a282a83eec61` installs the triggers on existing databases; downgrade removes them and leaves rows untouched
- test: SQLite tests prove `UPDATE`, `DELETE`, and whole-table `DELETE` are rejected while inserts still work
- docs: README audit section describes the protection

## How to Test
1. `make test.backend` (ruff, mypy, pytest all green; 1968 passed locally)
2. `make migrations`, then `make db-shell` and run `UPDATE audit_events SET outcome = 'denied';`, `DELETE FROM audit_events;`, and `TRUNCATE audit_events;`: each fails with `audit_events is append-only`
3. Inserts still succeed: log in and confirm a new `auth.login` row appears

Verified locally on PostgreSQL 18 (scratch cluster): all three statements rejected, the downgrade statements reopen the table, the upgrade statements protect it again.

## Notes
Migration required (`make migrations`), app lineage, no data change.

Chose triggers over a separate read-only database role: one migration, no deployment or `.env` changes, and it covers TRUNCATE. The retention job planned in #507 will need a privileged path (a `SECURITY DEFINER` function, or a role allowed to disable the triggers); that is the upgrade point, not something added here.


hamza-56:
@claude-review

sparkth/migrations/app/versions/a282a83eec61_protect_audit_events_from_update_and_.py
```diff
@@ -0,0 +1,35 @@
...
+Revision ID: a282a83eec61
+Revises: 0a2dfb2685ad
+Create Date: 2026-09-16 14:51:28.965178
+
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
```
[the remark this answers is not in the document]
hamza-56: Fair point. Inlined the DDL in db757d30 and dropped the now-unused drop-DDL constant from models. Verified upgrade, downgrade and re-upgrade against a scratch Postgres.

sparkth/core/audit/models.py
```diff
@@ -164,3 +171,42 @@ def from_event(cls, event: BaseAuditEvent, context: AuditContext) -> "AuditEvent
...
+    "sqlite": [
+        f"""CREATE TRIGGER audit_events_no_update BEFORE UPDATE ON audit_events
+BEGIN SELECT RAISE(ABORT, '{APPEND_ONLY_MESSAGE}'); END""",
+        f"""CREATE TRIGGER audit_events_no_delete BEFORE DELETE ON audit_events
+BEGIN SELECT RAISE(ABORT, '{APPEND_ONLY_MESSAGE}'); END""",
+    ],
+}
+
+
+@event.listens_for(SQLModel.metadata.tables[AuditEvent.__tablename__], "after_create")
```
abdulrafey1, line 208: `create_all(checkfirst=True)` skips an existing table, so `after_create never` fires.

Pre-existing E2E SQLite DBs get no triggers, silently.

hamza-56: The E2E script deletes e2e.db before every run (scripts/e2e-backend.sh, rm -f e2e.db e2e.db-*), so the table is always created fresh and after_create fires. The only other create_all path is the in-memory test engine, also fresh each session; Postgres gets the migration. So no persistent SQLite database goes without the triggers today. Happy to add a CREATE TRIGGER IF NOT EXISTS pass to e2e_init_db if you'd rather not rely on the rm.

abdulrafey1: yes, please lets add `CREATE TRIGGER IF NOT EXISTS`

zaira-bibi (approved):

sparkth/core/audit/models.py
hamza-56: Done in abc2a34d: SQLite triggers are CREATE TRIGGER IF NOT EXISTS behind a public install_append_only_triggers(), and e2e_init_db runs it after create_all. Verified on a db whose triggers had been dropped.

abdulrafey1 (approved):

PR #661, opened 2026-09-16
feat(plugins): emit conversation and message analytics events

zaira-bibi:
> **Stack 1 of 2** — Analytics Milestone 2, chat producers. This is the bottom layer; [the completion-seam PR](https://github.com/edly-io/sparkth/pull/662) builds on it.
>
> Supersedes #576, which was closed unmerged after the chat plugin was refactored beneath it (#633, #634, #638, #639) — both functions it patched had moved. This is the same work re-sited against the current code.

## What
Wires the first two of Milestone 2's four `chat.*` analytics events into the real chat request path, so a course-authoring conversation is actually measured rather than only measurable.

## Changes
- feat(plugins): emit `chat.conversation_started` from the completions route's create branch, queued after `schedule_title_generation`
- feat(plugins): return the stored rows from `ChatService.add_incoming_messages` and emit `chat.message_sent` per instructor turn
- fix(plugins): document that `message_length` measures the stored row (so an attachment-only turn measures the `"[Document attachment]"` placeholder) and that `has_attachment` covers only an inline upload, not `document_ids`
- refactor(plugins): move the emit scheduling behind `ChatTurnAnalytics`, built once per turn, so each route seam names what happened instead of repeating the conversation/provider/model/actor identity at every call
- test(plugins): add end-to-end producer tests driving the real `POST /api/v1/chat/completions` and asserting on rows landed in the analytics database

## How to Test
1. The producer tests, which drive the real endpoint and assert on landed analytics rows:
   ```bash
   ```
   Expected: 7 passed.
2. Confirm a second turn on an existing conversation does not re-emit `conversation_started`:
   ```bash
   uv run pytest sparkth/plugins/chat/tests/test_analytics_producers.py -k continuing -v
   ```
3. No regressions in the plugin — in particular the conversation-title suite, which shares the background-task queue this PR touches:
   ```bash
   ```
   Expected: 393 passed.
4. Confirm nothing in the analytics write path catches broadly:
   ```bash
   grep -rn "except Exception" sparkth/lib/analytics/ sparkth/plugins/chat/
   ```
   Expected: no output.

## Notes
- **Emit ordering is deliberate and load-bearing.** Starlette's `BackgroundTasks.__call__` is a plain `for task in self.tasks: await task()` with no per-task isolation, and this milestone's emits propagate their failures by design (#572). An emit queued *ahead* of real work would let an analytics outage silently stop that work — concretely, `generate_conversation_title` would never run. `emit_conversation_started` is queued last for exactly that reason, with a comment at the call site so it does not get "tidied" back up next to the create. **Please keep analytics emits queued after functional background work in any future change to these seams.**
- **`ChatService.add_incoming_messages` changed its return type** from `None` to `list[Message]`. Its one production call site is updated here; the three test callers ignore the return value. The rows are returned so the caller can measure the *stored* content — this method is the only place that flattens a request's content blocks into message text, so deriving a length from the raw request would duplicate that flattening and could drift from it.
- **No message text reaches any payload.** `chat.message_sent` carries a character count and an attachment flag; a test asserts the question string appears nowhere in the landed payload, and the base schema's `extra="forbid"` rejects anything unexpected.
- **Known gaps, documented for the consumer work** (see the "Known gaps" note added to `sparkth/plugins/chat/analytics.py` in the stacked PR): events queued on `background_tasks` are discarded when the route raises, so these counts are lower bounds on failed turns.
- **No migration.** New event types are rows in the existing `raw_events` table, not DDL.


sparkth/plugins/chat/routes/completions.py
```diff
@@ -161,10 +162,35 @@ async def chat_completion(
...
+            provider=provider_name,
+            model=model,
+            actor_id=str(user_id),
+        )
 
     if request.document_ids:
         await service.attach_owned_documents(session, conversation_id, request.document_ids, user_id)
-    await service.add_incoming_messages(session, conversation_id, request.messages)
+    incoming_messages = await service.add_incoming_messages(session, conversation_id, request.messages)
+    for stored in incoming_messages:
```
hamza-56, line 181: The route was just thinned out in #639 and this puts two analytics blocks plus a loop straight back into it, with the same conversation/provider/model/actor kwargs repeated in each. Feels like the scheduling of these emits belongs in one place in `analytics.py`, so the route only says "record this turn" and stays a coordinator.

hamza-56 (commented):
Left one design comment inline, otherwise this reads well.

sparkth/plugins/chat/routes/completions.py
zaira-bibi: One part I didn't do: the route can't collapse to a single "record this turn" call. The events fire at three different points for reasons that aren't cosmetic:`conversation_started` only on the create branch and only after the title task is queued, `message_sent` only once the rows are persisted (that's the sole place request content blocks are flattened into text, so it's the only place a truthful length exists), and the completion events only once the LLM has answered. So it's a scheduling surface rather than one call.

sparkth/plugins/chat/routes/completions.py
```diff
@@ -161,10 +169,24 @@ async def chat_completion(
...
+        turn_analytics.schedule_conversation_started()
 
     if request.document_ids:
         await service.attach_owned_documents(session, conversation_id, request.document_ids, user_id)
-    await service.add_incoming_messages(session, conversation_id, request.messages)
+    incoming_messages = await service.add_incoming_messages(session, conversation_id, request.messages)
+    for stored in incoming_messages:
+        # Instructor turns only: an assistant turn in the request is history the client
+        # replayed, and assistant output is counted by chat.completion_served instead.
+        if stored.role != "user":
```
abdulrafey1, line 184: Why are we not emitting events for when the system sends a message.

That is data that could definitely be of use. WDYS?

This might not be the right place for this particular event emission (I don't know), but it should be recorded

zaira-bibi: Drafted issue #676 for this.

abdulrafey1 (approved):

PR #662, opened 2026-09-16
feat(plugins): emit completion and tool analytics events

zaira-bibi:
> **Stack 2 of 2** — Analytics Milestone 2, chat producers. Based on [#661](https://github.com/edly-io/sparkth/pull/661), which wires the conversation and message seams. **Review #661 first** — this PR targets that branch, so its diff shows only the completion seams.
>
> Together with #661 this supersedes #576, closed unmerged after the chat plugin was refactored beneath it.

## What
Wires the remaining two `chat.*` events — `chat.completion_served` and `chat.tool_invoked` — into both completion paths, and keeps a failing analytics write from ever reaching the instructor's conversation.

## Changes
- feat(analytics): add `emit_events` to `sparkth.lib.analytics` — lands a group of related events in one session and one transaction; `ingest_event` grows a `commit` flag to allow it, default unchanged
- feat(plugins): emit `completion_served` and one `tool_invoked` per execution from the non-streaming branch, reading `metadata["tool_executions"]`
- feat(plugins): give `ChatStreamProcessor` an optional `CompletionAnalyticsContext` and emit the same two events from the streaming path
- fix(plugins): move the streaming emits out of the guarded region in `_process_and_stream`, so an analytics failure cannot write a bogus error into the conversation
- fix(plugins): hold a strong reference to the detached stream task via `live_stream_tasks`, so it cannot be garbage-collected mid-flight
- docs(plugins): correct `ChatCompletionServed`'s docstring and record two known gaps for the consumer work
- refactor(plugins): collapse the two completion emit helpers into `emit_completion`, and replace `CompletionAnalyticsContext` with `AnalyticsAttribution` — the stream processor already owns the model and whether RAG ran
- refactor(plugins): return the stream completion facts instead of stashing them on `self._completion`
- test(analytics): add contract tests for `emit_events` — ordering, one session, and the all-or-nothing guarantee
- test(plugins): add streaming and non-streaming producer tests, plus a regression test for the corruption fix

## How to Test
1. The full producer suite, both paths:
   ```bash
   ```
   Expected: 14 passed.
2. The regression test for the corruption fix — a failing `chat.completion_served` write on the streaming path must not produce an error SSE or an `is_error` row:
   ```bash
   uv run pytest sparkth/plugins/chat/tests/test_analytics_producers.py -k corrupt -v
   ```
   To see it bite, move the emit block in `_process_and_stream` back inside the `try` and watch it fail.
3. The stream processor's own suite, which constructs the processor without an analytics context:
   ```bash
   uv run pytest sparkth/plugins/chat/tests/test_stream_processor.py -v
   ```
4. The whole backend gate:
   ```bash
   make test.backend
   ```
   Expected: 1984 passed, 3 skipped (the `pg`-marked TimescaleDB lane skips without `ANALYTICS_TEST_PG_URL`).
5. The analytics batching contract, including that a bad event anywhere in a group lands none of it:
   ```bash
   uv run pytest tests/analytics/test_emit_events.py -v
   ```
   Expected: 6 passed.

## Notes
- **This PR reaches outside the chat plugin**, into `sparkth/lib/analytics` and `sparkth/core/analytics/gateway.py`, to add `emit_events` and the `commit` flag it needs. That came out of review: `emit_event` opens a session per call, so a turn with N tool executions paid 1+N session acquisitions, and sequential emits meant a failed completion write dropped every tool event behind it.
- **One transaction, not just one session.** The first cut batched into one session but still committed per event, and that made the streaming producer test flaky (~40%, one tool event invisible to the reader): the test engine uses a `StaticPool`, so every session shares one connection and the reader's view interleaved with the writer's repeated commits. Committing the group once fixed it, and gives consumers a guarantee worth having — a completion never lands without the tool events belonging to it.
- **The streaming emits must stay outside the guard.** `_persist_and_emit_done` runs inside `_run()`, which `_process_and_stream` wraps in `except BaseException` — and that handler writes an `is_error=True` assistant row and pushes an error SSE. With the emits inside it, a failing analytics write made a *successful* streamed reply render an error banner and permanently polluted the transcript with a fake error message. The emits now run after the whole `try/except/finally` and after the SSE sentinel, so a failure is genuinely unhandled on the detached task and never touches the stream. Please do not move them back.
- **`tool_calls` is a trap.** `response["tool_calls"]` is hard-coded to `None` by every provider (`sparkth/llm/providers.py:173`, `:278`); executions are exposed only as `metadata["tool_executions"]`. Reading `tool_calls` would pin `tool_call_count` to zero forever while still passing a no-tool test.
- **`ChatStreamProcessor.stream()`'s task holder is now a real caller seam**, not a test hook: asyncio only weakly references running tasks, so before this the detached stream task was garbage-collectable mid-flight — a pre-existing hole that the later-running emits widened. `live_stream_tasks` is a module-level `set` that discards each task on completion.
- **Two known gaps are documented rather than fixed**, in `sparkth/plugins/chat/analytics.py`'s module docstring, because closing either means changing already-merged analytics plumbing: (1) `occurred_at` is stamped at write time, so on a streamed turn `completion_served` is timestamped *before* that turn's `conversation_started`; consumers must not infer per-turn ordering from it. (2) Events queued on `background_tasks` are dropped when the route raises, and a mid-stream disconnect yields a `completion_served` with no `conversation_started` — so the counts are lower bounds, and failed turns are systematically under-counted. Both are worth picking up in the Milestone 3 consumer work.
- **No payload carries tool arguments or outputs.** `tool_names()` reads only the name key, and a test asserts the tool input string appears nowhere in a landed payload.
- **No migration.** New event types are rows in the existing `raw_events` table.


sparkth/plugins/chat/routes/utils/stream_processor.py
```diff
@@ -383,14 +422,33 @@ async def _process_and_stream(self) -> None:
             finally:
                 await self._put(None)
 
-    async def stream(self, _task_holder: list[asyncio.Task[None]] | None = None) -> AsyncGenerator[str, None]:
+        if self.analytics is None or self._completion is None:
+            return
+        tool_call_count, executed_tools = self._completion
+        await emit_completion_served(
```
hamza-56: This is an N+1 on the analytics DB: `emit_event` opens its own session per call, so a turn with N tool executions does 1 + N sequential sessions/round trips. Same on the non-streaming branch. And because they run one after another, a failed `completion_served` write drops every `tool_invoked` behind it. One session for the whole completion would fix both.

sparkth/plugins/chat/routes/completions.py
```diff
@@ -282,9 +290,17 @@ async def chat_completion(
                 rag_llm,
                 rag_search_required,
                 rag_search_declined,
+                analytics=CompletionAnalyticsContext(
```
hamza-56: The context duplicates state the processor already owns: `rag_used` mirrors `rag_search_required` (passed two args up) and `model` mirrors `provider.model`. Two sources of truth for the same facts that can drift. The processor only lacks the provider name and actor id.

sparkth/plugins/chat/routes/utils/stream_processor.py
```diff
@@ -348,6 +371,13 @@ async def _persist_and_emit_done(
                 },
             }
         )
+        # Recorded, not emitted here. Everything in this method runs inside _run(), which
+        # _process_and_stream wraps in `except BaseException` — a failed analytics write
+        # there would be turned into an is_error assistant row and an error SSE, so a
+        # successful reply would render as a failure and permanently pollute the
+        # transcript. _process_and_stream emits these after that guard, and after the
+        # stream has closed.
+        self._completion = (len(completed_tool_calls), tool_names(completed_tool_calls))
```
hamza-56: Stashing the result on `self._completion` for a later method to pick up is hidden mutable state with temporal coupling. `_run` could just return the completion facts and `_process_and_stream` emit from the return value; same effect, no side channel.

hamza-56 (commented):
The guard placement and the strong task reference both make sense. Three design points inline.

sparkth/plugins/chat/routes/utils/stream_processor.py
zaira-bibi: Agreed, and done in dc921ef4. `_run_llm_phase` and `_run` now return the completion facts (`tuple[int, list[str]] | None`, None meaning the turn produced no completion), and `_process_and_stream` emits from the return value. `self._completion` is gone.

The emits still run after the guard and after the sentinel — that placement is what keeps a failed analytics write from being caught as a stream error and written into the conversation as an `is_error` row, so only the side channel changed, not the ordering.

sparkth/plugins/chat/routes/utils/stream_processor.py
zaira-bibi: Fixed in 5f540045, and it turned out to be worth more than the round trips.

Added `emit_events` to `sparkth.lib.analytics`: it takes a group, acquires the session once, and commits the whole group as **one transaction** — so the events land together or not at all, rather than a completion landing without the tools behind it. `ingest_event` grew a `commit` flag to allow that; the default leaves every existing caller unchanged. On the chat side `emit_completion_served` and `emit_tool_invoked` collapse into one `emit_completion`, so a caller can't queue the completion and forget the tools, and `tool_call_count` is derived from the same list instead of passed alongside it.

Worth flagging what this exposed: batching into one long-lived session with a commit per event made `test_streaming_completion_emits_served_and_tool_invoked` flaky (~40% fail, one tool event invisible to the reader). The test engine uses a `StaticPool`, so every session shares one connection, and the reader's view interleaved with the writer's repeated commits. One transaction per group fixed it — 8/8 on the file, then 3 clean `make test.backend` runs at 1984 passed. Going one-session-but-still-N-commits would have shipped a flaky test.

sparkth/plugins/chat/routes/completions.py
zaira-bibi: Correct on both counts — fixed in 5f540045. `CompletionAnalyticsContext` is gone. The processor now takes `AnalyticsAttribution`, which carries only the provider name and actor id, and derives the rest from what it already holds:

```python
await emit_completion(
    conversation_id=self.conversation_uuid,
    attribution=self.analytics,
    model=self.provider.model,
    rag_used=self.rag_search_required,
    streamed=True,
    executed_tools=executed_tools,
)
```

The actor id has to keep travelling in rather than being read off the processor, for the reason the old docstring gave: `self.user_id` is set from `rag_user_id = user_id if rag_search_required else None`, so it's `None` on every non-RAG completion, and reading it would drop the actor from most events. The provider *name* the processor genuinely has no source for. Those two are all that's left.

sparkth/plugins/chat/analytics.py
```diff
@@ -9,6 +9,26 @@
...
+----------
+
+Two deliberate limitations of the producer seams, recorded here so consumers do not
+read more into these rows than they carry:
+
+**Ordering.** ``occurred_at`` is stamped when the row is written, not when the seam
+fires. The route's events run as Starlette background tasks after the response has
+completed, while the streaming completion event is awaited inside the stream's
+detached task — so on a streamed turn ``chat.completion_served`` is timestamped
+*before* the ``chat.conversation_started`` and ``chat.message_sent`` of that same
```
abdulrafey1: Can we look more into it, so the event's timestamp represents the actual time of when the event occured.

zaira-bibi: `occured_at` is now stamped when the event actually happened, instead of when the event landed.

abdulrafey1 (approved):

PR #663, opened 2026-09-16
fix(analytics): wait out a concurrent refresh when backfilling aggregates

hamza-56:
## What
The TimescaleDB CI lane fails intermittently with `could not refresh continuous aggregate "login_activity_daily" due to a concurrent refresh` (seen on #659 and twice on #660); the same race can hit `make migrations` in production, so the backfill now waits out a concurrent refresh instead of failing.

## Changes
- fix(analytics): retry `refresh_continuous_aggregate` on `lock_not_available` (SQLSTATE 55P03), up to 10 times with a 0.5 s pause, in `backfill_continuous_aggregates`
- refactor(analytics): let `backfill_continuous_aggregates` take an explicit engine, and point the TimescaleDB test lane's reset at it instead of duplicating the Timescale SQL
- test(analytics): cover the retry, the retry budget, non-retry of other database errors, and the explicit engine
- docs(migrations): explain the race in the database-migrations skill

## How to Test
1. Unit tests for the retry logic (stub engine, no database needed):
   ```bash
   uv run pytest tests/analytics/test_continuous_aggregate_backfill.py -v
   ```
   Expected: 11 passed.
2. The TimescaleDB lane against the local services:
   ```bash
   make test.backend.analytics
   ```
   Expected: 3 passed.
3. To see the retry absorb real contention, keep a full refresh looping in one shell while the lane runs in another:
   ```bash
   while true; do docker compose exec -T db psql -U sparkth -d sparkth_analytics_test -qtA -c "CALL refresh_continuous_aggregate('login_activity_daily', NULL, NULL)" >/dev/null 2>&1; done
   ```
   ```bash
   ANALYTICS_TEST_PG_URL=postgresql://sparkth:sparkth_password@localhost:5432/sparkth_analytics_test uv run pytest -m pg tests/analytics/pg -o log_cli=true -o log_cli_level=WARNING
   ```
   Expected: `is being refreshed concurrently; retrying` warnings, 3 passed. On `main` the same setup errors with `LockNotAvailableError`.

## Notes
Root cause: `add_continuous_aggregate_policy` schedules the policy's first run immediately (verified: `job_stats.total_runs = 1` with `last_run_started_at` inside the test run, right after the migration). TimescaleDB allows one refresh per aggregate at a time, so a full refresh issued right after the migration (the CI fixture's per-test reset, or the post-migration backfill) collides with that first policy run on a slow runner. No migration, no env vars, no dependency changes.


hamza-56:
Closing without merging for now; the TimescaleDB lane flake stays open.

PR #667, opened 2026-09-18
feat(plugins): add local_sparkth moodle plugin skeleton and ci

abdulrafey1:
## What

Moodle core web services cannot author activity content, so Sparkth cannot publish a
chat-authored course into Moodle; this adds the companion Moodle-side plugin's skeleton and its
two-version PHPUnit CI, before any of the three functions it will expose are implemented.

## Changes

- feat(plugins): add the `local_sparkth` Moodle plugin at `third_party_plugins/moodle/sparkth/`
  with `version.php`, `db/services.php` declaring all three web service functions, and the
  English language file
- feat(plugins): ship an external service `Sparkth publishing` exposing the three functions, so a
  site admin authorises a token user rather than hand-building a service on every client install
- ci(plugins): add `.github/workflows/moodle.yml`, a PHPUnit job with a real matrix over Moodle
  4.5 LTS and 5.x

## How to Test

1. `git checkout rafe/moodle-plugin-skeleton`
2. Clone Moodle 5.x and `moodlehq/moodle-docker`, set `MOODLE_DOCKER_WWWROOT` and
   `MOODLE_DOCKER_DB=pgsql`, and bind-mount `third_party_plugins/moodle/sparkth` to
   `<wwwroot>/public/local/sparkth` via `moodle-docker/local.yml`
3. `bin/moodle-docker-compose up -d webserver && bin/moodle-docker-wait-for-db`
4. `bin/moodle-docker-compose exec webserver php public/admin/tool/phpunit/cli/init.php`
5. `bin/moodle-docker-compose exec webserver vendor/bin/phpunit --testsuite local_sparkth_testsuite`
6. Expect 3 tests passing: the plugin is installed, the three functions are registered in
   `external_functions`, and the external service links all three
7. Repeat against Moodle 4.5 LTS, mounting to `<wwwroot>/local/sparkth` (4.5 has no `public/`) and
   using `admin/tool/phpunit/cli/init.php`

## Notes

- **No database tables and no migration.** This plugin declares no `db/install.xml`.
- **Installed on the target Moodle, not on Sparkth.** A site admin installs `local_sparkth` on each
  Moodle that Sparkth publishes into, enables web services and the REST protocol, authorises the
  token user on the `Sparkth publishing` service, and mints a token for a user holding
  `moodle/course:create`, `moodle/course:manageactivities` and `moodle/question:add`. Because the
  service is declared `restrictedusers => 1`, an unauthorised token returns `accessexception` on
  every call — this is the most likely first-install failure.
- **`$plugin->version` must be bumped whenever `db/services.php` or `lang/` changes**, or installed
  sites never pick the change up. Moodle only re-reads those on a version increase.
- The CI job pins `MOODLE_DOCKER_PHP_VERSION: "8.3"`. Moodle 4.5 tops out at PHP 8.3 and Moodle 5.2
  requires at least 8.3.0, so 8.3 is the only value satisfying both matrix legs.
- First of a four-PR stack. Merge in order; do not delete a base branch until the whole stack has
  landed, as that closes the PRs stacked on it.


hamza-56 (commented):

.github/workflows/moodle.yml
```diff
@@ -0,0 +1,74 @@
+name: Moodle plugin
+
+on:
+  pull_request:
```
hamza-56, line 4: This spins up two full Moodle stacks (clone, docker, phpunit init) on every PR in the repo, frontend-only ones included. Feels like it should be scoped to changes under the plugin dir and this workflow.

PR #668, opened 2026-09-18
feat(plugins): add local_sparkth_create_section web service

abdulrafey1:
## What

Adds the first of the three web service functions the companion plugin exists to provide:
appending a named, described section to a Moodle course, which core web services cannot do.

## Changes

- feat(plugins): add `local_sparkth\external\create_section`, wrapping core
  `course_create_section()` and `course_update_section()`
- test(plugins): cover section numbering, persistence of name and summary, and the
  capability-denied path

## How to Test

1. `git checkout rafe/moodle-create-section`
2. With the local Moodle stack from PR #667 running, re-run
3. Expect 6 tests passing on Moodle 5.x **and** on Moodle 4.5 LTS
4. Confirm the first call returns `sectionnum` 1, not 0 — a course created without
   `courseformatoptions` already has section 0, Moodle's unnamed general section

## Notes

- Sections are never pre-allocated. `numsections` is deliberately left unset at course creation so
  that `create_section` is the only thing that creates sections, and authored content starts at
  section 1 with no stray empty topic above it.
- The tests extend `\core_external\tests\externallib_testcase`, not the legacy
  `\externallib_advanced_testcase` global alias — that alias only exists once
  `webservice/tests/helpers.php` has run, so it breaks under a scoped `--testsuite` run.
- They also create courses with `['numsections' => 0]`: the test data generator defaults
  `numsections` to 5, unlike core `create_course()`.
- This function gates section creation on `moodle/course:manageactivities`, where core gates the
  same operation on `moodle/course:update` and additionally enforces `get_max_sections()`. Under
  default roles the two are co-extensive, but a caller looping on this function can exceed the
  site's section limit.
- Second of a four-PR stack, based on #667. Merge in order.


hamza-56 (approved):
LGTM.

PR #669, opened 2026-09-18
feat(plugins): add local_sparkth_create_page web service

abdulrafey1:
## What

Adds the second web service function: creating a Page activity that actually carries the
chat-authored lesson HTML, which no core web service can do.

## Changes

- feat(plugins): add `local_sparkth\external\create_page`, wrapping core `add_moduleinfo()`
- test(plugins): assert the persisted `page.content` column, section placement, and the
  capability-denied path

## How to Test

1. `git checkout rafe/moodle-create-page`
2. With the local Moodle stack running, re-run
3. Expect 9 tests passing on Moodle 5.x **and** on Moodle 4.5 LTS
4. The assertion that matters is in `test_creates_a_page_with_its_html_body`: it reads the `page`
   table and compares `content` to the literal HTML, rather than trusting the returned ids

## Notes

- **The non-obvious part, and the reason this function exists in this shape:**
  `page_add_instance()` copies the editor array into the `content` column only `if ($mform)`, and
  `add_moduleinfo()` passes `null`. Passing `page => ['text' => ...]` therefore creates a Page with
  an **empty body and no error at all**. This implementation sets the `content` and
  `contentformat` columns directly. Verified both ways against a live Moodle: the editor-array form
  produced `content=''`, the direct form produced the expected HTML. The test asserts the column so
  the trap cannot silently return.
- `add_moduleinfo()` also needs `module` — the `modules` table row id, not just `modulename` — and
  `page_add_instance` reads `printintro` and `printlastmodified` unconditionally.
- `content` and `intro` use `PARAM_RAW`, matching Moodle's own posture: `page_get_editor_options()`
  sets `noclean`, so the web UI stores page HTML raw too. This grants nothing the UI does not.
- Third of a four-PR stack, based on #668. Merge in order.


hamza-56 (approved):
LGTM.

PR #670, opened 2026-09-18
feat(plugins): add local_sparkth_create_quiz web service

abdulrafey1:
## What

Adds the last of the three web service functions — creating a Quiz activity with real questions —
and completes the companion plugin, including the one place where Moodle 4.5 and 5.x genuinely
diverge.

## Changes

- feat(plugins): add `local_sparkth\external\create_quiz`, creating the quiz, saving `multichoice`
  and `truefalse` questions, and binding them as slots
- feat(plugins): add `local_sparkth\question_bank_resolver`, the single file expressing the
  4.5-vs-5.x question bank difference
- fix(plugins): validate multichoice answer counts and `correctindex`, and verify the target
  section exists before creating a Page or a Quiz
- docs(plugins): add GPL headers to all plugin files and record the constraints behind the code
- test(plugins): cover both question types, unsupported types, the capability-denied paths, bank
  reuse, and the two guards above

- fix(plugins): attach the three core functions the Sparkth plugin calls to the shipped
  `Sparkth publishing` service, and bump `$plugin->version`
- fix(plugins): make the question bank this plugin creates visible to authors
- feat(plugins): add `local_sparkth_enrol_creator`, enrolling the token user in a course it just
  created

## How to Test

1. `git checkout rafe/moodle-create-quiz`
2. With the local Moodle stack running, re-run
3. Expect 22 tests / 36 assertions passing on Moodle 5.x **and** on Moodle 4.5 LTS. Running only
   one version proves nothing here — the two exercise different branches of the resolver.
4. To see it end to end, publish through the three classes against a live site and confirm the
   course renders a section, a Page with its body, and a quiz whose questions appear, with
   `quiz.sumgrades` non-zero.

## Notes

- **Moodle 5.0 moved question banks into `mod_qbank` instances.** On 4.5 questions live in the
  course context; on 5.x `question_get_default_category()` returns `NULL` for a course context and
  a `mod_qbank` instance must be created. `question_bank_resolver` is the only file that knows
  this — it branches on whether the `qbank` module exists, which is a capability probe rather than
  a version comparison.
- On 5.x the plugin creates its own bank named `Sparkth question bank`, separate from the course's
  default bank. Questions authored by Sparkth will **not** appear in the course's default bank.
- `BANK_NAME` is deliberately an English literal, not a lang string: it is an identity key matched
  on to find the bank again, so translating it would make a site whose language changes silently
  fork a second bank and split a course's questions.
- Question types are limited to `multichoice` and `truefalse`. Others are rejected at the boundary.
- `quiz_defaults()` is a 25-line literal because Moodle's admin defaults store review options as
  packed bitmasks while the `quiz` table carries 24 separate columns. Drift is caught by the
  two-version CI matrix, which inserts a real row on every run.
- Test class docblocks do not yet carry `@package`/`@license`; Moodle codechecker has not been run.
- Last of a four-PR stack, based on #669. Merge in order, and do not delete a base branch until the
  whole stack has landed.

- **Two fixes here were found by integrating end to end, not by reviewing the PHP.** Both live in
  this PR because they change this plugin, even though the symptoms only appear once the Sparkth
  side calls it:
  - the shipped external service listed only this plugin's own three functions, so a token on it
    returned `accessexception` for the three **core** functions Sparkth also calls
    (`core_webservice_get_site_info`, `core_enrol_get_users_courses`, `core_course_create_courses`).
    The `$plugin->version` bump is required — Moodle re-reads `db/services.php` only when the
    version increases, so without it an already-installed site keeps the broken service.
  - `question_bank_helper::create_default_open_instance()` sets `visible = 0`, so the bank this
    plugin creates gave authors "Sorry, this activity is currently hidden" on its activity link.
    It is now created `visible = 1, visibleoncoursepage = 0` — reachable, but not listed on the
    course page beside the lessons and quizzes.

- **`local_sparkth_enrol_creator` closes a three-symptom gap.** Creating a course over web services
  does not enrol the creator, which surfaced as an empty author dashboard, an empty
  `moodle_list_courses`, and "Sorry, this activity is currently hidden" on the question bank
  (`$cm->uservisible` is false for a non-enrolled user). This mirrors Moodle's own
  `course/edit.php` rather than inventing a scheme: the role comes from `$CFG->creatornewroleid`,
  the already-enrolled guard is core's, and the user enrolled is always `$USER` — never a
  parameter, so no caller can enrol somebody else. Verified live: the token user ends up enrolled
  with the configured role and the bank's `uservisible` becomes true.


third_party_plugins/moodle/sparkth/classes/external/create_quiz.php
```diff
@@ -0,0 +1,241 @@
...
+            'section'     => $sectionnum,
+            'name'        => $name,
+            'intro'       => $intro,
+            'introformat' => FORMAT_HTML,
+            'visible'     => 1,
+            // edit_module_post_actions() reads cmidnumber unguarded for graded
+            // activities, so quiz needs it set and page does not.
+            'cmidnumber'  => '',
+        ]);
+        $created = add_moduleinfo($moduleinfo, $course);
```
hamza-56: Nothing from here down is transactional. If a question save or `quiz_add_quiz_question` throws midway, the quiz is left behind with some of its questions and the caller gets an exception without the cmid, so it can't clean up or retry safely.

third_party_plugins/moodle/sparkth/classes/external/create_quiz.php
```diff
@@ -0,0 +1,241 @@
...
+            'intro' => $intro, 'questions' => $questions,
+        ] = self::validate_parameters(self::execute_parameters(), [
+            'courseid' => $courseid, 'sectionnum' => $sectionnum, 'name' => $name,
+            'intro' => $intro, 'questions' => $questions,
+        ]);
+
+        $course = $DB->get_record('course', ['id' => $courseid], '*', MUST_EXIST);
+        $context = context_course::instance($course->id);
+        self::validate_context($context);
+        require_capability('moodle/course:manageactivities', $context);
```
hamza-56: The course lookup, validate_context, capability check and section-exists guard are now copy-pasted across all three externals. Worth pulling into one shared place in the plugin before more functions land.

hamza-56 (commented):

PR #672, opened 2026-09-18
feat(plugins): add the sparkth moodle plugin foundation

abdulrafey1:
## What

Adds `sparkth/plugins/moodle/`, the Sparkth-side plugin that publishes chat-authored courses into
Moodle, with its client, config and first tool — the foundation PRs 6-8 build on.

## Changes

- feat(plugins): add the `moodle` plugin mirroring `sparkth/plugins/canvas/` — `config.py`,
  `client.py`, `schemas.py`, `plugin.py`, `tools.py`, `constants.py`, `locale/`
- feat(plugins): add `moodle_authenticate`, backed by core `core_webservice_get_site_info`
- feat(core): register `MoodlePlugin` in `PLUGINS`
- fix(frontend): render secret plugin config fields as password inputs

## How to Test

1. `git checkout rafe/moodle-plugin-foundation`
2. `uv run pytest sparkth/plugins/moodle -v` — 19 tests
3. `uv run pytest -q` — the full suite stays green
3b. `cd frontend && npx vitest run tests/components/settings/ConfigField.test.tsx` — 2 tests
4. `uv run mypy --strict sparkth/plugins/moodle` and `uv run ruff check sparkth/plugins/moodle`
5. The two tests worth reading are the flattener table in `tests/test_client.py` and
   `test_a_body_without_an_exception_key_is_not_an_error` — see Notes for why

## Notes

- **The chat plugin is untouched, by design.** `build_lms_credentials_message` discovers any
  `PluginConfig` overriding `lms_tool_prefix()`, so credential injection and the "My Plugins"
  routing follow automatically from `MoodleConfig`.
- **`MoodleClient` is the one component that cannot copy its Canvas counterpart**, because both of
  `BaseHttpClient`'s assumptions are wrong for Moodle's REST server:
  - It takes **form-encoded** parameters with nested structures flattened into bracketed keys
    (`courses[0][fullname]=Intro`), not JSON. `flatten_ws_params` is pure and table-tested.
  - It answers a **failed call with HTTP 200** plus `{"exception": ..., "errorcode": ...}`. A
    status-only check would read every failure as success — publishing nothing while reporting
    success. The discriminator is the `exception` key, not `errorcode`, because a legitimate
    payload may carry `errorcode`; there is a test pinning exactly that.
- The client exposes only `call_dict()` and `call_list()` over a private `_call`, each raising
  `ValueError` naming the wsfunction when the response is the other shape, so `Any` never reaches a
  caller.
- `MoodleConfig` exposes `api_url` / `api_key`, matching `CanvasConfig`. The `Auth` payload uses
  `api_url` / `api_token`. The wire body still sends Moodle's own `wstoken` parameter — that name
  is Moodle's, not ours.
- `MOODLE_WS_ENDPOINT` is a plain constant, not an env read: Moodle fixes that path itself, so it
  is not site-configurable. No new environment variable.
- **Secret config fields no longer render as plain text.** `isSecretKey` in
  `frontend/components/settings/utils.ts` matches key/token/secret/password, and `ConfigField`
  renders those as `type="password"`. This addresses the review comment on this PR; it applies to
  every plugin's settings form, not just Moodle's.
- First of a four-PR stack, based on #670. Merge in order; do not delete a base branch until the
  whole stack has landed.


sparkth/plugins/moodle/config.py
```diff
@@ -0,0 +1,15 @@
+from pydantic import Field, HttpUrl
+
+from sparkth.lib.plugins import PluginConfig
+
+
+class MoodleConfig(PluginConfig):
+    base_url: HttpUrl = Field(..., description="Moodle site URL")
+    ws_token: str = Field(..., description="Moodle web service token", min_length=1)
```
hamza-56: `ws_token` will render as plain text in the settings form. Canvas marks `api_key` with the password widget for exactly this.

sparkth/plugins/moodle/constants.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+# Path of Moodle's REST web service entry point, relative to the site base URL.
+MOODLE_WS_ENDPOINT = os.getenv("MOODLE_WS_ENDPOINT", "webservice/rest/server.php")
```
hamza-56: Only spot in the codebase reading env directly rather than through `Settings`, which is why (as the .env note in #675 admits) the .env line has no effect on a native backend. Not convinced this needs to be configurable at all, the path is fixed in Moodle.

hamza-56 (commented):

PR #673, opened 2026-09-18
feat(plugins): add moodle course listing and creation tools

abdulrafey1:
## What

Adds the two course tools, and fixes two defects that only surfaced when the plugin was run against
a real Moodle — one of which made course creation fail on its default path.

## Changes

- feat(plugins): add `moodle_list_courses` and `moodle_create_course`, both backed by core web
  services
- fix(plugins): omit `lang` when it is unset instead of sending an empty string
- fix(plugins): label each of `moodle_list_courses`' two calls with the operation that actually
  failed

## How to Test

1. `git checkout rafe/moodle-course-tools`
2. `uv run pytest sparkth/plugins/moodle -v` — 28 tests
3. `uv run pytest -q`, `uv run mypy --strict sparkth/plugins/moodle`, `uv run ruff check sparkth/plugins/moodle`

## Notes

- **`lang: ""` was rejected by Moodle** with `errorinvalidparam`, and `CoursePayload.lang` defaults
  to `""` — so the default path failed for every caller that did not set a language, while all
  mocked tests passed. Verified against a live Moodle 5.2: `lang` omitted → OK, `lang: "en"` → OK,
  `lang: ""` → fails. Fixed by sending `None`, which `flatten_ws_params` already drops.
- `moodle_create_course` deliberately does not send `numsections`, so section creation has exactly
  one owner and authored content starts at section 1. There is a test asserting its absence.
- `moodle_list_courses` returns courses the token user is **enrolled in**. See #674 for why a
  published course does not appear there.
- Second of a four-PR stack, based on #672.


hamza-56 (approved):
LGTM.

PR #674, opened 2026-09-18
feat(plugins): add moodle section and page authoring tools

abdulrafey1:
## What

Adds the first two tools that call the companion Moodle plugin, so a chat-authored lesson can reach
a real course section.

## Changes

- feat(plugins): add `moodle_create_section` and `moodle_create_page`, calling the
  `local_sparkth_*` web service functions

## How to Test

1. `git checkout rafe/moodle-content-tools`
2. `uv run pytest sparkth/plugins/moodle -v` — 31 tests
3. `uv run pytest -q`, `uv run mypy --strict sparkth/plugins/moodle`, `uv run ruff check sparkth/plugins/moodle`
4. `test_missing_companion_plugin_becomes_an_error_dict` documents what a Moodle *without*
   `local_sparkth` actually returns, so that failure is recognisable rather than mysterious

## Notes

- Both tools require the `local_sparkth` companion plugin on the target Moodle (PRs #667-#670).
- `summary` and `intro` default to `""`, which is safe here: the companion declares them
  `PARAM_RAW` with `VALUE_DEFAULT ''`, and `PARAM_RAW` does no cleaning. This was checked against
  the companion's source rather than assumed — the `lang` parameter in #673 is a validated type and
  rejects `""`, so the two cases genuinely differ.
- Each tool names its `wsfunction` in a local and uses it for the call and both error helpers, so a
  failure always reports the call that actually failed.
- Third of a four-PR stack, based on #673.


hamza-56 (approved):
LGTM.

PR #675, opened 2026-09-18
feat(plugins): add moodle quiz authoring and the live test lane

abdulrafey1:
## What

Adds the quiz tool and closes the slice: a chat-authored course now publishes into a live Moodle
end to end, with an opt-in test lane that proves it and the site-admin docs to set it up.

## Changes

- feat(plugins): add `moodle_create_quiz` with a `QuestionType` enum limiting questions to
  `multichoice` and `truefalse`
- test(plugins): add the opt-in live-Moodle lane and a `make test.backend.moodle` target
- fix(plugins): guard every step of the live lane so a failure reports the status and message
- docs(plugins): document the target-Moodle prerequisites, and correct the token-role guidance
- chore(plugins): append the `moodle` pytest marker alongside `pg`

- feat(plugins): `moodle_create_course` enrols the token user in the course it creates

## How to Test

1. `git checkout rafe/moodle-quiz-and-docs`
2. `uv run pytest sparkth/plugins/moodle -v` — 35 tests, 1 skipped (the live lane)
3. `uv run pytest -q` — 2004 passed, 5 skipped; the live lane must **skip** here
4. `uv run mypy --strict sparkth`, `uv run ruff check sparkth`, `make lint.format.backend`, `make docs`
5. Against a real Moodle with `local_sparkth` installed and a token authorised on the
   `Sparkth publishing` service:
   `MOODLE_TEST_URL=... MOODLE_TEST_TOKEN=... make test.backend.moodle` — publishes a course with a
   section, a page carrying its HTML body, and a 2-question quiz

## Notes

- **The live lane's variables are never read from `.env` or `.env.local`.** Those files populate
  the `Settings` object via pydantic-settings and never reach `os.environ`, so an `os.getenv` read
  always falls back to its default — the lane would report a green "skipped" forever. It follows
  the `pg` lane's precedent instead: pass them as real environment variables, via the make target.
- **The token user needs five capabilities, not three**, and the obvious role choices fail. Each of
  these was found by running the lane and watching it fail, and is now in the docs:
  - `webservice/rest:use` is held by **no stock role** (a site admin has it implicitly); without it
    every call returns `401 Access control exception`
  - `moodle/course:view` is required because web service course creation does **not** enrol the
    creator; without it, course creation succeeds and the next call fails with
    `400 Course or activity not accessible`
  - the stock `coursecreator` role holds only `moodle/course:create`, so it can create a course and
    nothing inside it
  - verified working non-admin recipe: stock **Manager** at system context plus
    `webservice/rest:use`
- `moodle_list_courses` returns courses the token user is **enrolled in**, and a course created by
  `moodle_create_course` is not enrolled — so it will not appear there. Its docstring says so,
  which matters because MCP surfaces that text to the LLM as the tool description. Use the `id`
  the create call returns.
- An `accessexception` now carries a hint pointing at service authorisation, because the likeliest
  first-install failure otherwise reaches the agent as a vague 401 and gets reported as a bad token.
- The live lane has no CI coverage: it needs `local_sparkth` installed and a token user authorised,
  which no environment variable can supply.
- Last of a four-PR stack, based on #674. Merge in order, and do not delete a base branch until the
  whole stack has landed.

- **`moodle_create_course` now returns `enrolled` alongside `course`.** It calls
  `local_sparkth_enrol_creator` after creating, because web service course creation does not enrol
  the creator. An enrolment failure deliberately does **not** fail course creation — the course
  exists by then, so an error would misreport it. The failure is logged and surfaced as
  `enrolled: false`, which was confirmed against a live site rather than only in tests.


sparkth/plugins/moodle/tests/test_tools.py
```diff
@@ -267,3 +277,54 @@ async def test_missing_companion_plugin_becomes_an_error_dict(self) -> None:
...
+
+        wsfunction, params = client.call_dict.call_args.args
+        assert wsfunction == "local_sparkth_create_quiz"
+        assert params["questions"][0]["qtype"] == "multichoice"
+        assert params["questions"][0]["answers"] == ["3", "4"]
+        assert params["questions"][1]["qtype"] == "truefalse"
+
+    def test_an_unsupported_question_type_is_rejected_at_the_boundary(self) -> None:
+        with pytest.raises(ValidationError):
+            Question(qtype="essay", name="E", questiontext="<p>Discuss.</p>")  # type: ignore[arg-type]
```
hamza-56: New `type: ignore`. We've been keeping these out of the codebase, tests included.

docs/guides/plugins.md
```diff
@@ -392,12 +392,67 @@ creates its per-language catalogs, and so on); the containment is enforced by
...
     "sparkth.plugins.canvas.plugin:CanvasPlugin",
     "sparkth.plugins.openedx.plugin:OpenEdxPlugin",
+    "sparkth.plugins.moodle.plugin:MoodlePlugin",
     # add your plugin here
 ]
 ```
 
 Format: `"path.to.module:ClassName"`
 
+### Moodle: target-site prerequisites
```
hamza-56: This guide is about writing plugins; a page of Moodle site-admin setup and capability recipes in the middle of it is going to get lost. Reads more like the plugin's own README.

sparkth/plugins/moodle/tools.py
```diff
@@ -126,3 +140,26 @@ async def moodle_create_page(payload: PagePayload) -> dict[str, Any]:
         return _lms_error(e, wsfunction)
     except ValueError as e:
         return _malformed_response_error(e, wsfunction)
+
+
+async def moodle_create_quiz(payload: QuizPayload) -> dict[str, Any]:
```
hamza-56, line 172: Seventh copy of the same `async with MoodleClient` plus two-except scaffold. Canvas has a similar shape, but with two except arms per tool it's getting noisy here; a small wrapper would collapse all of these.

hamza-56 (commented):

PR #678, opened 2026-09-18
feat(core): add per-category audit retention and gdpr erasure

hamza-56:
## What
Gives the audit trail a per-category retention window and an erasure path that honours GDPR Art. 17 without rewriting or deleting sealed rows, as the two only writes allowed past the append-only guard from #660.

## Changes
- feat(core): `AUDIT_RETENTION_DAYS` (default 3 years) and `AUDIT_RETENTION_OVERRIDES` (JSON, per category, 0 = forever) settings
- feat(core): `purge_expired_events()` deletes rows past their category's window, one `audit.retention_purged` event per slice that lost rows
- feat(core): `erase_actor(user_id)` blanks `actor_label`, `request_ip`, `user_agent` on a user's events and records `audit.actor_erased`; `actor_id` stays as a pseudonym, sealed content untouched
- feat(core): the actor label leaves the canonical bytes, so the seal only covers what erasure never touches
- feat(core): `audit purge` and `audit erase-user <id>` CLI commands
- feat(migrations): `86af5a3d3959` replaces the PostgreSQL guard function so it stands down while `SET LOCAL sparkth.audit_maintenance = 'on'` (TRUNCATE stays rejected); SQLite drops and recreates its triggers inside the maintenance transaction
- docs: README audit section, `.env`, configuration reference

## How to Test
1. `uv run pytest tests/audit` (purge per category, 0 = forever, erasure scoped to one actor, guard re-armed after both).
2. On PostgreSQL: `make migrations`, then `psql` `DELETE FROM audit_events` and `TRUNCATE audit_events` are still rejected with `audit_events is append-only`.
3. Set `AUDIT_RETENTION_DAYS=1` in `.env.local`, log in, then `make cli -- audit purge`: nothing yet; backdate a row's `occurred_at` is impossible by design, so instead run with `AUDIT_RETENTION_OVERRIDES={"auth": 1}` a day later, or trust step 1.
4. `make cli -- audit erase-user <id>`: that user's rows show `actor_label`, `request_ip`, `user_agent` as NULL, other users' rows untouched, and a new `audit.actor_erased` row targets the user.

## Notes
Stacked on #660 (the base branch); merge that first. Migration `86af5a3d3959` is PostgreSQL-only and rewrites the trigger function only. New env vars `AUDIT_RETENTION_DAYS`, `AUDIT_RETENTION_OVERRIDES` with defaults in `.env`. Verified the migration's upgrade and downgrade and the full purge/erase path on a scratch PostgreSQL 18.

Deliberately left out: legal holds and per-record retention (add when a customer needs Title IV holds), batching the purge DELETE, erasing a claimed username on failed-login rows (it sits in the sealed target and was never verified as that user's), and account deletion itself, which does not exist yet. The guard is triggers plus a session flag, not a separate maintenance role: it stops the app and accidents, not a holder of the owner credentials.

Capture rule this introduces: never put the actor's own identity (username, email, name, IP) into a sealed payload. #659's `auth.registered` payload carries the username and will get a one-line fix there.


sparkth/cli/audit.py
```diff
@@ -0,0 +1,39 @@
+"""Audit trail maintenance commands: retention purge and GDPR erasure (issue #507).
```
abdulrafey1: lets remove the issue number from any doc-string or comments

sparkth/core/audit/models.py
```diff
@@ -176,16 +184,27 @@ def from_event(cls, event: BaseAuditEvent, context: AuditContext) -> "AuditEvent
 APPEND_ONLY_MESSAGE = "audit_events is append-only"
 
 # Per dialect: the statements that make the table append-only, run by ``create_all``
-# on a fresh database. Migration a282a83eec61 carries its own frozen copy for existing
-# databases; a change here needs a new migration. SQLite has no TRUNCATE; its
-# truncate optimization is disabled by the presence of a DELETE trigger, so a
-# bare ``DELETE FROM audit_events`` is caught row by row.
-# ponytail: triggers, not role separation; a retention job (#507) will need a
-# privileged path (SECURITY DEFINER function or a role that may disable them).
+# on a fresh database. Migrations a282a83eec61 and its successor carry their own
```
abdulrafey1: lets not make explicit references to migrations, we can simple say `previous migrations`

abdulrafey1 (commented):
Good work!
Just a couple of minor comments

zaira-bibi (approved):
just  the existing comments, rest all good!

sparkth/cli/audit.py
hamza-56: Removed in 9425f6ab, along with the one in the README.

sparkth/core/audit/models.py
hamza-56: Reworded to "previous migrations" in the rebase; the README now says "a migration" too (9425f6ab).

abdulrafey1 (approved):

